PR #6, opened 2016-02-21
Support arbitrary titles

sebastianbenz:
+ use encodeURI for file names
+ add specs for file name conversions
+ double encode links to match file name

I'm not sure double URI encoding the URLs is the best way to go. Here is how it's handled at the moment:

* Title: 'Let's go'
* Filename: 'Let%27 go'
* URL: 'Let%2527%20go' -> server decodes it automatically, decodes version matches the filename

//cc @juliantoledo @jkingyens please review


jkingyens:
LGTM

PR #7, opened 2016-02-22
add amp-anim example

ymotongpoo:
add an example of amp-anim and related assets (animation gif and placeholder image file) with it.


jmadler:
LGTM.  Created #8 for one missing use case which is dependent on a FR I've yet to complete.

sebastianbenz:
LGTM - Merged. Thanks!

PR #18, opened 2016-02-24
Clear current tag after tag end

sebastianbenz:
Fixes error where each header line is split into a different section


jkingyens:
LGTM

PR #24, opened 2016-02-24
add travis config

sebastianbenz:
@jkingyens does this travis config make sense?


sebastianbenz:
Build works. Only deploying to app engine is untested.

jkingyens:
LGTM

PR #25, opened 2016-02-26
add amp-accordion

sebastianbenz:
//cc @jkingyens this is an example for a component that won't validate as it is still in experimental.

### DIFF
--- a/src/amp-accordion.html
+++ b/src/amp-accordion.html
@@ -1,0 +1,81 @@
+<!--
+  #### Introduction
+
+  An accordion provides a way for viewers to have a glance at the outline of the content and jump to a section or their choice at their will. 
+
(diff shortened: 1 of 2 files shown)


juliantoledo:
LGTM, just one detail, should we put (experimental) next to it, and any other, in the main page to highlight that fact?

sebastianbenz:
I thought the same - what's a good way to mark examples as experimental?

juliantoledo:
Maybe just a text next to them in the list (experimental: what does this mean?) with a link to https://github.com/ampproject/amphtml/tree/master/tools/experiments in the question.

sebastianbenz:
but where do we add the flag when writing the sample?

juliantoledo:
Would it be possible to read a <meta>? We can do it later in another pull.

Issue #12, opened 2016-02-23
Generate link to next example into each example file


sebastianbenz:

sebastianbenz:
Fixed by #34

PR #34, opened 2016-02-29
Navigate between articles

kul3r4:
Added support for navigation between pages (prev and next) as for [Generate link to next example into each example file](https://github.com/ampproject/amp-by-example/issues/12) .

The change is made to the compile-example.js. 
When streaming the files in order to apply the template, when examining the first file only the info about the files are saved. When examining the second file, the previous arguments are used to apply the template using the current file name as "next", and so on.
The file "last.html" is a dummy file used for enabling the last file to be added correctly.


kul3r4:
I signed it!

2016-02-29 15:11 GMT+00:00 googlebot <notifications@github.com>:

> Thanks for your pull request. It looks like this may be your first
> contribution to a Google open source project. Before we can look at your
> pull request, you'll need to sign a Contributor License Agreement (CLA).
>
> [image: :memo:] *Please visit https://cla.developers.google.com/
> <https://cla.developers.google.com/> to sign.*
>
> Once you've signed, please reply here (e.g. I signed it!) and we'll
> verify. Thanks.
> ------------------------------
>
>    - If you've already signed a CLA, it's possible we don't have your
>    GitHub username or you're using a different email address. Check your
>    existing CLA data <https://cla.developers.google.com/clas> and verify
>    that your email is set on your git commits
>    <https://help.github.com/articles/setting-your-email-in-git/>.
>    - If you signed the CLA as a corporation, please let us know the
>    company's name.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ampproject/amp-by-example/pull/34#issuecomment-190251842>
> .
>



-- 
Chiara Chiappini

src/last.html
```diff
@@ -0,0 +1,54 @@
+<!--
```
sebastianbenz: Can we find the next sample via the filesystem instead? I'm currently working on adding support for sections which requires examples being put into different subfolders. The last marker via file might no longer work.

src/templates/example.html
```diff
@@ -34,6 +34,14 @@
         </div>
       </div>
       {{/sections}}
+      <div style="text-align: center">
+        <div style="display: inline-block; margin: 0 auto;">
+          <a href="{{prev}}">Prev</a>
```
sebastianbenz: I'd prefer to only have the next element (the previous is not relevant when going through the samples.

src/templates/example.html
```diff
@@ -34,6 +34,14 @@
...
+        </div>
+        <div style="display: inline-block; margin: 0 auto;">
+          <a href="{{next}}">Next</a>
```
sebastianbenz: Two things:

1. Can we include the title of the next sample?
2. Make the next element optional. You can define predicates with mustache:

```
{{#next}}
<div style="display: inline-block; margin: 0 auto;">
   <a href="{{next}}">Next: {{nextTitle}}</a>
</div>
{{/next}}
```

tasks/compile-example.js
```diff
@@ -43,7 +43,10 @@ module.exports = function(templateRoot, template) {
     throw new PluginError('gulp-index',
         'Missing template in template options for gulp-index');
   }
-
+  var prev;
+  var next;
```
sebastianbenz: Can we create a method `nextFile(fileName)` in FileName.js which returns the next file name? To implement nextFile you could hardcode the src folder in FileName. We can find a better solution for this later.

sebastianbenz:
Great! Added a few suggestions.

kul3r4:
PTAL

src/templates/example.html
```diff
@@ -34,6 +34,13 @@
         </div>
       </div>
       {{/sections}}
+      <div style="text-align: center">
+        {{#next}}
```
sebastianbenz: Can you put the {{next}} around the div as well? I doesn't make sense to have an empty diff there.

tasks/compile-example.js
```diff
@@ -57,13 +57,16 @@ module.exports = function(templateRoot, template) {
...
 
       const args = {
         head: document.head,
         title: title,
         subHeading: title,
         exampleStyles: document.styles,
-        sections: document.sections
+        sections: document.sections,
+        next: next,
+        nextTitle: gutil.replaceExtension(next, '')
```
sebastianbenz: You have to use FileName.toString(next) to unescape special chars.

tasks/lib/FileName.js
```diff
@@ -42,3 +42,56 @@ module.exports.toString = function(file) {
   fileName = gutil.replaceExtension(path.basename(fileName), '');
   return fileName.replace(/_/g, ' ');
 };
+
+/**
+* Return the next file in alphabetical orded
+*/
+module.exports.nextFile = function(file) {
```
sebastianbenz: can you please add a test for this in FileNameSpec?

tasks/lib/FileName.js
```diff
@@ -42,3 +42,56 @@ module.exports.toString = function(file) {
...
+    fileName = file;
+  } else if (typeof file.path === 'string') {
+    fileName = path.basename(file.path);
+  } else {
+    return '';
+  }
+  fileName = gutil.replaceExtension(path.basename(fileName), '');
+
+  var ampFiles = getAmpFilesFromDir('src')
+  ampFiles.sort(function(file1, file2){
```
sebastianbenz: I think it is save to use the build in sort function.

tasks/lib/FileName.js
```diff
@@ -42,3 +42,56 @@ module.exports.toString = function(file) {
...
+    var currentFileNameWithoutExtension = gutil.replaceExtension(path.basename(ampFiles[i]), '')
+    if (currentFileNameWithoutExtension == fileName) {
+      nextFileName = path.basename(ampFiles[++i]);
+    }
+  }
+  return nextFileName;
+};
+
+function getAmpFilesFromDir(dir) {
+  var filesystem = require("fs");
```
sebastianbenz: require should be at the top

tasks/lib/FileName.js
```diff
@@ -42,3 +42,56 @@ module.exports.toString = function(file) {
...
+
+    if (stat && stat.isDirectory()) {
+      //avoid other folders like img and template
+      if(path.basename(file).startsWith('amp')){
+        fileResults = fileResults.concat(getAmpFilesFromDir(file))
+      }
+    } else fileResults.push(file);
+  });
+  return fileResults;
+}
```
sebastianbenz: I think it is ok to ignore directories for now. Let that be my problem when I add the sections (and we know how they are structured. This is what an idiomatic findNext version in node could look like:

```
function nextFile(dir, currentFile) {
  const files = fs.readdirSync(dir).sort().filter(function(file) { 
    return path.extname(file) == '.html'; 
  });
  const pos = files.indexOf(currentFile);
  if (pos == -1) {
    return null;
  }
  return files[pos+1];
}
```

sebastianbenz:
Can you also please run gulp lint before committing?

kul3r4:
PTAL

tasks/compile-example.js
```diff
@@ -57,14 +57,21 @@ module.exports = function(templateRoot, template) {
       const document = DocumentParser.parse(contents);
       const stream = this;
       const title = FileName.toString(file);
+      let next = FileName.nextFile(file, 'src');
+      if (next) {
+        next = FileName.toString(next) + '.html';
```
sebastianbenz: Why do we need this case?

tasks/compile-example.js
sebastianbenz: use `nextTitle: FileName.toString(next)` instead

sebastianbenz:
Can you fix the amp validation errors? (travis build -> Details)

spec/compiler/FileNameSpec.js
```diff
@@ -47,6 +47,13 @@ describe("FileName", function() {
     });
   });
 
+  describe('nextFile', function() {
```
sebastianbenz: can you add another test for b.html, to make sure it currently returns undefined?

tasks/compile-example.js
sebastianbenz: otherwise samples with special chars will break, you can test it with the following sample:

```
gulp create -n "What's up?"
```

sebastianbenz:
I've just tried your PR locally, works well! 

I was wondering whether it would make sense to add the next link directly below the doc text on the left. What do you think?

tasks/compile-example.js
kul3r4: I got some undefined at some point and the toString doesn't handle that correctly.

sebastianbenz: I'd prefer to fix this in toString in that case. It should simply return an empty string if next is undefined.

kul3r4:
PTAL

tasks/compile-example.js
```diff
@@ -57,19 +57,28 @@ module.exports = function(templateRoot, template) {
...
+        next: next,
+        nextTitle: nextTitle
       };
+
       Metadata.add(args);
       // hack to avoid duplicate canonical refs as some examples define a canonical link
       if (document.head.indexOf('rel="canonical"') > -1) {
         args.skipCanonical = 'true';
       }
+      document.sections[document.sections.length - 1].appendDoc(
```
sebastianbenz: only add this if there is a next element, otherwise we have an empty: 'next up:'

tasks/lib/FileName.js
```diff
@@ -18,11 +18,15 @@
...
 const gutil = require('gulp-util');
 const path = require('path');
+const fs = require("fs");
 
 /**
  * Encodes a string into file system compatible representation.
  */
 module.exports.fromString = function(string) {
+  if (!string) {
+    return undefined;
```
sebastianbenz: nit: I'd prefer to return an empty string

tasks/lib/FileName.js
```diff
@@ -34,6 +38,9 @@ module.exports.fromString = function(string) {
  * Decodes a string from a file name.
  */
 module.exports.toString = function(file) {
+  if (!file) {
+    return undefined;
```
sebastianbenz: return empty string

tasks/lib/FileName.js
```diff
@@ -48,3 +55,21 @@ module.exports.toString = function(file) {
...
+  const files = fs.readdirSync(path.dirname(file))
+      .sort().filter(function(file) {
+        return path.extname(file) == '.html';
+      });
+  const pos = files.indexOf(path.basename(file));
+  if (pos == -1) {
+    return undefined;
+  }
+  //if it's the last one
+  if (pos == files.length - 1) {
```
sebastianbenz: you don't need this in JS, if the element doesn't exist it already returns undefined.

sebastianbenz:
Looks good - only small comments left.

kul3r4:
PTAL

sebastianbenz:
LGTM. Nice work Chiara!

PR #35, opened 2016-02-29
add optional validation task

jkingyens:
adds a "gulp validate" to invoke the runtime validator on each AMP example. If success output "PASS", otherwise output "FAIL" and show AMP validation errors from the javascript console.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -31,6 +31,7 @@ const favicons = require("gulp-favicons");
 const runSequence = require('run-sequence');
 
 const index = require('./tasks/compile-index');
+const validateExample = require('./tasks/validate-example');
 const compileExample = require('./tasks/compile-example');
 const createExample = require('./tasks/create-example');
 const FileName = require('./tasks/lib/FileName');
@@ -120,6 +121,12 @@ gulp.task("compile:favicons", function() {
       .pipe(gulp.dest("dist/favicons"));
 });
 
+gulp.task('validate:example', 'validate example html files', function() {
+  return gulp.src('src/*.html')
+    .pipe(compileExample('./src/templates/', 'example.html'))
+    .pipe(validateExample('./src/templates/', 'example.html'));
+});
+
 gulp.task('compile:example', 'compile example html files', function() {
   return gulp.src('src/*.html')
     .pipe(compileExample('./src/templates/', 'example.html'))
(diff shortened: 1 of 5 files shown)


gulpfile.js
```diff
@@ -120,6 +121,12 @@ gulp.task("compile:favicons", function() {
       .pipe(gulp.dest("dist/favicons"));
 });
 
+gulp.task('validate:example', 'validate example html files', function() {
+  return gulp.src('src/*.html')
+    .pipe(compileExample('./src/templates/', 'example.html'))
```
sebastianbenz: We could avoid duplicating the compileExample step by adding a dependency to the build (and clean) task. For a sample see the deploy task, which does the same.

tasks/validate-example.js
```diff
@@ -0,0 +1,91 @@
...
+        child.on('exit', function() {
+          let printedOutput = '';
+          const parsedOutput = JSON.parse(output);
+          const exampleKey = 'http://localhost:30000/' + inputFilename;
+          if (parsedOutput[exampleKey].success) {
+            printedOutput = 'PASS';
+          } else {
+            const errorList = parsedOutput[exampleKey].errors;
+            printedOutput = 'FAIL:\n';
+            errorList.forEach(function(item) {
```
sebastianbenz: idea: we could use gutils.colors.red to highlight errors (are there warnings as well?).

tasks/validate-example.js
```diff
@@ -0,0 +1,91 @@
...
+    if (file.isStream()) {
+      this.emit('error', new PluginError('validate-example',
+            'Streams not supported!'));
+    } else if (file.isBuffer()) {
+
+
+      // write file to disk, invoke validator, capture output & cleanup
+      const inputFilename = path.basename(file.path);
+      const tmpFile = os.tmpdir() + '/' + inputFilename;
+      fs.writeFile(tmpFile, file.contents, 'utf8', function(err) {
```
sebastianbenz: should we use the passed encoding instead of explicitly using utf8?

gulpfile.js
jkingyens: The validate task doesn't perform a build or touch the output directory. It just tests that the code in "src" will produce valid amp. If we can get validate to run fast (~ 1s) then we could insert it into the build.

sebastianbenz: What I meant was running validate after build:example, e.g.:

jkingyens: How do we know which files in dist need to be validated?

sebastianbenz: All AMP HTML files...

...but we should leave this for another PR.

sebastianbenz:
LGTM

PR #37, opened 2016-03-01
Switching arg parsing to yargs

paul-matthews:
* Included yargs library
* Used to get name for create task
* Used to get fix flag for lint


sebastianbenz:
LGTM

PR #38, opened 2016-03-01
Gulp server add port support

paul-matthews:
* Adding port support to serve task
* Adding docs explaining use

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -37,17 +37,19 @@ const compileExample = require('./tasks/compile-example');
 const createExample = require('./tasks/create-example');
 const FileName = require('./tasks/lib/FileName');
 
-gulp.task('serve', 'starts a local webserver', function() {
-  const server = gls.static('dist', 8000);
-  server.start();
-  gulp.watch(['dist/*.html',
-              'dist/img/*.{png,jpg,gif}',
-              'dist/video/*.{mp4,webm}'], function(file) {
-    /* eslint-disable */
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -36,17 +37,19 @@ const compileExample = require('./tasks/compile-example');
...
-  server.start();
-  gulp.watch(['dist/*.html',
-              'dist/img/*.{png,jpg,gif}',
-              'dist/video/*.{mp4,webm}'], function(file) {
-    /* eslint-disable */
-    server.notify.apply(server, [file]);
-    /* eslint-enable */
+gulp.task('serve', 'starts a local webserver (--port specifies bound port)',
+  function() {
+    const port = (argv.port) ? argv.port : 8000;
```
sebastianbenz: use `argv.port || 8000` instead

sebastianbenz:
one small comment - otherwise lgtm

gulpfile.js
```diff
@@ -163,7 +166,7 @@ gulp.task('test', function() {
 });
 
 gulp.task('lint', function() {
-  const hasFixFlag = (process.argv.slice(2).indexOf('--fix') >= 0);
+  const hasFixFlag = argv.fix ? true : false;
```
sebastianbenz: same as above - do we even need the false? (not an JS expert)

sebastianbenz:
LGTM

PR #43, opened 2016-03-02
Changed font and Removed bullet points

kul3r4:
Just few restyles, let me know if you like the idea.

PR #45, opened 2016-03-03
Creating Changed font and Removed bullet points again for discussion.

juliantoledo:
Reopening https://github.com/ampproject/amp-by-example/pull/43 for discussion.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -164,7 +164,7 @@ gulp.task('compile:index', 'generate index.html', function() {
 });
 
 gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
-  return gulp.src(SAMPLES_DIR + '*/*.html')
+  return gulp.src(paths.samples)
       .pipe(sitemap())
       .pipe(gulp.dest('dist'));
 });
@@ -187,7 +187,7 @@ gulp.task('create', 'create a new AMP example', function() {
   }
   return file(examplePath, '', {src: true})
       .pipe(createExample(paths.templates.dir, 'new-example.html'))
-      .pipe(gulp.dest(SAMPLES_DIR));
+      .pipe(gulp.dest('src'));
 });
 
 function throwInvalidArgumentError(message) {
--- a/templates/css/styles.css
+++ b/templates/css/styles.css
@@ -5,6 +5,7 @@ figure {
 /* layout */
 .body {
   background: #fff;
+  font-family: Roboto,sans-serif;
 }
 .box {
   min-height: -webkit-min-content;
@@ -15,6 +16,10 @@ figure {
 .column {
   flex: 1;
   width: 0;
+}
+
+ul {
+  list-style-type: none;
 }
 
 /* doc column */
--- a/templates/css/styles.css
+++ b/templates/css/styles.css
@@ -5,7 +5,11 @@ figure {
 /* layout */
 .body {
   background: #fff;
-  font-family: Roboto,sans-serif;
+  font-family: 'Roboto', 'Helvetica', 'Arial', sans-serif;
+}
+.body p a{
+  text-decoration: none;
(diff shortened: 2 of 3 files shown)


sebastianbenz:
//cc @kul3r4

src/templates/css/styles.css
```diff
@@ -17,6 +18,10 @@ figure {
   width: 0;
 }
 
+ul {
+  list-style-type: none;
```
sebastianbenz: This sets the default for all lists, which is something I would like to avoid. If we remove the bullet, we should align the list items with the surrounding text.

src/templates/css/styles.css
```diff
@@ -5,6 +5,7 @@ figure {
 /* layout */
 .body {
   background: #fff;
+  font-family: Roboto,sans-serif;
```
sebastianbenz: I was following this [template](https://www.getmdl.io/templates/article/index.html) which doesn't use Roboto for the text (only for headers). Otherwise LGTM. If we do this we should use `'Roboto','Helvetica','Arial',sans-serif;` Can you also check whether this changes the alignment between the lines in the code and doc sections?

kul3r4: We can continue following Material template you are following. I'll revert my change.

juliantoledo:
@sebastianbenz @kul3r4 Should I revert this one then?

juliantoledo:
the only problem is that we need that one soon, the project does not work right now.

kul3r4:
Result here: https://www.dropbox.com/s/voe14xzc9gh12p8/font-video.mov?dl=0

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
Can you upload the resulting amp file via dropbox?

kul3r4:
https://www.dropbox.com/s/k1jc5ut8vut186h/index.html?dl=0

kul3r4:
a normale file would look like this: https://www.dropbox.com/s/jhru9f1nbf51haz/index.html?dl=0

sebastianbenz:
I like it.

gulpfile.js
```diff
@@ -164,7 +164,7 @@ gulp.task('compile:index', 'generate index.html', function() {
 });
 
 gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
-  return gulp.src(SAMPLES_DIR + '*/*.html')
+  return gulp.src(paths.samples)
```
sebastianbenz: You have to rebase against master to include the fix

templates/css/styles.css
```diff
@@ -5,6 +5,11 @@ figure {
 /* layout */
 .body {
   background: #fff;
+  font-family: 'Roboto', 'Helvetica', 'Arial', sans-serif;
+}
+.body p a{
```
sebastianbenz: nit: space before opening brace

templates/css/styles.css
```diff
@@ -16,7 +21,14 @@ figure {
   flex: 1;
   width: 0;
 }
```
sebastianbenz: move the .example-list into the index.html template

sebastianbenz:
Added a few comments

sebastianbenz:
The different font size for links in the list in the index page doesn't look right. I'd prefer using the same font size, but increasing the line height.

kul3r4:
PTAL

templates/index.html
```diff
@@ -12,8 +12,16 @@
         padding: 0 16px;
       }
       ul {
-        padding-left: 18px;
+        padding-left: 0;
```
sebastianbenz: can you move this line into the class below?

kul3r4:
PTAL

sebastianbenz:
LGTM

Issue #32, opened 2016-02-26
Add metadata to examples


sebastianbenz:
Make it possible to add metadata to examples via special comment tag, e.g.

```
<!---{ 
  experiment: true,
  component: amp-accordion
}--->
```

Metadata can contain arbitrary JSON. The [DocumentParser](tasks/lib/DocumentParser.js) should extract the metadata and add it to a new field in [Document](tasks/lib/Document.js). 

This is the foundation for:

- #33 marking experimental samples
- #1 don't validate experimental samples

//cc @jkingyens @juliantoledo does anyone of you want to tackle this?

PR #51, opened 2016-03-03
Add metadata to examples #32

juliantoledo:


sebastianbenz:
Nice! Can you please add tests in DocumenParserSpec.

juliantoledo:
Sure, will do, I let you know when I have it.

juliantoledo:
@sebastianbenz PTAL

spec/compiler/DocumentParserSpec.js
```diff
@@ -138,6 +143,19 @@ describe("DocumentParser", function() {
     expect(doc.title).toEqual('hello');
   });
 
+  it("adds metadata to document befor comment", function() {
```
sebastianbenz: nit: typo

sebastianbenz: the idiomatic way would be to wrap these two tests in a describe block:

```
describe('adds metadata to document', function() {

  it("after comment", function() {
     ...
  }
it("before comment", function() {
     ...
  }
});
```

tasks/lib/DocumentParser.js
```diff
@@ -81,8 +90,14 @@ class DocumentParser {
...
       if (line.trim().endsWith('-->')) {
-        this.inComment = false;
-        this.currentTag = this.nextTag(i);
+        if (line.trim().endsWith('--->')) {
+          this.inMetadata = false;
+          this.document.metadata = JSON.parse(this.metadata);
+          this.currentTag = this.nextTag(i);
+        } else {
+          this.inComment = false;
+          this.currentTag = this.nextTag(i);
```
sebastianbenz: move `this.currentTag = this.nextTag(i);` in the outer diff so that we don't need to repeat it

tasks/lib/DocumentParser.js
```diff
@@ -81,8 +90,14 @@ class DocumentParser {
         }
       }
       if (line.trim().endsWith('-->')) {
-        this.inComment = false;
-        this.currentTag = this.nextTag(i);
+        if (line.trim().endsWith('--->')) {
+          this.inMetadata = false;
+          this.document.metadata = JSON.parse(this.metadata);
+          this.currentTag = this.nextTag(i);
```
sebastianbenz: I'm not sure we should record the nextTag here. This is only relevant if we start a new section. Better to remove this line.

spec/compiler/DocumentParserSpec.js
```diff
@@ -138,6 +143,19 @@ describe("DocumentParser", function() {
...
+    expect(doc.sections.length).toEqual(1);
+  });
+
+  it("adds metadata to document after comment", function() {
+    var doc = parse(COMMENT, DOCUMENT_METADATA_2, HEAD, TITLE, HEAD_END, BODY, COMMENT, BODY_END);
+    expect(doc.metadata.experiment).toEqual(true);
+    expect(doc.metadata.component).toEqual("amp-accordion");
+    expect(doc.sections.length).toEqual(4);
+  });
+
```
sebastianbenz: What happens if the JSON is invalid? Can we add a test for this case? I think it should fail loudly...

sebastianbenz:
nice work. left some comments

juliantoledo:
@sebastianbenz commits squashed

sebastianbenz:
LGTM - well done!

PR #55, opened 2016-03-04
Fix & refactorings for  gulpfile

sebastianbenz:
* extract all paths into global var
* fix watch for templates
* validate performs now all checks
* formattings


sebastianbenz:
@kul3r4 PTAL

PR #73, opened 2016-03-10
Use metadata to inject the experiment.html when "experiment": true

juliantoledo:


juliantoledo:
@sebastianbenz I don't know why, but when I inject experiment.html it does not rendered properly and the variables are empty. Could you take a look?

templates/example.html
```diff
@@ -13,9 +13,12 @@
   <body>
     <div class="body">
       {{> header.html}}
-      {{#sections}}
+      {{#data.sections}}
       <div class="box">
         <div class="column doc {{^inBody}}hide-on-mobile{{/inBody}}">
+          {{#data.experimental}}
+            {{> experiment.html}}
```
juliantoledo: @sebastianbenz here, it where I inject experiment.html and it does not render {{component}}, etc.

juliantoledo:
@sebastianbenz fixed the two questions I asked. PTAL

tasks/compile-example.js
```diff
@@ -60,6 +60,15 @@ module.exports = function(templateRoot, template) {
       const example = ExampleFile.fromPath(file.path);
       const nextExample = example.nextFile();
 
+      // Experimet must be injected in the first section only
```
sebastianbenz: nit: typo in Experiment

tasks/compile-example.js
```diff
@@ -67,7 +76,8 @@ module.exports = function(templateRoot, template) {
         github: example.githubUrl(),
         subHeading: example.title(),
         exampleStyles: document.styles,
-        sections: document.sections
+        component: example.title(),
```
sebastianbenz: why are you not using the component from the metadata?

juliantoledo: I was not sure, I will switch it.

templates/example.html
```diff
@@ -13,9 +13,12 @@
   <body>
     <div class="body">
       {{> header.html}}
-      {{#sections}}
+      {{#data.sections}}
```
sebastianbenz: why do we need the additional 'data' indirection?

juliantoledo: In order to pass:
        "experimental": function() {
          return (idx++ == 0 && document.metadata.experiment);
Experimental needs to be inside the first section.

tasks/compile-example.js
```diff
@@ -60,6 +60,15 @@ module.exports = function(templateRoot, template) {
       const example = ExampleFile.fromPath(file.path);
       const nextExample = example.nextFile();
 
+      // Experimet must be injected in the first section only
+      let idx = 0;
+      const data = {
+        "sections": document.sections,
+        "experimental": function() {
+          return (idx++ == 0 && document.metadata.experiment);
```
sebastianbenz: Can we rename `experimental` to `inExperiment`? This makes it clear that it is a boolean flag.

juliantoledo: sure.

templates/example.html
sebastianbenz: @juliantoledo understood, but couldn't we just add the function as a field in the args object?

jkingyens:
I have a commit based on this branch that passes this experimental flag to validation task and skips validation for experimental examples. I can wait for this PR to land in master or add the commit here.

juliantoledo:
@jkingyens Sounds good, the metadata is already supported in the document parser. I will finish this PR in a bit.

juliantoledo:
@sebastianbenz fixed your comments and added the metadata instructions to the readme.

tasks/compile-example.js
```diff
@@ -67,9 +66,17 @@ module.exports = function(templateRoot, template) {
...
         subHeading: example.title(),
         exampleStyles: document.styles,
-        sections: document.sections
+        component: document.metadata.component,
+        sections: document.sections,
+        isExperiment: document.metadata.experiment
       };
 
+      if (document.metadata.experiment && (!document.metadata.component ||
+        document.metadata.component.length === 0)) {
```
sebastianbenz: `if (document.metadata.experiment && !document.metadata.component)` is enough"

```
> false == ''
true
```

juliantoledo: done.

tasks/compile-example.js
```diff
@@ -67,9 +66,17 @@ module.exports = function(templateRoot, template) {
...
+        throw new Error("Example (" + example.url() +
```
sebastianbenz: throw a gulp exception: `throw new PluginError(plugin: 'compile-example', message: '...');`

sebastianbenz:
LGTM - good work!

PR #77, opened 2016-03-11
skip validation for examples flagged as experiments

jkingyens:
When running "gulp validate", the task will now skip over any examples that are marked as experimental since in many cases we expect them to fail.


jkingyens:
@sebastianbenz @juliantoledo PTAL thanks!

sebastianbenz:
LGTM. Nice trick to attach the experimental flag to the file object.

PR #78, opened 2016-03-11
add setting to skip validation. fail validation fails build.

jkingyens:
This adds a flag to skip validation. This is used to skip amp-list which currently exposes a validation bug right now.

### DIFF
--- a/src/20_Components/amp-anim.html
+++ b/src/20_Components/amp-anim.html
@@ -1,7 +1,7 @@
 <!--
   #### Introduction
 
-  Embed animations (gif, webp) into your AMP HTML files. Animation files should be loaded via HTTPS. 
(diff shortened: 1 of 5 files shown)


jkingyens:
I intentionally submitted PR without the amp-list skip first so we could see the build fail and then pass after the fix (see above). @sebastianbenz @juliantoledo PTAL thanks!

sebastianbenz:
We currently duplicate knowledge about the different validation flags across validate-example and compile-example. If we simply attach the metadata object to the file object instead, compile-example no longer needs to know about skipValidation.

jkingyens:
Good idea. Fixed. PTAL thanks!

sebastianbenz:
LGTM

PR #80, opened 2016-03-14
handle escaped links

sebastianbenz:
* go backend accepts paths with and without trailing slash
* all special chars are removed from urls


sebastianbenz:
@juliantoledo PTAL

juliantoledo:
LGTM

PR #97, opened 2016-03-15
output ignored test status as yellow warning

jkingyens:
Instead of silently skipping ignored tests, the results log now reports the skipped test in yellow (as a warning). This will not trigger a build failure and is just for information purposes only.


jkingyens:
@sebastianbenz PTAL thanks!

sebastianbenz:
Yay!

PR #115, opened 2016-03-29
Make analytics dependency explicit

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

tasks/lib/Document.js
```diff
@@ -74,6 +74,13 @@ module.exports = class Document {
     return this.head.indexOf('rel="canonical"') > -1;
   }
 
+  /**
+   * Returns true if the document imports the given component
+   */
+  importsComponent(componentName) {
```
kul3r4: Rename to importComponent (similarly to the method below which does not add an s).
Also, from the name it seems that the responsibility is to "import the component" while we only want to check if the component has been imported. Maybe "isComponentImported" may be more appropriate?

PR #137, opened 2016-04-21
Support incomplete tags

sebastianbenz:
Example:

```
<!-- Sample -->
<amp-img
    src="..."></amp-img>
```


sebastianbenz:
@juliantoledo PTAL

juliantoledo:
LGTM, By incomplete tags we mean tags that continue in the following lines (broken into multiple lines).

Issue #138, opened 2016-04-25
<amp-list> example generates four lists of one item each


ithinkihaveacat:
The [`<amp-list> example`](https://ampbyexample.com/components/amp-list/) generates four lists of one item each:

````html
<ul><li>...</li></ul>
<ul><li>...</li></ul>
<ul><li>...</li></ul>
<ul><li>...</li></ul>
````

Is that the intention?

An alternative would be to make `<ul></ul>` the parent of `<amp-list>`, which would generate one list of four items.

PR #139, opened 2016-04-25
Make <ul> the parent of <amp-list>

ithinkihaveacat:
Generate one list of four items, instead of four lists of one item each.

Not tested; to illustrate #138.


sebastianbenz:
Good find Michael. Thanks! Tested.

PR #144, opened 2016-04-26
Updating the CONTRIBUTING.md to include the squash commit best practice

kul3r4:
Making clearer that we require to squash all the commits into one when reviewing a pull request.
@sebastianbenz  PTAL


sebastianbenz:
LGTM

cramforce:
This repository has the new "Allow squash merging" option turned on that makes the Merge button have the option of doing the squashing, so this should not longer be needed (Thankfully!!!!).

juliantoledo:
That would be very cool! we should try it!

PR #151, opened 2016-05-02
Make mobile great again

sebastianbenz:
New formatting for mobile view:

* show code sections if they have a comment
* show preview sections if they have a comment
* hide sections with empty docs
* show single column view for screen < 1024px
* various formatting fixes for samples

This required changing lots of samples to add an empty comment
after the introduction. The intro is usually not associated to a
code blocks and otherwise the first html tags will be shown after
introduction.


sebastianbenz:
Current version can be seen [here](https://amp-by-example-staging.appspot.com/). 

@adewale @kul3r4 @juliantoledo PTAL

What doesn't work yet is showing code blocks spanning multiple separate tags, e.g.:

```
  <!--
    Import the libraries.
  -->
  <script async custom-element="amp-instagram" src="https://cdn.ampproject.org/v0/amp-instagram-0.1.js"></script>
  <script async custom-element="amp-vine" src="https://cdn.ampproject.org/v0/amp-vine-0.1.js"></script>
  <script async custom-element="amp-youtube" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>
  <script async custom-element="amp-vimeo" src="https://cdn.ampproject.org/v0/amp-vimeo-0.1.js"></script>
...
```

will only show the first line of the imports. In this case it would make more sense to show all imports. 

```
    Import the libraries.
  <script async custom-element="amp-instagram" src="https://cdn.ampproject.org/v0/amp-instagram-0.1.js"></script>
```

To solve this we would need an explicit way to define doc comment spans :-/ , e.g.:

```
  <!--
    Import the libraries.
  -->
<!--BEGIN-->
  <script async custom-element="amp-instagram" src="https://cdn.ampproject.org/v0/amp-instagram-0.1.js"></script>
  <script async custom-element="amp-vine" src="https://cdn.ampproject.org/v0/amp-vine-0.1.js"></script>
  <script async custom-element="amp-youtube" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>
  <script async custom-element="amp-vimeo" src="https://cdn.ampproject.org/v0/amp-vimeo-0.1.js"></script>
<!--END-->

**Comment by adewale** [comment]
LGTM expect that I couldn't work out how to get a preview of the page without the code.

**Comment by kul3r4** [comment]
Hi, I was going through all the samples and  I noticed that https://amp-by-example-staging.appspot.com/advanced/long_list_of_amp-instagram_embeds/ does not work.

**Comment by sebastianbenz** [comment]
@adewale we haven't enabled the preview for all samples, IMO it only works well for more complex samples. The lack of consistency might be a problem though.

@kul3r4 good find. The algorithm doesn't cover this scenario. A preview is only shown for sections with a comment. This usually works well. Except for examples were the preview is important, but not the code as in this case. Any ideas how we could handle these?

**Comment by sebastianbenz** [comment]
Have been playing around with the mobile layout:

* toolbar
* FAB for Preview (does this make sense on desktop? )

What do you think?

Screenshot

**Comment by sebastianbenz** [comment]
I've created two new samples versions for desktop and mobile:

* [desktop](https://dl.dropboxusercontent.com/u/7234606/desktop.png)
* [mobile](https://dl.dropboxusercontent.com/u/7234606/mobile.png)

What do you think?

**Comment by sebastianbenz** [comment]
@kul3r4 @juliantoledo PTAL

I've updated the PR with a fix for the long list of instagram embeds. I solved it using a simple hack:

```
 <!-- &nbsp; -->
  <amp-instagram
      data-shortcode="BAb2gbSRJaM"
      width="1"
      height="1"
      layout="responsive">
  </amp-instagram>
```
The HTML space character makes sure that the code section + preview section is generated even if there is no doc. It's a hack, but I'd be fine with it as long as it stays an exception.

I've deployed this version to staging:

https://amp-by-example-staging.appspot.com/advanced/long_list_of_amp-instagram_embeds/

adewale:
Please add a comment to the `<!-- &nbsp; -->` so that people know why it's there.

I'm, personally, not a fan of FABs especially on desktop where it's much harder to hit the target area. Can we find another solution?

kul3r4:
Hi, thanks for fixing the example.
Regarding the preview button, what about rolling out to production and measuring via amp-analytics, if it gets clicked?

juliantoledo:
I don't see the toolbars, I guess that is still not in here.

I still find it a bit confusing on mobile and the white space makes the page too big. What about using MD cards?

sebastianbenz:
I removed the header design changes from this PR and will create another one for them. This PR only fixes the content of the mobile site + adds a FAB for the preview for the mobile version.

@adewale I've added a comment for the hack (had to misuse metadata for this). 
@juliantoledo which whitespace is too big? Where do you want to use cards?

juliantoledo:
The white space between the element and the code: 
https://goo.gl/photos/fRyzk8mVUVb3odXa7

If it is not too difficult, we could group the code with the element it generates on a card. That would make clear which code belongs to what element.

sebastianbenz:
It think it works good enough for most samples. The instagram one is an exception because the empty comment hack generates more whitespace - I don't see a way around that at the moment. 

Putting both on cards is not trivial as they are contained in different `div` blocks.

juliantoledo:
LGTM

Cool, we can revisit later if we see the need somewhere else.

sebastianbenz:
Thanks for the feedback everyone!

PR #130, opened 2016-04-16
add amp-install-serviceworker example

jmadler:
Resolves #119

PR #156, opened 2016-05-12
add working example of amp-install-serviceworker

ymotongpoo:
Though @jmadler has already created template example of amp-install-serviceworker, as discussed in #130, I created working example of it using sw-precache as a practice of using it.

For PoC, I deployed it in https://yoshifumi-amp-demo.firebaseapp.com/components/amp-install-serviceworker/ On accessing here, the SW starts caching contents under `img` and `video`. (To install SW, you need to access there twice.)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -189,9 +189,11 @@ gulp.task('compile:sw-precache',
 
             swPrecache.write(path.join(staticDir, 'sw.js'), {
               staticFileGlobs: [
-                path.join(paths.dist.img, '*.{png,jpg,gif}'),
-                path.join(paths.dist.video, '*.{mp4,webm}'),
-                paths.dist.html
+                path.join(paths.dist.dir, 'LICENSE.txt'),
+                path.join(paths.dist.img, 'gist.png'),
+                path.join(paths.dist.img, 'abe_preview.png'),
+                path.join(paths.dist.favicons, '*.png'),
+                path.join(paths.dist.dir, 'components/amp-install-serviceworker/*.html')
               ],
               stripPrefix: 'dist',
               verbose: true
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -185,9 +185,7 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
 
 gulp.task('compile:sw-precache',
   ['copy:images', 'copy:videos', 'compile:example'], function() {
-    const staticDir = 'static';
-
-    swPrecache.write(path.join(staticDir, 'sw.js'), {
+    swPrecache.write(path.join(paths.dist.dir, 'sw.js'), {
       staticFileGlobs: [
         path.join(paths.dist.dir, 'LICENSE.txt'),
         path.join(paths.dist.img, 'gist.png'),
@@ -199,28 +197,6 @@ gulp.task('compile:sw-precache',
       stripPrefix: 'dist',
       verbose: true
     });
-  });
-
-gulp.task('copy:sw-precache', ['compile:sw-precache'], function() {
-  const staticDir = 'static';
-
-  return gulp.src(path.join(staticDir, 'sw.js'))
-    .pipe(cache(staticDir))
-    .pipe(gulp.dest(paths.dist.dir));
-});
-
-gulp.task('clean:sw-precache', 'delete temporary file', function() {
-  const staticDir = 'static';
-
-  return del([path.join(staticDir, 'sw.js')]);
-});
-
-gulp.task('sw-precache', 'generate sw.js and clean temporary file',
-  function(callback) {
-    runSequence('compile:sw-precache',
-                'copy:sw-precache',
-                'clean:sw-precache',
-                callback);
   });
 
 gulp.task('create', 'create a new AMP example', function() {
@@ -327,7 +303,7 @@ gulp.task('build', 'build all resources', [
   'compile:favicons',
   'compile:sitemap',
   'compile:example',
-  'sw-precache']);
+  'compile:sw-precache']);
 
 function isFixed(file) {
   return file.eslint.fixed;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -185,22 +185,22 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
 
 gulp.task('compile:sw-precache',
           ['copy:images', 'copy:videos', 'compile:example'], function() {
-  const staticDir = 'static';
+            const staticDir = 'static';
 
-  swPrecache.write(path.join(staticDir, 'sw.js'), {
-    staticFileGlobs: [
-      path.join(paths.dist.img, '*.{png,jpg,gif}'),
-      path.join(paths.dist.video, '*.{mp4,webm}'),
-      paths.dist.html
-    ],
-    stripPrefix: 'dist',
-    verbose: true
-  });
+            swPrecache.write(path.join(staticDir, 'sw.js'), {
+              staticFileGlobs: [
+                path.join(paths.dist.img, '*.{png,jpg,gif}'),
+                path.join(paths.dist.video, '*.{mp4,webm}'),
+                paths.dist.html
+              ],
+              stripPrefix: 'dist',
+              verbose: true
+            });
 
-  return gulp.src(path.join(staticDir, 'sw.js'))
-      .pipe(cache('static'))
-      .pipe(gulp.dest(paths.dist.dir));
-});
+            return gulp.src(path.join(staticDir, 'sw.js'))
+              .pipe(cache('static'))
+              .pipe(gulp.dest(paths.dist.dir));
+          });
 
 gulp.task('create', 'create a new AMP example', function() {
   const title = argv.n || argv.name;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -183,16 +183,22 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
       .pipe(gulp.dest(paths.dist.dir));
 });
 
-gulp.task('compile:sw-precache', 'run sw-precache process', function() {
+gulp.task('compile:sw-precache',
+          ['copy:images', 'copy:videos', 'compile:example'], function() {
   const staticDir = 'static';
 
   swPrecache.write(path.join(staticDir, 'sw.js'), {
-    staticFileGlobs: [paths.images, paths.videos],
-    stripPrefix: 'src'
+    staticFileGlobs: [
+      path.join(paths.dist.img, '*.{png,jpg,gif}'),
+      path.join(paths.dist.video, '*.{mp4,webm}'),
+      paths.dist.html
+    ],
+    stripPrefix: 'dist',
+    verbose: true
   });
 
-  return gulp.src('static/sw.js')
-      .pipe(cache('static/sw.js'))
+  return gulp.src(path.join(staticDir, 'sw.js'))
+      .pipe(cache('static'))
       .pipe(gulp.dest(paths.dist.dir));
 });
 
@@ -299,8 +305,8 @@ gulp.task('build', 'build all resources', [
   'copy:static',
   'compile:favicons',
   'compile:sitemap',
-  'compile:sw-precache',
-  'compile:example']);
+  'compile:example',
+  'compile:sw-precache']);
 
 function isFixed(file) {
   return file.eslint.fixed;
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -193,14 +193,15 @@ gulp.task('compile:sw-precache',
                 path.join(paths.dist.img, 'gist.png'),
                 path.join(paths.dist.img, 'abe_preview.png'),
                 path.join(paths.dist.favicons, '*.png'),
-                path.join(paths.dist.dir, 'components/amp-install-serviceworker/*.html')
+                path.join(paths.dist.dir,
+                   'components/amp-install-serviceworker/*.html')
               ],
               stripPrefix: 'dist',
               verbose: true
             });
 
             return gulp.src(path.join(staticDir, 'sw.js'))
-              .pipe(cache('static'))
+              .pipe(cache(staticDir))
               .pipe(gulp.dest(paths.dist.dir));
           });
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -184,26 +184,44 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
 });
 
 gulp.task('compile:sw-precache',
-          ['copy:images', 'copy:videos', 'compile:example'], function() {
-            const staticDir = 'static';
+  ['copy:images', 'copy:videos', 'compile:example'], function() {
+    const staticDir = 'static';
 
-            swPrecache.write(path.join(staticDir, 'sw.js'), {
-              staticFileGlobs: [
-                path.join(paths.dist.dir, 'LICENSE.txt'),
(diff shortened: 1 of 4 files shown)


src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,41 @@
...
+  amp-install-serviceworker should have both properties `src` and `data-iframe-src`; `src` is the location of Service Worker to be installed when the original AMP pages is accessed, `data-iframe-src` is the location of HTML file which calls Service Worker installation process when the cached AMP pages is accessed.
+  The reason why the latter option is required is because of Service Worker's security protection function. For details of the security considerations of Service Worker, [see official specs](https://www.w3.org/TR/service-workers/#security-considerations). 
+  -->
+  <amp-install-serviceworker
+      src="https://ampbyexample.com/sw.js"
+      data-iframe-src="https://ampbyexample.com/sw.html"
+      layout="nodisplay">
+  </amp-install-serviceworker>
+  <!--
+  `sw.js` here is generated using [sw-precache](https://github.com/GoogleChrome/sw-precache), which generates a Service Worker JavaScript file that caches specifiled static files. See example in the repository of `sw-precache` and [gulpfile.js](https://github.com/ampproject/amp-by-example/blob/master/gulpfile.js) for how to use it.
```
sebastianbenz: * typo: s/specifiled/specified
* Can you describe what the effect of this SW is and how users can see it working?

sebastianbenz:
Great! Stupid question: what happens if the SW is installed, will it run for all ABE pages?  Could we instead cache all resources (inkl. html) only for the amp-install-service-worker sample? This way users could test the SW by disconnecting from their network and reloading the page.

ymotongpoo:
I added comments on your questions. This SW will be installed only from the demo page of amp-install-serviceworker and it caches all image files and video files once it is installed.

> Could we instead cache all resources (inkl. html) only for the amp-install-service-worker sample?
Let me try if it works on my test server.

ymotongpoo:
ok, changed Service Worker setting to cache all images, videos and HTML files. It works for all demo pages and the demos works when offline.

sebastianbenz:
What happens if an article is updated? Will the cache be automatically invalidated? Otherwise we'd have a problem. That's why I'd prefer to only cache the sw sample.

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,47 @@
...
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+
+</head>
+<body>
+
+  <!-- #### Basic Usage -->
+  <!--
+  amp-install-serviceworker should have both properties `src` and `data-iframe-src`; `src` is the location of Service Worker to be installed when the original AMP pages is accessed, `data-iframe-src` is the location of HTML file which calls Service Worker installation process when the cached AMP pages is accessed.
+  The reason why the latter option is required is because of Service Worker's security protection function. For details of the security considerations of Service Worker, [see official specs](https://www.w3.org/TR/service-workers/#security-considerations).
```
jmadler: it might make sense to clarify a bit about why cached amp pages undergo this additional security restriction.  e.g. "is the location of the HTML file which calls the Service Worker installation process when the AMP page is accessed from a third-party origin, such as the Google AMP Cache"

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,47 @@
...
+  <script async custom-element="amp-install-serviceworker" src="https://cdn.ampproject.org/v0/amp-install-serviceworker-0.1.js"></script>
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+
+</head>
+<body>
+
+  <!-- #### Basic Usage -->
+  <!--
```
jmadler: Does it make sense to talk a bit about Service Workers and why someone might want to install one?

sebastianbenz: +1

jmadler:
Hey Yoshi!  Feel free to grab some of the text from my PR.  I'll close it after we commit this one.

ymotongpoo:
Thanks for comments Sebastian and Jordan! As first step, I'll change the commit as follows:
1. Add more background derived from @jmadler 's last commit in #130 
2. Shrink the caching scope only to amp-install-serviceworker, which may require elimination of usage of sw-precache and manual creation of sw.js.

I personally think caching larger objects such as image and videos makes difference on rendering pages, but I agree that it depends on the frequency of contents updates. Once I find better way to control cache update management, I hope to expand the caching scope to other demo pages so that users can experience the power of combination of AMP & Cache API.

Does this make sense to you guys?

jmadler:
That sounds like a great approach.  I think having a self-encapsulated example for amp-install-serviceworker is useful.  It might also be useful to have the SW send console messages that explain the kind of caching behavior a SW could do, since you'll be doing less actual caching now :)

Keep in mind also that we may want to actually make a SW for the overall ABE site at some point to enable offline usage and push notifications, so make sure the SW's scope is limited to this example and not the directory.

ymotongpoo:
I added comments on the benefits of Service Worker and shrunk the scope of caching within amp-install-serviceworker demo. Unfortunately, the last commit from @jmadler was deleted and I couldn't refer to that.

jmadler:
I can issue a PR to add further text after this PR is merged.

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,48 @@
...
+  The reason why the latter option is required is because browsers can't install Service Worker from other domains for security reasons. For details of the security considerations of Service Worker, [see official specs](https://www.w3.org/TR/service-workers/#security-considerations).
+  -->
+  <amp-install-serviceworker
+      src="https://ampbyexample.com/sw.js"
+      data-iframe-src="https://ampbyexample.com/sw.html"
+      layout="nodisplay">
+  </amp-install-serviceworker>
+  <!--
+  `sw.js` here is generated using [sw-precache](https://github.com/GoogleChrome/sw-precache), which generates a Service Worker JavaScript file that caches specified static files. See example in the repository of `sw-precache` and [gulpfile.js](https://github.com/ampproject/amp-by-example/blob/master/gulpfile.js) for how to use it. In this sample, when users access this AMP page for the 2nd time, the Service Worker is installed and start caching all image files and video files under root path of this domain.
```
sebastianbenz: link "Service Worker JavaScript file" to "/sw.js"

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,48 @@
...
+  
+  In order to confirm if the Service Worker is installed and it is working as expected, follow the procedure below:
+  1. Access to https://ampbyexample.com/components/amp-install-serviceworker/ twice.
+  2. Open Chrome Dev Tools and navigate to "Resources" tab. You may see "Service Workers" section and see `/sw.js` is installed.
+  3. Then open "Cached Storage". You'll find a lot of files whose names start with "sw-precache-v1--", which are cached images and videos.
+  4. Try accessing amp-install-serviceworker when you are offline. On "Network" tab in Dev Tools, you'll find "(from ServiceWorker)" notation in the rows of image/video file fetch events.
```
sebastianbenz: s/Try accessing amp-install-serviceworker when you are offline./Disable your network connection and reload this page. It still works!/

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,48 @@
...
+<body>
+
+  <!-- #### Basic Usage -->
+  <!--
+  amp-install-serviceworker should have both properties `src` and `data-iframe-src`; `src` is the location of Service Worker to be installed when the original AMP pages is accessed, `data-iframe-src` is the location of HTML file which calls Service Worker installation process when the cached AMP pages is accessed.
+  The reason why the latter option is required is because browsers can't install Service Worker from other domains for security reasons. For details of the security considerations of Service Worker, [see official specs](https://www.w3.org/TR/service-workers/#security-considerations).
+  -->
+  <amp-install-serviceworker
+      src="https://ampbyexample.com/sw.js"
+      data-iframe-src="https://ampbyexample.com/sw.html"
```
sebastianbenz: Use relative links so that it also works in development mode on localhost.

sebastianbenz:
Sorry for the delayed feedback. I've added some more comments. Also: I've tried running it locally, but offline mode didn't work even though the sw has been installed. It would also be nice to have a console message when the sw is installed.

src/20_Components/amp-install-serviceworker.html
ymotongpoo: Thanks. Fixed.

gulpfile.js
```diff
@@ -182,6 +183,28 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
...
+            const staticDir = 'static';
+
+            swPrecache.write(path.join(staticDir, 'sw.js'), {
+              staticFileGlobs: [
+                path.join(paths.dist.dir, 'LICENSE.txt'),
+                path.join(paths.dist.img, 'gist.png'),
+                path.join(paths.dist.img, 'abe_preview.png'),
+                path.join(paths.dist.favicons, '*.png'),
+                path.join(paths.dist.dir,
+                   'components/amp-install-serviceworker/*.html')
```
sebastianbenz: Can we cache Google fonts?

src/20_Components/amp-install-serviceworker.html
ymotongpoo: I added some comments, but @jmadler will add more and sophisticate after this p-r.

sebastianbenz:
Sorry for the delayed feedback. I've added some more comments. Also: I've tried running it locally, but offline mode didn't work even though the sw has been installed. It would also be nice to have a console message when the sw is installed.

src/20_Components/amp-install-serviceworker.html
ymotongpoo: I like it 😄

gulpfile.js
ymotongpoo: Do you mean fonts under https://fonts.googleapis.com/ ? Service Worker can't access resources out side of its scope.

src/20_Components/amp-install-serviceworker.html
ymotongpoo: changed path for `src` in relative path, but left `data-iframe-src` in absolute path since sw.html won't be cached on cdn.ampproject.org.

gulpfile.js
sebastianbenz: Ok - too bad.

ymotongpoo:
> It would also be nice to have a console message when the sw is installed.
That can be achieved in the JS which is installing the service worker, not in the service worker JS itself. So it should be done in amp-install-serviceworker itself. More precisely, in this function. https://github.com/ampproject/amphtml/blob/master/extensions/amp-install-serviceworker/0.1/amp-install-serviceworker.js#L109

sebastianbenz:
> That can be achieved in the JS which is installing the service worker, not in the service worker JS itself. 

Too bad. 

Am I doing something wrong, that the offline support doesn't work?

jmadler:
@ymotongpoo You should be able to log a console message in the SW's event handler for the 'install' event

ymotongpoo:
Do we really need to show the log using `console.log`? In order to confirm if the service worker and target contents are cached in the browser, we need to access devtools, and in the case of seeing `console.log`, we open up devtools anyway. If adding `console.log` is the necessary requirement, using sw-precache is not fitting to the use, and I would rewrite sw.js from the scratch. (I checked option of sw-precache, and creating new template is required to add own event listener for `install` event)

sebastianbenz:
@ymotongpoo depends on how much effort it is. It would be nice, but we should spend too much time on it.

ymotongpoo:
@sebastianbenz I think current implementation is good to go, but it is all up to you. Rewriting all takes a couple of days more.

sebastianbenz:
Let's start with sw-precache then. One last thing: can you generate sw.js directly into the dist folder? Otherwise we need to add it to the .gitignore.

ymotongpoo:
changed gulpfile.js to delete temporarily generate file in static dir.

gulpfile.js
```diff
@@ -182,6 +183,46 @@ gulp.task('compile:sitemap', 'generate sitemap.xml', function() {
...
+      staticFileGlobs: [
+        path.join(paths.dist.dir, 'LICENSE.txt'),
+        path.join(paths.dist.img, 'gist.png'),
+        path.join(paths.dist.img, 'abe_preview.png'),
+        path.join(paths.dist.favicons, '*.png'),
+        path.join(paths.dist.dir,
+          'components/amp-install-serviceworker/*.html')
+      ],
+      stripPrefix: 'dist',
+      verbose: true
```
sebastianbenz: we should disable the verbose mode

sebastianbenz:
Why can't we directly generate sw.js into the dist folder. This way we wouldn't need a cleanup task.

ymotongpoo:
Agreed. I did something wrong at that time. Now I made the process more simpler and `gulp build` generates `sw.js` directly into `dist` dir.

sebastianbenz:
LGTM

src/20_Components/amp-install-serviceworker.html
```diff
@@ -0,0 +1,48 @@
...
+  -->
+  <amp-install-serviceworker
+      src="/sw.js"
+      data-iframe-src="https://ampbyexample.com/sw.html"
+      layout="nodisplay">
+  </amp-install-serviceworker>
+  <!--
+  <p class="card warning">`sw.js` here is generated using [sw-precache](https://github.com/GoogleChrome/sw-precache), which generates a [Service Worker JavaScript file](https://ampbyexample.com/sw.js) that caches specified static files. See example in the repository of `sw-precache` and [gulpfile.js](https://github.com/ampproject/amp-by-example/blob/master/gulpfile.js) for how to use it.</p> In this sample, when users access this AMP page for the 2nd time, the Service Worker is installed and start caching all image files and video files under root path of this domain.
+  
+  In order to confirm if the Service Worker is installed and it is working as expected, follow the procedure below:
```
ithinkihaveacat: Note that the iframe is injected [after a 20 secs delay](https://github.com/ampproject/amphtml/blob/master/extensions/amp-install-serviceworker/0.1/amp-install-serviceworker.js#L83), so if someone is really speedy they won't see the service worker.

sebastianbenz:
@ithinkihaveacat Thanks! I've added this now [here](https://github.com/ampproject/amp-by-example/pull/176/files).

PR #173, opened 2016-06-01
New drawer menu

juliantoledo:
Still early and WIP, @sebastianbenz can you take a quick first look?

I am fixing the content of the sidebar, is empty in the example pages and only populated in the index page.


sebastianbenz:
Great! I'm surprised that two sidebars work at the same time.

juliantoledo:
Yes, I was also surprise, it was the first think I checked.

I also need to fix the scrolling, it currently scrolls both in the page and the sidebar at the same time.

sebastianbenz:
Two things:

* we should follow the [MD guidelines](https://www.google.com/design/spec/patterns/navigation-drawer.html)
* I'm working on a PR for a responsive header redesign. You could use this as well.

juliantoledo:
Cool. I should stop working on this header and focus only on the sidebar then.

juliantoledo:
@sebastianbenz I made some progress, I also somehow broke the alignment of your "View on Github" title, can you find what is wrong?

tasks/compile-example.js
```diff
@@ -84,12 +84,57 @@ module.exports = function(config, updateTimestamp) {
...
+  }
+
+  function endStream(cb) {
+    // no examples passed in, no file goes out
+    if (!latestFile || !examples) {
+      cb();
+      return;
+    }
+
+    const categories = mapToCategories(examples);
```
sebastianbenz: Can you extract a method `compileIndex` as well so that the order is:

```
const stream = this;
const categories = mapToCategories(examples);
compileExamples(examples, categories, stream);
compileIndex(categories, stream);
```

tasks/compile-example.js
```diff
@@ -139,61 +180,31 @@ module.exports = function(config, updateTimestamp) {
...
-    gutil.log('Generated ' + indexFile.relative);
-    stream.push(indexFile);
-    cb();
+    });
   }
 
   function mapToCategories(examples) {
     const categories = [];
     let currentCategory;
     sort(examples).forEach(function(exampleFile) {
```
sebastianbenz: more idiomatic:

```
sort(examples)
.filter(example => !exampleFile.document.metadata.draft)
.forEach(...){
```

sebastianbenz:
Looks good! Some comments:

* can you use the same drawer style as https://material.google.com/
* add a mouse over effect to the close drawer button
* suggestion: can we try adding "AMP by Example" after the home icon? Or even replace the home icon?

sebastianbenz:
Regarding "View on Github", add a `float: right;`:

```
nav .button > span {
  float: right;
  ...
```

tasks/compile-example.js
juliantoledo: done

tasks/compile-example.js
juliantoledo: done

juliantoledo:
Fixed most of the comments, I still need to work on making active the current category/example.

I just noticed that when clicking at a sidebar link and pressing back in the browser it goes back to http://localhost:8000/# and the sidebar stops working.

sebastianbenz:
Have you filed a bug report?

sebastianbenz:
Seems like your commit got a bit messed up and includes previous changes.

juliantoledo:
Continuing work at https://github.com/ampproject/amp-by-example/pull/200

sriramkrish85:
FYI : @sebastianbenz @juliantoledo , letting multiple sidebars open at the same time is a bug and here is how it would be fixed - https://github.com/ampproject/amphtml/issues/3320

PR #178, opened 2016-06-06
Add example for amp-live-list

kul3r4:
The html page contains a template so that the server can add new elements.
I am showing 2 cases:

- insertion of new elements

- update of existing items


kul3r4:
@sebastianbenz Cookie logic is added

kul3r4:
@sebastianbenz Added cookie logic to show only one item when landing on the page the first time.
PTAL.

server.go
```diff
@@ -20,10 +20,12 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time" 
+	"strconv"
 )
 
 const (
-	MAX_AGE_IN_SECONDS = 60 * 60 * 24 // 1 day
+	MAX_AGE_IN_SECONDS = 1 // 1 second
```
sebastianbenz: the comment is redundant

server.go
```diff
@@ -50,10 +52,22 @@ var REDIRECTS [18][2]string = [18][2]string{
...
 }
 
+type BlogItem struct {
+	Text string
+	Image string
+	Timestamp string
+	Time string
+}
+
+var ArrayOfBlogs map[int]BlogItem
```
sebastianbenz: don't encode type in variable names:

ArrayOfBlogs -> blogs

server.go
```diff
@@ -67,6 +81,59 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	ArrayOfBlogs[3] = BlogItem{Text: "A trees landscape", Image: "/img/landscape2.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[4] = BlogItem{Text: "A village landscape", Image: "/img/landscape3.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[5] = BlogItem{Text: "A canyon", Image: "/img/landscape4.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[6] = BlogItem{Text: "A dog", Image: "/img/Shetland_Sheepdog.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[7] = BlogItem{Text: "A phone", Image: "/img/silver_nexus.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[8] = BlogItem{Text: "Some Sushi", Image: "/img/image3.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[9] = BlogItem{Text: "Hi Gopher!", Image: "/img/gopher.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+}
```
sebastianbenz: why do you use a map? Can't you simply use an array/slice? These automatically have numbers...

server.go
```diff
@@ -67,6 +81,59 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+		for i := 0; i < size; i++ {
+			temp := ArrayOfBlogs[i]
+      temp.Time = time.Now().Format("15:04:05")
+      temp.Timestamp = time.Now().Format("20060102150405")
+    	ArrayOfBlogsToReturn[i] = temp
+		}
+		return ArrayOfBlogsToReturn
+}
+
+func renderPage(w http.ResponseWriter, r *http.Request){
```
sebastianbenz: renderPage is a too generic name for this method

server.go
```diff
@@ -67,6 +81,59 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	  cookie, err := r.Cookie("itemsWhenLanding")
+		t, _ := template.ParseFiles("dist/components/amp-live-list/index.html")
+		expire := time.Now().AddDate(0, 0, 1)
+    if err != nil {
+    	cookie = &http.Cookie{
+      	Name:  "itemsWhenLanding",
```
sebastianbenz: sug: s/itemsWhenLanding/blogCount/

server.go
```diff
@@ -67,6 +81,59 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	ArrayOfBlogs[7] = BlogItem{Text: "A phone", Image: "/img/silver_nexus.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[8] = BlogItem{Text: "Some Sushi", Image: "/img/image3.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[9] = BlogItem{Text: "Hi Gopher!", Image: "/img/gopher.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+}
+
+//returns a map of blogs which contains a number of items specified by the parameter
+func returnMapOfBlogOfSize(size int) map[int]BlogItem {
+		ArrayOfBlogsToReturn := make(map[int]BlogItem)
+		for i := 0; i < size; i++ {
+			temp := ArrayOfBlogs[i]
```
sebastianbenz: Avoid duplication: extract `Time` and `Timestamp` formatting into its own method.

server.go
```diff
@@ -67,6 +81,59 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	ArrayOfBlogs[4] = BlogItem{Text: "A village landscape", Image: "/img/landscape3.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[5] = BlogItem{Text: "A canyon", Image: "/img/landscape4.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[6] = BlogItem{Text: "A dog", Image: "/img/Shetland_Sheepdog.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[7] = BlogItem{Text: "A phone", Image: "/img/silver_nexus.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[8] = BlogItem{Text: "Some Sushi", Image: "/img/image3.jpg", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+	ArrayOfBlogs[9] = BlogItem{Text: "Hi Gopher!", Image: "/img/gopher.png", Timestamp: time.Now().Format("20060102150405"), Time: time.Now().Format("15:04:05")}
+}
+
+//returns a map of blogs which contains a number of items specified by the parameter
+func returnMapOfBlogOfSize(size int) map[int]BlogItem {
```
sebastianbenz: you don't need this method if you're storing the blogs in a slice:

```
	s := make([]string, 0)
	s = append(s, "blog 1", "blog 2", "blog 3")

	fmt.Println("all:", s)
	fmt.Println("first:", s[0])
	fmt.Println("first two:", s[:2])
```

Here is a full example: https://play.golang.org/p/Z8chQtqzmU

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+    "preview": true
+  }--->
+<!--
+  #### Introduction
+
+  The [amp-live-list](https://github.com/ampproject/amphtml/blob/f2aed679a4c0611ea077985cbe0d1926c7d0a6f5/extensions/amp-live-list/amp-live-list.md) component add support for publishing content that updates live in the AMP page as new content is available in the source document.
+
+  This example shows an implementation with a server rendering a series of 10 elements, `amp-live-list` will poll the page every 15 seconds. Every time `amp-live-list` polls the page, the server adds a new item to the page. Refer to [server.go](https://github.com/ampproject/amp-by-example/blob/master/server.go) for implementation details.
+
+  When the user lands to this page the first time, the server returns one item and creates a cookie containing the number of times the user has accessed the page (it will be 1 on creation). For every other access, this number is incremented by one and a new item is added to the page server side.
```
sebastianbenz: Remove this section - these implementation details are irrelevant for the reader.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <!-- #### Setup -->
+  <!--
+  Import the `amp-live-list` component in the header.
+
+  Use `amp-live-list-item` to style the children of the `amp-live-list`. Refer to the [documentation](https://github.com/ampproject/amphtml/blob/f2981e72843c8e71a9cd2535282e4c4947bb6882/extensions/amp-live-list/amp-live-list.md) for all the details about styling.
```
sebastianbenz: s/for all the details about styling/for details/

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+      display: flex;
+      flex-direction: column;
+    }
+
+    .amp-live-list-item {
+      background-color: #f8f8f8;
+      border-bottom: 1px solid rgba(0, 0, 0, .12);
+    }
+
+    .card-list {
```
sebastianbenz: s/card-list/card/

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+      background-color: #fff;
+      box-shadow: rgba(0, 0, 0, .3) 0 0 5px;
+    }
+
+    .content {
+      font-size: 1.5rem;
+      margin-left: 15px;
+    }
+
+    .button {
```
sebastianbenz: styling a `button` element with a class `.button` is a bit redundant. We should need to style buttons for each example but instead create a generic button style.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+    data-poll-interval="15000"
+    data-max-items-per-page="5"
+    id="amp-live-list-insert-blog">
+    <button id="btn" update on="tap:amp-live-list-insert-blog.update" class="button">You have updates</button>
+    <div items>
+      {{range .}}
+          <div id="{{.Text}}" data-sort-time="{{.Timestamp}}">
+            <div class="card-list">
+              <div>
+                <amp-img src="{{.Image}}"
```
sebastianbenz: material design: make the image fill the card.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+    layout="container"
+    data-poll-interval="15000"
+    data-max-items-per-page="5"
+    id="amp-live-list-insert-blog">
+    <button id="btn" update on="tap:amp-live-list-insert-blog.update" class="button">You have updates</button>
+    <div items>
+      {{range .}}
+          <div id="{{.Text}}" data-sort-time="{{.Timestamp}}">
+            <div class="card-list">
+              <div>
```
sebastianbenz: why is this div required?

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+</head>
+<body>
+
+  <div>
+    <h3>Live updates</h3>
+    <h4>Refresh takes 15 seconds.</h4>
+  </div>
+  <!-- #### amp-live-list with insert -->
+  <!-- 
+    Use `data-poll-interval` attribute to specify how frequently the `amp-live-list` component will poll new data. Values are valid above the minimum of 15 seconds.
```
sebastianbenz: s/Values are valid above the minimum of 15 seconds./The minimum interval is 15 seconds./

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+<body>
+
+  <div>
+    <h3>Live updates</h3>
+    <h4>Refresh takes 15 seconds.</h4>
+  </div>
+  <!-- #### amp-live-list with insert -->
+  <!-- 
+    Use `data-poll-interval` attribute to specify how frequently the `amp-live-list` component will poll new data. Values are valid above the minimum of 15 seconds. 
+    Use `data-max-items-per-page` for specifying the maximum number displayed on the page. Additional elements are assumed to be on the next "page".
```
sebastianbenz: s/for specifying/to specify/

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+    For an extensive explanation of all the attributes, refer to [amp-live-list](https://github.com/ampproject/amphtml/blob/f2aed679a4c0611ea077985cbe0d1926c7d0a6f5/extensions/amp-live-list/amp-live-list.md).
+    
+    Use `on="tap:my-live-list.update"` to load the latest changes from the server. 
+    A button will appear when new items are discovered by the client on polling the server document. Note that the button has to be defined as a direct child of `amp-live-list` .
```
sebastianbenz: s/when new items are discovered by the client on polling the server document/when the AMP runtime discovers new document changes./

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+
+    A `amp-live-list` must contain a list of items, defined like `<div items>`. Each item has a specific id and the `amp-live-list` will look for items with new id.
```
sebastianbenz: s/list of items, defined like `<div items>`/list of items below `<div items>`/

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+    Use `data-max-items-per-page` for specifying the maximum number displayed on the page. Additional elements are assumed to be on the next "page".
+    For an extensive explanation of all the attributes, refer to [amp-live-list](https://github.com/ampproject/amphtml/blob/f2aed679a4c0611ea077985cbe0d1926c7d0a6f5/extensions/amp-live-list/amp-live-list.md).
+    
+    Use `on="tap:my-live-list.update"` to load the latest changes from the server. 
+    A button will appear when new items are discovered by the client on polling the server document. Note that the button has to be defined as a direct child of `amp-live-list` .
+
+    A `amp-live-list` must contain a list of items, defined like `<div items>`. Each item has a specific id and the `amp-live-list` will look for items with new id. 
+
+    By clicking on the button "You have updates", new content is displayed on the demo column (the one on the right if the demo is accessed by desktop). The code column (the one in the center if the demo is accessed by desktop) does not update itself, but if you want to see all the items on the page, reload the page.
```
sebastianbenz: put this within `<p class="card">...</p>` to make it explicit that this is something ABE specific.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,147 @@
...
+                </amp-img>
+              </div>
+              <div class="content">{{.Text}}</div>
+            </div>
+          </div>
+        {{end}}
+    </div>
+  </amp-live-list>
+
+  <!-- #### amp-live-list with update -->
```
sebastianbenz: Do we need this sample?

sebastianbenz:
Nice work! I like the sample! Some comments inline also: can you put the server side code into a separate file: amp-live-list.go. This way we clearly separate server-side sample code. (we should have done the same for the amp-cache sample).

src/20_Components/amp-live-list.html
sebastianbenz: Just discovered that we've been doing the same. 😳

server.go
kul3r4: The main reason of having this method was that I wanted to update the timestamp. Just doing a slice of the existing array, will return items with the time of their creation. The reason why I need new timestamps, is that, for the second example of amp-live-list I am demonstrating the update capability. This capabiity depends on "data-update-time" parameter: if the server page generates recent values for this parameter, then the component will change.

src/20_Components/amp-live-list.html
kul3r4: This explains the sentence "If you want to see the demo again, delete the cookie or access the page using an Incognito tab." which makes possible for the user to see the demo again and again. I could change it in this way:

"This example uses a cookie to emulate a live blog starting with one blog post. After 10 times the amp-live-list loads new blog posts, the server stops generating new blog posts. If you want to see the demo again, delete the cookie or access the page using an Incognito tab."

src/20_Components/amp-live-list.html
kul3r4: what about using class=button-primary?

sebastianbenz: I think that's ok

src/20_Components/amp-live-list.html
sebastianbenz: Just say: "The sample live blog stops after ten articles. If you want to start over, delete cookies from ampbyexample.com or open this page again in an incognito/private tab."

src/20_Components/amp-live-list.html
kul3r4: what does next "page" mean? Will remove this because it doesn't mean anything, there is not a mechanism for "next page" at the moment. If the value of this attribute is x and the number of items retrieved from the server is y, where y>x, only x items will be displayed.

sebastianbenz: ok

src/20_Components/amp-live-list.html
kul3r4: Looking at [feed reader](https://material.google.com/components/cards.html#cards-usage): image and text share the card. If the image would fill the card, where should I put the text?

kul3r4:
Updated with the comments, changed the second example with a match points instead of a timestamp. @sebastianbenz PTAL

kul3r4:
Sorry, I didn't see the last comment about separating in a separate file, will do ASAP.

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+		createBlogEntry("Some Sushi", "/img/image3.jpg"),
+		createBlogEntry("Hi Gopher!", "/img/gopher.png"))
+}
+
+func createBlogEntry(text string, imagePath string) BlogItem {
+	return BlogItem{Text: text, Image: imagePath, Timestamp: currentTimestamp()}
+}
+
+//create a clone of the blog item, instead of pointing
+func cloneBlogItem(blogItem BlogItem) BlogItem{
```
sebastianbenz: make this a method of BlogItem:
```
func (blogItem BlogItem) clone() BlogItem {
  return createBlogEntry(blogItem.Text, blogItem.Image)
}

...

blogItem.clone();
```

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	return createBlogEntry(blogItem.Text, blogItem.Image)
+}
+
+//returns a map of blogs which contains a number of items specified by the parameter
+func returnMapOfBlogOfSize(size int) []BlogItem {
+		ArrayOfBlogsToReturn := make([]BlogItem, 0)
```
sebastianbenz: s/ArrayOfBlogsToReturn/result/ <- don't encode type information in names

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	return BlogItem{Text: text, Image: imagePath, Timestamp: currentTimestamp()}
+}
+
+//create a clone of the blog item, instead of pointing
+func cloneBlogItem(blogItem BlogItem) BlogItem{
+	return createBlogEntry(blogItem.Text, blogItem.Image)
+}
+
+//returns a map of blogs which contains a number of items specified by the parameter
+func returnMapOfBlogOfSize(size int) []BlogItem {
```
sebastianbenz: the name of the method is misleading, suggestion. Sug:

`func getBlogEntries(count int)`

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+func updateScore(score string) Score{
+	return Score{Timestamp: currentTimestamp(), Score: score}
+}
+
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405")
+}
+
+func renderLiveBlog(w http.ResponseWriter, r *http.Request) {
+	  cookie, err := r.Cookie("abe_amp-live-list_blogCount")
```
sebastianbenz: s/abe_amp-live-list_blogCount/ABE_AMP_LIVE_LIST_BLOG_COUNT/

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+	return Score{Timestamp: currentTimestamp(), Score: score}
+}
+
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405")
+}
+
+func renderLiveBlog(w http.ResponseWriter, r *http.Request) {
+	  cookie, err := r.Cookie("abe_amp-live-list_blogCount")	
+		expire := time.Now().AddDate(0, 0, 1)
```
sebastianbenz: when does it expire? You could encode this info in the variable name, e.g. expireInOneDay.

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+      	Name:  "abe_amp-live-list_blogCount",
+        Expires: expire,
+        Value: "1",
+    	}
+    } else {
+			cookie.Value = incBlogCounter(cookie)
+    }
+
+    http.SetCookie(w, cookie)
+		sizeOfBlogMap, err := strconv.Atoi(cookie.Value)
```
sebastianbenz: s/sizeOfBlogMap/blogCount/

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+		//pass to the template a map containing a number of items which is equal to the cookie value
+		mapToBeShown := returnMapOfBlogOfSize(sizeOfBlogMap)
```
sebastianbenz: mapToBeShown -> it's not a map. Can you find a better name?

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+
+func updateScore(score string) Score{
+	return Score{Timestamp: currentTimestamp(), Score: score}
+}
+
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405")
+}
+
+func renderLiveBlog(w http.ResponseWriter, r *http.Request) {
```
sebastianbenz: This method does too many things:

* read a cookie
* write a cookie
* render a template
* parses the blog count

Try to split it up into different methods.

server.go
sebastianbenz: The name is also a bit misleading as it is the blog count and the current score. I see two possibilities:

* rename to ABE_AMP_LIVE_LIST_STATUS
* store score in a separate value

server.go
```diff
@@ -50,10 +52,31 @@ var REDIRECTS [18][2]string = [18][2]string{
 	{"/Hello_World.html", "/components/Hello_World.html"},
 }
 
+type BlogItem struct {
+	Text string
+	Image string
+	Timestamp string
+}
+
+type Score struct {
```
sebastianbenz: Doesn't a score have at least to numbers?

server.go
```diff
@@ -67,6 +90,83 @@ func NoDirListing(h http.Handler) http.HandlerFunc {
...
+		if size > len(blogs) {
+			size = len(blogs)
+		}
+		for i := 0; i < size; i++ {
+    	ArrayOfBlogsToReturn = append(ArrayOfBlogsToReturn, cloneBlogItem(blogs[i]))
+		}
+		return ArrayOfBlogsToReturn
+}
+
+func updateScore(score string) Score{
```
sebastianbenz: suggestion: make the score a bit more sophisticated, e.g. "1:0", "2:0", ...

Don't update the score put create a new one with the current list status, the second team could always have the score `0`:

```
createScore(status, 0);
```

sebastianbenz: Just seen that you're handling this on template side. This is also OK, but I'd prefer to make this a bit more 'realistic'.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
+<!---{
+    "experiment": true,
+    "component": "amp-live-list"
+  }--->
+<!--
+  #### Introduction
+
+  The [amp-live-list](https://github.com/ampproject/amphtml/blob/f2aed679a4c0611ea077985cbe0d1926c7d0a6f5/extensions/amp-live-list/amp-live-list.md) component add support for publishing content that updates live in the AMP page as new content is available in the source document.
```
sebastianbenz: s/component add/component adds/

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+
+  This example shows an implementation with a server rendering a series of 10 elements, `amp-live-list` will poll the page every 15 seconds. Every time `amp-live-list` polls the page, the server adds a new item to the page. Refer to [server.go](https://github.com/ampproject/amp-by-example/blob/master/server.go) for implementation details.
```
sebastianbenz: This section applies only to the first sample. Remove it.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+
+  The sample live blog stops after ten articles. If you want to start over, delete cookies from ampbyexample.com or open this page again in an incognito/private tab.
```
sebastianbenz: s/The sample live blog stops after ten articles. /This sample stops after ten updates./

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <!-- #### Setup -->
+  <!--
+  Import the `amp-live-list` component in the header.
+
+  Use `amp-live-list-item` to style the children of the `amp-live-list`. Refer to the [documentation](https://github.com/ampproject/amphtml/blob/f2981e72843c8e71a9cd2535282e4c4947bb6882/extensions/amp-live-list/amp-live-list.md) for details.
```
sebastianbenz: What does this sentence apply to?

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+  </style>
+</head>
+<body>
+
+  <div>
+    <h3>Live updates</h3>
+    <h4>Refresh takes 15 seconds.</h4>
+  </div>
```
sebastianbenz: remove the whole div - it doesn't add value to the sample.

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+  <!-- #### amp-live-list with insert -->
```
sebastianbenz: suggestion: maybe a better headline would be "Implementing a Live Blog"

src/20_Components/amp-live-list.html
sebastianbenz: suggestion: maybe a better headline would be "Implementing a Live Score"

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+                <p>{{.Text}}</p>
+            </div>
+          </div>
+        {{end}}
+    </div>
+  </amp-live-list>
+
+  <!-- #### amp-live-list with update -->
+  <!-- 
+    When using `amp-live-list` with only updates on elements, the update button will not be shown.
```
sebastianbenz: sug: s/When using `amp-live-list` with only updates on elements, the update button will not be shown.
Use `data-update-time` to specify when the update has been performed./If you want to update existing list items, you need  to specify via the  `data-update-time` attribute when the update has been performed./

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+    Use `data-update-time` to specify when the update has been performed.
+
+    The following example will display a soccer match score whose content is changed every time the `amp-live-list` polls the page on the server.
+  -->
+    <amp-live-list
+    layout="container"
+    data-poll-interval="15000"
+    data-max-items-per-page="5"
+    id="amp-live-list-update">
+    <button id="btn" update on="tap:amp-live-list-update.update">You have updates</button>
```
sebastianbenz: do we need `id=btn`?

src/20_Components/amp-live-list.html
```diff
@@ -0,0 +1,146 @@
...
+    <div items>
+        <div id="list-item-1" data-sort-time="1234567" data-update-time="{{.FootballScore.Timestamp}}">
```
sebastianbenz: Does data-sort-time="1234567" need to be present?

sebastianbenz:
I've added some comments. You're getting there!

Tip: are you using [gofmt](https://blog.golang.org/go-fmt-your-code)? I have it as a save action configured in my editor.

server.go
kul3r4: Do you mean adding more scores? E.g.: at the moment there is only one match, going on, would you like to have more than one when the page loads for the first time and then updating the values at every iteration?

sebastianbenz: Simply score: 1:2 instead of 1

src/20_Components/amp-live-list.html
kul3r4: Right, I was not using that class anymore but I could in order to center the cards.
Look at [this](https://www.dropbox.com/s/6bikotoh0xa4la2/am-live-list-center.png?dl=0)
I realized that by doing:
`.amp-live-list-item {
      justify-content: center;  
      display: flex;     
    }
    .blog{
      padding: 0;
      min-width: 70%;
      background: #fff;
    }
    .blog p {
      padding: 8px;
      margin: 0;
    }`
What do you think?

src/20_Components/amp-live-list.html
kul3r4: From [doc](https://github.com/ampproject/amphtml/blob/master/extensions/amp-live-list/amp-live-list.md)  data-sort-time is required. In this corner case it does not make a lot of sense, will follow up on that on a separate thread.

kul3r4:
@sebastianbenz PTAL

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,131 @@
...
+	http.SetCookie(w, cookie)
+	renderLiveBlogTemplate(w, createPage(cookie))
+}
+
+func renderLiveBlogTemplate(w http.ResponseWriter, page Page) {
+	t, _ := template.ParseFiles("dist/components/amp-live-list/index.html")
+	t.Execute(w, page)
+}
+
+func getAMPLiveListCookie(r *http.Request) *http.Cookie {
```
sebastianbenz: When a method name starts with "get" it shouldn't change the state.

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,131 @@
...
+	blogItems := getBlogEntries(ampLiveListStatus)
+	score := createScore(ampLiveListStatus, 0)
+	return Page{BlogItems: blogItems, FootballScore: score}
+}
+
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405")
+}
+
+func RenderLiveBlog(w http.ResponseWriter, r *http.Request) {
```
sebastianbenz: You use the cookie to pass values between the different methods. However, all other methods don't need to know about cookies. They just care about the status. This even leads to the effect that you have to parse the cookie value twice into an int.

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func renderLiveBlogTemplate(w http.ResponseWriter, page Page) {
+	t, _ := template.ParseFiles("dist/components/amp-live-list/index.html")
+	t.Execute(w, page)
+}
+
+func getAMPLiveListCookie(r *http.Request) *http.Cookie {
+	cookie, err := r.Cookie(AMP_LIVE_LIST_COOKIE_NAME)
+	expireInOneDay := time.Now().AddDate(0, 0, 1)
```
sebastianbenz: expireInOneDay is only needed in one if branch -> move it there

src/20_Components/amp-live-list.html
sebastianbenz: Good idea. Move the sentence then above the CSS block.

sebastianbenz:
Almost there :-). Only smaller nits left.

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
One comment is still open.

kul3r4:
@sebastianbenz Applied changes to the RenderLiveBlog method as discussed. PTAL.

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405") 
+}
+
+func writeStatus(w http.ResponseWriter, newValue int) {
+	cookie := createAMPLiveListCookie(strconv.Itoa(newValue))
+	http.SetCookie(w, cookie)
+}
+
+func updateAMPLiveListCookie(cookie *http.Cookie, ampLiveListNewStatus int) {
```
sebastianbenz: where is this used?

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+  newStatus := readStatus(r) + 1
+  page := createPage(newStatus)
+  writeStatus(w, newStatus);
+  renderAmpLiveListSample(w, page)
+  
+}
+
+func readStatus(r *http.Request) int {
+	cookie, err := r.Cookie(AMP_LIVE_LIST_COOKIE_NAME)		
+	if (err == nil) {
```
sebastianbenz: it's common in go to handle the error case first. Also, Atoi return 0 if the string can't be parsed:

```
cookie, err := r.Cookie(AMP_LIVE_LIST_COOKIE_NAME)
if (err != nil) {
  return 0
}
return strconv.Atoi(cookie.Value)
```

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+	score := createScore(newStatus, 0)
+	return Page{BlogItems: blogItems, FootballScore: score}
+}
+
+func renderAmpLiveListSample(w http.ResponseWriter, page Page) {
+	t, _ := template.ParseFiles("dist/components/amp-live-list/index.html")
+	t.Execute(w, page)
+}
+
+func (blogItem BlogItem) clone() BlogItem {
```
sebastianbenz: move this below BlogItem

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+func createScore(scoreTeam1 int, scoreTeam2 int) Score {
+	return Score{Timestamp: currentTimestamp(), ScoreTeam1: scoreTeam1, ScoreTeam2: scoreTeam2}
+}
+
+func currentTimestamp() string {
+	return time.Now().Format("20060102150405") 
+}
+
+func writeStatus(w http.ResponseWriter, newValue int) {
+	cookie := createAMPLiveListCookie(strconv.Itoa(newValue))
```
sebastianbenz: We could now even inline createAMPLiveListCookie

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+func createAMPLiveListCookie(value string) *http.Cookie{
+	expireInOneDay := time.Now().AddDate(0, 0, 1)
+	return &http.Cookie{
+		Name:    AMP_LIVE_LIST_COOKIE_NAME,
+		Expires: expireInOneDay,
+		Value:   value,
+	}
+}
+
+func incBlogCounter(cookieValue int) int {
```
sebastianbenz: where is this used?

amplivelist/amp-live-list.go
```diff
@@ -0,0 +1,144 @@
...
+	ScoreTeam2 int
+}
+
+type Page struct {
+	BlogItems     []BlogItem
+	FootballScore Score
+}
+
+var blogs []BlogItem
+var startLoop int
```
sebastianbenz: where is this used?

sebastianbenz:
Nice! Only a few nits left.

kul3r4:
@sebastianbenz PTAL, agree on all the previous comments.

sebastianbenz:
LGTM - well done!

sebastianbenz:
Can you please push to staging after merging to check that everything works as expected.

PR #180, opened 2016-06-07
Change alt text to "close sidebar"

ithinkihaveacat:
Improve accessibility (?). Not sure if AMP is already doing something here.


sebastianbenz:
makes sense - LGTM

PR #186, opened 2016-06-10
Added ad to amp-fx-flying-carpet

kul3r4:
Added an ad to the amp-fx-flying-carpet example, while keeping the example with the image.
The best experience in on mobile where the ad aligns itself on the center, as specified in the comments. On desktop (when 3 column are displayed), the ad float on the right. The reason is that the parent class, -amp-fx-flying-carpet-container, inherits:

    position: fixed!important;
    top: 0!important;
    left: 0!important;
    width: 100%!important;
so the ad can be centered easily when having only one ABE column.

@sebastianbenz PTAL


src/20_Components/amp-fx-flying-carpet.html
```diff
@@ -33,13 +45,32 @@
...
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis venenatis, metus a tristique aliquet, enim nibh efficitur sem, ut iaculis urna justo eu diam. Nullam cursus sapien et sodales posuere. Vestibulum ante ipsum primis in faucibus orci luctus et ultrices posuere cubilia Curae; Suspendisse potenti. Donec vitae ornare risus. Maecenas eleifend ante vel dui laoreet, et porttitor libero rutrum. Nam arcu mi, ullamcorper at risus et, pulvinar ultrices erat. In pellentesque sem vel purus auctor, ut venenatis tellus tristique. Phasellus molestie diam orci, nec gravida turpis bibendum ut. Sed sagittis aliquet lorem sed dictum.
     </p>
 
   <!--
-    `amp-fx-flying-carpet` can be used to display images.
-    Use `height` parameter to specify the height of the flying carpets "window".
+    `amp-fx-flying-carpet` can be also used to display images.
 
     Use vw/vh to size elements inside the `amp-fx-flying-carpet` relative to the size of the viewport. This will allow the best scrolling effect having the element using the entire viewport.
```
sebastianbenz: suggestions:  

* mark vw and vh as code `vw`/`vh` 
* s/This will allow the best scrolling effect having the element using the entire viewport./Having the element cover the entire viewport provides a nice scrolling effect./

sebastianbenz:
The sample works great! One small suggestion.

kul3r4:
Thanks for the suggestions, comments applied. PTAL.

sebastianbenz:
LGTM

PR #189, opened 2016-06-11
Remove AMP URL API Wrapper

sebastianbenz:
* no longer needed as API now supports CORS
* rename view folder -> iframe
* increase result view size to match default result
* API key no longer needs to be encrypted as its access controlled via request host


sebastianbenz:
@kul3r4 @juliantoledo PAL

api/server.go
```diff
@@ -15,36 +15,13 @@
...
 import (
-	"encoding/json"
-	"google.golang.org/appengine"
 	"net/http"
 )
 
-// init the handlers
 func init() {
-	http.HandleFunc("/url", apiHandler)
+	http.HandleFunc("/", serveNotFound)
```
kul3r4: Small typo: serveNotFound -> serverNotFound

sebastianbenz: serveNotFound is the intended name => serve the notFoundPage.

kul3r4:
LGTM.

PR #195, opened 2016-06-13
Improvements to amp-fx-flying-carpet

kul3r4:
- add separator between amp-fx-flying-carpet and the text

- add orange background 

- use another ad


kul3r4:
@sebastianbenz PTAL.

sebastianbenz:
LGTM

PR #198, opened 2016-06-14
activate experiments via button

### LINKED ISSUE #196: Turn experiment activation into button
I suppose this might be problematic in terms of AMP validity, but it would be super nice if the experiment instructions on https://ampbyexample.com/components/amp-fx-flying-carpet/ would just run the code instead of asking to paste it into the console. Maybe the button could use an iframe :)

Something like `<button onclick="AMP.toggleExperiment('amp-fx-flying-carpet')">Activate experiment</button>`

sebastianbenz:
* Enable experiment via a button
* Hides the experiment card if already enabled
* Show experimental flask icon after example name

Fixes #196


sebastianbenz:
Try 
[here](https://dl.dropboxusercontent.com/u/7234606/amp-flying-carpet-experiment.html). 

Reading the enabled experiments from cookies as I couldn't find a callback for when the AMP runtime has loaded.

@cramforce PTAL 

//cc @kul3r4 @juliantoledo

sebastianbenz:
The nice side effect is that you'll finally be able to run experimental samples on a mobile device.

templates/experiment.html
```diff
@@ -1,12 +1,50 @@
-<div class="card info-card">
+<!-- BEGIN INVALID AMP - don't try at home -->
+<div id="experiment-warning" class="card info-card">
   <h4>Experimental Mode</h4>
-  <p>The <code>{{component}}</code> component is currently in experimental mode and needs to be manually enabled:</p>
-
-  <ol>
-    <li>check the box under <a href="https://cdn.ampproject.org/experiments.html">CDN experiments</a>.
```
cramforce: You still want to do this, right?

You could detect whether the user opted in by looking at `window.AMP_CONFIG`

templates/experiment.html
```diff
@@ -1,12 +1,50 @@
...
+function updateExperimentWarning() {
+  if (!isExperimentOn('{{component}}')) {
+    return;
+  }
+  const experimentWarning = document.getElementById('experiment-warning');
+  experimentWarning.style.display = 'none';
+}
+
+/* 
+ * Not using AMP.isExperimentOn(...) as we don't know when the AMP runtime is
```
cramforce: You could use our super undocumented API to do this. Might be better than using our undocumented cookie API :)

```js
window.AMP = window.AMP || []).push(function(AMP) {
  // Code that can use AMP goes here.
});
```

templates/experiment.html
sebastianbenz: Does AMP.toggleExperiment('AMP_CANARY') work or does it need to be called from cdn.ampproject.org?

templates/experiment.html
sebastianbenz: aah - much better. done.

sebastianbenz:
@cramforce PTAL

templates/experiment.html
cramforce: It doesn't work, since that cookie is being used on the server side on cdn.ampproject.org to select the delivered files.

sebastianbenz: too bad - added a button for cdn.ampproject.org/experiments.html

sebastianbenz:
@cramforce PTAL - sample [here](https://dl.dropboxusercontent.com/u/7234606/amp-flying-carpet-experiment.html)

cramforce:
LGTM

sebastianbenz:
Thanks for the review!

PR #199, opened 2016-06-14
Fix live-reload

### LINKED ISSUE #191: gulp / live-reload not effective in the long run
I've been iterating on a new sample and now regardless of the changes I make to the source HTML, live-reload triggers the page reload, but the change isn't reflected. I have to manually force refresh in order to see the changes.

Node 4.4.5 on Ubuntu 16.04. Followed [the steps in the README](https://github.com/ampproject/amp-by-example#creating-a-new-sample).

sebastianbenz:
Delay live reload for 500ms after a file is changed to
make sure that all changes have been written to the file.

Fixes #191

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -83,9 +83,11 @@ gulp.task('serve', 'starts a local webserver (--port specifies bound port)',
     const server = gls.static(paths.dist.dir, port);
     server.start();
     gulp.watch([paths.dist.html], function(file) {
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Not nice, but works. Better suggestions are welcome.

PTAL @dandv @kul3r4 @juliantoledo

juliantoledo:
LGTM, we can improve it if we find another way to do it.

PR #205, opened 2016-06-16
hide column in experimental mode

sebastianbenz:
Otherwise columns will be shown even if the sample is in column mode.

### DIFF
--- a/templates/css/styles.css
+++ b/templates/css/styles.css
@@ -24,7 +24,7 @@ article {
   max-width: 2400px;
   margin: auto;
 }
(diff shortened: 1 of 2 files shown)


sebastianbenz:
@kul3r4 PTAL

kul3r4:
Which columns will be hidden when in experimental mode with this change? I run it and I couldn't spot the difference with the current one.

sebastianbenz:
I don't think this affects any of our examples yet.

kul3r4:
LGTM

PR #212, opened 2016-06-22
Add Serviceworker and use for Caching

sebastianbenz:
* install serviceworker for all samples
* precache font and landing page
* update amp-install-serviceworker sample to reflect these changes
* small manifest fixes
* remove sw-precache and use sw-toolbox instead
* use bower to install JS client dependencies


sebastianbenz:
Theoretically this would give us a score of 100 in [Lighthouse](https://github.com/GoogleChrome/lighthouse). However, I've still disabled the manifest as the AMP validator doesn't support it. Not sure whether it is worth breaking the AMP compatibility. 

PTAL @juliantoledo @jmadler 

//CC @ymotongpoo @adewale @kul3r4

sebastianbenz:
Deployed [here](https://amp-by-example-staging.appspot.com/)

juliantoledo:
It's super fast! What exactly does not validate? each example or just amp-install-serviceworker?

sebastianbenz:
It currently validates - but to become a proper PWA it needs a manifest.json file. The manifest.json link is currently not supported by the AMP validator that's why I've commented it out in templates/favicons.html.

juliantoledo:
LGTM then, we can uncomment the link once fixed or decide to break compatibility later if it isn't in a while.

static/sw.js
```diff
@@ -0,0 +1,13 @@
...
+
+toolbox.options.debug = true;
+toolbox.router.default = toolbox.fastest;
+
+const filesToCache = [
+    '/',
+    'https://fonts.googleapis.com/css?family=Roboto:regular,bold,italic,thin,light,bolditalic,black,medium&lang=en'
+];
+
+self.addEventListener('install', function(e) {
```
wibblymat: toolbox.precache doesn't actually go in an install handler, just in the body of the script. Toolbox creates it's own install handler, and if they get executed in the wrong order you might find that precaching isn't happening.

sebastianbenz: ok - done. Thanks!

static/sw.js
```diff
@@ -0,0 +1,11 @@
+importScripts('/bower_components/sw-toolbox/sw-toolbox.js');
+
+toolbox.options.debug = true;
+toolbox.router.default = toolbox.fastest;
+
+const filesToCache = [
+    '/',
+    'https://fonts.googleapis.com/css?family=Roboto:regular,bold,italic,thin,light,bolditalic,black,medium&lang=en'
```
ithinkihaveacat: I'm not sure of the intention here but looks like this will only cache the CSS, not the fonts themselves--I don't think this will actually provide fonts offline. (Although it will "work" on modern Android devices, because Roboto is pre-installed.)

Providing fonts offline is quite difficult (especially via Google Fonts) due to font subsetting and other browser-based optimizations that the service worker doesn't have visibility of--are there some other resources that can be cached instead?

sebastianbenz: Good point - I was desperately looking for some kind of external resource to cache. It doesn't make sense to explicitly precache it anyway as the font will be loaded (and cached) together with every page.

Regarding caching other resources, I plan to:

* precache samples + supporting files (img, json resources). 
* precache AMP components (which don't have CORS enabled, so I'm not sure yet if that's a good idea)

sebastianbenz: Removed the font caching. Will add more caching later.

sebastianbenz:
PTAL

static/sw.js
```diff
@@ -0,0 +1,10 @@
+importScripts('/bower_components/sw-toolbox/sw-toolbox.js');
+
+toolbox.options.debug = true;
+toolbox.router.default = toolbox.fastest;
```
jmadler: Based on the staging site, there are a number of requests repeated on each pageload that could be added to the cache using either cacheFirst or cacheOnly:

* favicon.ico
* GitHub-Mark-Light-64px.png
* favicon-*.png

It may also make sense to have the cache header be something other than maxage=1 for these

static/sw.js
```diff
@@ -0,0 +1,10 @@
...
+
+const filesToCache = [
+    '/'
```
jmadler: Is there any reason not to precache all the published content?

static/manifest.json
```diff
@@ -4,9 +4,8 @@
   "description": "Learn AMP by Examples",
   "start_url": "https://ampbyexample.com",
   "display": "standalone",
```
jmadler: One thought:  Since you're using standalone display, Chrome is unable to display a back button.  Some devices don't include back buttons, and ABE doesn't have one, which means that on some mobile devices, it's possible that a user can hit a part of the stack that's not reversible.

As it stands, I'm not sure if the current application flow has any bit that's not accessible by the menu, or that doesn't have access to the menu.  I'm not sure what happens when navigating to GitHub when in standalone mode.  I can't easily test that from staging.

static/sw.js
sebastianbenz: Good point about the max-age, we haven't really looked in the cache header. I've created #226 to fix this.

static/sw.js
sebastianbenz: Precaching all samples only makes sense if they work offline as well. This isn't the case for all samples. Some require amp components which might not be in the browser cache yet. Other require external components which can't be cached. We first need to implement a proper offline strategy for ABE first. (see #225 for more about this).

static/manifest.json
sebastianbenz: As long as the device has a back button everything works fine (even navigating to Github). I'd leave it for know for demo purposes, but we definitely have to keep this in mind. For example, we would need to add a sharing button in standalone mode as there is no other way to access the URL. 

BTW which devices don't have a back button (besides Amazon Fire with manually installed Chrome)?

sebastianbenz:
Update the PR to perform proper precaching:

* Show an offline page if page is not available offline
* Generate placeholder images when offline
* Precache all icons and logos

Also for the offline page:
* Add support for standalone / non-example pages. 

@jmadler @ithinkihaveacat @juliantoledo PTAL

src/20_Components/amp-install-serviceworker.html
```diff
@@ -34,18 +34,18 @@
   -->
   <amp-install-serviceworker
       src="/sw.js"
-      data-iframe-src="https://ampbyexample.com/sw.html"
+      data-iframe-src="/sw.html"
```
ithinkihaveacat: I think the domain is necessary here (so that it works on the cache).

sebastianbenz: Done.

src/20_Components/amp-install-serviceworker.html
```diff
@@ -34,18 +34,18 @@
...
       layout="nodisplay">
   </amp-install-serviceworker>
   <!--
-  <p class="card info">The service worker in this sample ([sw.js](https://ampbyexample.com/sw.js)) is generated using [sw-precache](https://github.com/GoogleChrome/sw-precache). With sw-precache you can generate service worker code that will precache specific resources.</p> 
+  <p class="card info">The [service worker](/sw.js) in this sample uses [sw-toolbox](https://googlechrome.github.io/sw-toolbox/). </p>
```
jmadler: What does sw-toolbox do?

src/You%27re_Offline.html
```diff
@@ -0,0 +1,44 @@
...
+  #dinosaur {
+    margin-top: 0;
+  }
+}
+  </style>
+
+</head>
+<body>
+<!--
+#### Serviceworker to the rescue
```
jmadler: I think "Serviceworker" is better styled as "ServiceWorker" or "Service Worker" or "service worker"

src/You%27re_Offline.html
```diff
@@ -0,0 +1,44 @@
...
+
+The fact that you're seeing this page is because a serviceworker is running in the background. Serviceworkers are great for building offline experiences.
```
jmadler: What is that SW doing?

static/manifest.json
```diff
@@ -2,11 +2,10 @@
   "name": "AMP by Example",
   "short_name": "AMP Examples",
   "description": "Learn AMP by Examples",
-  "start_url": "https://ampbyexample.com",
+  "start_url": "http://192.168.0.13:8000",
```
jmadler: Mistaken commit? :)

sebastianbenz: doh! fixed.

src/You%27re_Offline.html
sebastianbenz: added an explanation.

src/20_Components/amp-install-serviceworker.html
sebastianbenz: added a note.

src/You%27re_Offline.html
sebastianbenz: done

sebastianbenz:
Addressed the comments. 

@jmadler @ithinkihaveacat @juliantoledo PTAL

static/manifest.json
jmadler: iOS, FireOS, and possibly some AOSP devices without CTS. 

I'm comfortable deferring given that Safari doesn't honor link rel=manifest, and we're not enabling standalone display on iOS.  If you'd like to do that, I'd suggest opening a separate ticket to track.

andreban:
LGTM

sebastianbenz:
Thanks for the reviews!

PR #226, opened 2016-06-26
more fine grained max age control

sebastianbenz:
* 3min for samples (no revalidation)
* 1d for images, videos, css, scripts (we currently don't have any of
  the latter two)
* 10s for amp-live-list sample
* 30d for favicons

### DIFF
--- a/amplivelist/amp-live-list.go
+++ b/amplivelist/amp-live-list.go
@@ -15,6 +15,7 @@
 package amplivelist
 
 import (
+	"fmt"
 	"html/template"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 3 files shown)


sebastianbenz:
@jmadler PTAL

//cc @kul3r4

jmadler:
I'm not familiar enough with the infrastructure to review implementation, so I'm only reviewing the caching rules.

3min seems a bit low for samples, but assuming that ETags are used in the front-end, not really relevant.

sebastianbenz:
We are not using etags. I was following [Jake's](https://jakearchibald.com/2016/caching-best-practices/) advice. I think 3min for samples should be fine. We can always adjust it later.

@kul3r4 can you review the server-side stuff?

kul3r4:
server-side: LGTM

Issue #161, opened 2016-05-13
Validate sample files


jmadler:
A number of the documents under src/*/*.html are invalid AMP.  It might make sense to validate these as part of gulp validate, and also to modify them to be valid.

kul3r4:
Could you enumerate which ones are invalid?

sebastianbenz:
We already validate the generated sample files (and this is what counts). AFAIK all samples are valid - only the ones marked as experimental are not validated as they the validator is often not yet up to date. Most of the samples in src are missing the canonical link which is generated automatically. Hence validation doesn't really make sense here.

jmadler:
Only amp-social-share is valid.  The rest seem to be missing 'link rel=canonical'.

kul3r4:
Hi, the generated pages have the link rel=canonical.
For example, I checked [amp-facebook](view-source:https://ampbyexample.com/components/amp-facebook/,) and it contains: <link rel="canonical" href="https://ampbyexample.com/components/amp-facebook/" >.

jmadler:
Yes, but the link rel=canonical isn't in the source pane of the sample

jkingyens:
I am re-opening this because I think there is a deeper issue here. @Gregable, myself and @jmadler have assumed the files in "src" should be valid AMP at one point or another. I think we should make a change that makes it more clear these files are not intended to be valid AMP. It will benefit all future users who are stumped by this. Ideas:

1) change the extension of the files from .html to html.input to re-enforce the idea that these files need to be processed to become html.
2) remove the <html ⚡><head>... from the src files to emphasize the source are just "fragments" that are combined during build to create the final AMP files.

I think 2) is best as it also removes the confusion around where the data comes from to produce the output AMPs. Right now it appears like the <html ⚡><head>... is coming from the source when in fact it is being pulled in from other snippets.

kul3r4:
Fixed by #234

Issue #14, opened 2016-02-23
Add link rel=canonical tags to examples


Gregable:
Currently none of the examples fully validate due to missing `<link rel=canonical>` tags.
https://www.ampproject.org/docs/reference/spec.html#canon

sebastianbenz:
Every page should contain a canonical link to itself, e.g. 

`<link rel="canonical" href="Hello_World.html">`

in https://amp-by-example.appspot.com/Hello_World.html

jkingyens:
I'm seeing this in my "gulp validate" task branch as well. However, the example html files appear fine according to the spec. The odd thing is Hello_World.html is passing for me. It looks significantly different from the other examples. Is there something different about its conformance?

jkingyens:
The cause for me was a typo/bug in my validator task. @Gregable are you trying to validate the html files in "src" directory? These are templates that need to be expanded in order to validate correctly.

Gregable:
Sorry I must have missed your reply.

The pages themselves contain a rel=canonical link, but the example that is rendered in the page does not. Consider for example:
https://ampbyexample.com/components/amp-access/
This document itself contains:
<link rel="canonical" href="https://ampbyexample.com/components/amp-access/" >

So if I view-source:https://ampbyexample.com/components/amp-access/ and control+f for `canonical`, I find it.

However the actual example doesn't contain rel=canonical. So, if I'm looking at the rendered example doc, and not the source,and control+f for `canonical`, I don't find it. https://ampbyexample.com/introduction/hello_world/ has a canonical in the example:
`&lt;<span class="hljs-name">link</span> <span class="hljs-attr">rel</span>=<span class="hljs-string">"canonical"</span> <span class="hljs-attr">href</span>=<span class="hljs-string">"https://ampbyexample.com/Introduction/Hello_World"</span>&gt;`

Does this make sense?

sebastianbenz:
Yeah - you're right. The canonical link is automatically generated to make authoring samples easier. We should add it to the code sections as well.

PR #234, opened 2016-06-30
add canonical link to source files

### LINKED ISSUE #233: Add canonical rel tag to all examples
It sounds like <link rel="canonical" href=".." gets auto-generated in all ABE pages. It'd be good to have it as part of the code itself, so anyone trying to reproduce their own versions doesn't get errors.

sebastianbenz:
* generate canonical link when creating a new sample
* add canonical link to all existing source samples
* add 'change' task to gulpfile in case we need to again
  batch process all sample files

Fixes #233, #161, #14


sebastianbenz:
@kul3r4 PTAL

PR #245, opened 2016-07-08
Add template to amp-form

kul3r4:
`submit-error` and `submit-success` need to contain a `template` element that is rendered, otherwise they are not valid. The alternative is to add our own style to the div.


sebastianbenz:
LGTM

PR #248, opened 2016-07-12
add license to missing license files

kul3r4:
Some files are missing license which is required, but I am not sure it's the right way for the .sh files, please advise.
@sebastianbenz @juliantoledo PTAL


sebastianbenz:
LGTM

PR #273, opened 2016-07-20
Improve a4a samples

sebastianbenz:
* Change add sample image size to 300x250 without white frame
* Use code formatting for attribute doc


sebastianbenz:
@carlosfuertes PTAL

carlosfuertes:
LGTM

PR #274, opened 2016-07-21
Refactor backend structure

sebastianbenz:
* move all supporting backend features into /backend dir
* extract redirects into separate file
* parse redirect urls from json & fix urls
* extract cache handlers into separate file

### DIFF
--- a/backend/redirects.json
+++ b/backend/redirects.json
@@ -1,0 +1,74 @@
+[
+  {
+    "source": "/amp-accordion.html",
+    "target": "/components/amp-accordion/"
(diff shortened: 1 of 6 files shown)


sebastianbenz:
@kul3r4 PTAL

backend/redirects.json
```diff
@@ -0,0 +1,74 @@
+[
+	{
```
kul3r4: remove spaces before {

sebastianbenz: done

PR #278, opened 2016-07-22
use downloaded font

kul3r4:
Use a downloaded font for amp-form. This fix the currently broken sample.
@sebastianbenz PTAL


sebastianbenz:
LGTM

PR #279, opened 2016-07-22
Refactor experiment buttons

sebastianbenz:
* move script after button declaration
* avoid blinking button by disabling it first
* renamings


sebastianbenz:
@kul3r4 moved the fix into it's own PR. PTAL

PR #283, opened 2016-07-25
Don't cache amp-live-list updates

sebastianbenz:
Our service worker uses a one-behind caching model for all html requests
to the ampbyexample.com domain. This means amp-live-list update requests
are always one-behind as well.


sebastianbenz:
@cramforce unexpected side effect of using a service worker for caching and amp-live-list. 
@kul3r4 PTAL

PR #288, opened 2016-07-26
Fix enable experiment button

sebastianbenz:
Wait until DOM is loaded before updating the button state.


sebastianbenz:
@kul3r4 PRAL

PR #294, opened 2016-07-29
show experiment toggle in preview mode

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

templates/css/preview.css
```diff
@@ -10,11 +10,20 @@ p {
...
   background: #fff;
   box-shadow: 0 2px 2px 0 rgba(0,0,0,.14),0 3px 1px -2px rgba(0,0,0,.2),0 1px 5px 0 rgba(0,0,0,.12);
   display: block;
+  margin: 0 auto;
+  max-width: 800px;
+}
+
+#experimental-mode {
+  max-width: 800px;
+  margin: 16px auto;
```
kul3r4: I would have also right and left margin to stay aligned with the text below.

sebastianbenz: I've set the left/right margin to auto in order to align it with the article below.

kul3r4: I tried to run it and it doesn't add any right and left margin.

sebastianbenz:
Fixed amp-live-list samples along the way. PTAL

kul3r4:
LGTM

PR #319, opened 2016-08-05
Use correct github link for amp-live-list.go file

kul3r4:
After the server refactoring, all the .go files are now under the backend folder.


sebastianbenz:
LGTM

PR #322, opened 2016-08-08
add amp-analytics sample

sebastianbenz:
This sample gives an interactive demonstration of howamp-analytics work. We use a realtime dashboard to show which events are fired.

Each user has a separate dashboard which is identified by the AMP CID. The CID is read from the cookie which is not available on the AMP Cache. If the sample is loaded via the AMP Cache, the dashboard will show all user events.

The sample does not set or store any cookies.


sebastianbenz:
Demo: https://amp-by-example-staging.appspot.com/components/amp-analytics/

@kul3r4 PTAL

src/20_Components/amp-analytics.html
```diff
@@ -0,0 +1,222 @@
...
+        "vars": {
+          "eventId": "clickOnSpecialAnchor"
+        }
+      }
+    }
+  }
+  </script>
+  </amp-analytics>
+  <!--  Only one of these links will trigger the `clickOnSpecialAnchor` event. -->
+  <a id="anchor-id">a special link</a>.
```
kul3r4: What about using a list like line 141 ( <!-- Try it by clicking one of the links.....)? Or maybe remove the list from the above, just to be consistent.

src/20_Components/amp-analytics.html
```diff
@@ -0,0 +1,222 @@
...
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <!-- #### Setup -->
+  <!--  Import the amp-analytics component in the header. -->
+  <script async custom-element="amp-analytics" src="https://cdn.ampproject.org/v0/amp-analytics-0.1.js"></script>
+  <!--
+    We need `amp-iframe` to embed an anlytics dashboard.
```
kul3r4: Typo: anlytics -> analytics

sebastianbenz: done

src/20_Components/amp-list.html
```diff
@@ -35,44 +33,43 @@
...
   <!-- #### Using an existing template -->
   <!--
     The template can also be specified using an ID of an existing
     `template` element. This example references the template from the
     previous sample.
   -->
-  <ul>
-  <amp-list width=auto height=100 layout=fixed-height src="<%host%>/json/examples.json"
+  <amp-list width=300 height=100 layout=responsive
+      src="https://ampbyexample.com/json/examples.json"
```
kul3r4: why not using <%host%> here?

kul3r4:
I think this is one of the coolest sample I have ever seen! Left few comments.

src/20_Components/amp-analytics.html
sebastianbenz: Good point. I actually have to use a list here, otherwise the second link won't be visible in the mobile version.

sebastianbenz: done

src/20_Components/amp-list.html
sebastianbenz: The changes to this file actually don't make any sense. I reverted the changes.

sebastianbenz:
addressed the comments. thanks! PTAL

kul3r4:
LGTM

PR #324, opened 2016-08-10
Wrap comments...

dandv:
... so they're visible [on GitHub](https://github.com/ampproject/amp-by-example/blob/master/src/20_Components/amp-form.html) without scrolling horizontally.


sebastianbenz:
LGTM

PR #327, opened 2016-08-11
Don't explicitly add anchors

### LINKED ISSUE #83: Make anchor links order independent
Currently anchor links depend on the order (example1, example2,...). These will break if a new section is added in between. We should use a slugged version of the content instead.

sebastianbenz:
They don't work half of the time and markdown already automatically
inserts tag ids for headings. Fixes #83


sebastianbenz:
@kul3r4 PTAL

kul3r4:
LGTM

Issue #270, opened 2016-07-16
Sample comment section using amp-form


sebastianbenz:
Build a sample comment section using amp-form.

//cc @kul3r4

PR #334, opened 2016-08-18
Add comment sample

kul3r4:
Solves #270 

@rowan-m PTAL


src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,110 @@
...
+  "preview": "default",
+  "default": "preview",
+  "draft": true,
+  "experiment": true,
+  "component": "amp-form"
+  }--->
+<!--
+#### Introduction
+
+This sample showcases how to build a comment section in AMP HTML.
```
sebastianbenz: can you explain how it works and which components we are using?

sebastianbenz: Can you also provide some guidance regarding login? How should this be handled?

kul3r4: If I provide guidance, does it mean I need to implement it in this sample? Or should i link to https://github.com/ampproject/amp-publisher-sample?

sebastianbenz: A comment section only makes sense together with a login. We should do this also in the sample. Is amp-access the recommended way to integrate a login?

backend/comment-section.go
```diff
@@ -0,0 +1,110 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend
+
+import (
+	"fmt"
+	"net/http"
+  "google.golang.org/appengine/datastore"
```
rowan-m: Indentation looks out. Does this need `gofmt` running on it?

rowan-m:
Adding some comments here to just summarise some discussion.

- If real comments are accepted, there's a lot of overhead. Input has to be sanitized, content has to be moderated, storage has to be maintained, etc. To me, the ROI in terms of the sample produced isn't enough to justify that. The backend code is not what people are really coming to look at.

- One option would be to just have a static set of dummy comments and a form that when submitted shows another dummy comment being added.
  comment
  The new comment only appears as the XHR response. Reloading the page goes back to to the static comments. The point of the sample is to demonstrate the potential UX.

- Alternatively, the potential input could be greatly restricted. e.g. radio buttons / dropdown allowing the user to select from a few canned responses. These could be stored in the backend with a date and a random gravatar img or similar. Stored responses would just be FIFO limited to 10 or similar.

kul3r4:
@rowan-m I tried the first suggested approach, PTAL.

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,114 @@
...
+    <p>[[.Text]]</p>
+</div>
+[[else]]
+<p>No comments found.</p>
+[[end]]
+</div>
+</div>
+
+    <!--
+    #### Analytics
```
rowan-m: I don't this this adds to the comment sample.

backend/comment-section.go
```diff
@@ -0,0 +1,92 @@
...
+	comments = append(comments, lastComment)
+	comments = append(comments, firstComment)
+	return CommentSection{Comments: comments}
+}
+
+func submitCommentXHR(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("AMP-Access-Control-Allow-Source-Origin", buildSourceOrigin(r.Host))
+	w.Header().Set("Content-Type", "application/json")
+	response := ""
+	text := r.FormValue("text")
```
rowan-m: If a blank submission gets through, this returns a 400 and it doesn't look like the page handles it correctly. An empty comment should trigger the `submit-error` block

kul3r4: Will look again at this when I clarify how this behaviour works ([reference](https://github.com/ampproject/amphtml/issues/4723))

kul3r4:
@sebastianbenz @jmadler This is the comment sample in draft. You could add the login on top of that.

kul3r4:
@rowan-m @sebastianbenz I added a commit with a login flow before being able to add a comment. The login page is a non AMP page as it's not required it to be an AMP page and can be the same page a publisher already has.

backend/comment-section.go
```diff
@@ -0,0 +1,169 @@
...
+	} else {
+		w.WriteHeader(http.StatusBadRequest)
+	}
+}
+
+func submitComment(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusBadRequest)
+}
+
+func handleCommentAuthorization(w http.ResponseWriter, r *http.Request) {
```
rowan-m: Could you move the auth functions into a separate `amp-access.go`? You'd save me a bunch of time as I can just re-use it for the hotel sample. :D

static/login.html
```diff
@@ -0,0 +1,42 @@
+<!doctype html>
+
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+     <link rel="stylesheet" href="http://fonts.googleapis.com/css?family=Roboto:300,400,500,700" type="text/css">
+    <title>Login</title>
+    <style>
```
rowan-m: Make this a little prettier and consistent with some of the styles from https://github.com/ampproject/amp-by-example/blob/master/templates/css/styles.css

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,115 @@
...
+  <!--
+    Use the `amp-access` to realize a login flow. Learn more about amp-access [here](/components/amp-access). 
+    Use the `amp-form` component to realize a comment section that would be available only after a successfull login.
+  -->
+  <div>
+    <span amp-access="user = 'test-user'">
+          <a href="<%host%>/samples_templates/comment_section/logout" class="button button-primary space-left">Logout</a>
+    </span>            
+    <span amp-access="NOT user = 'test-user'" role="button" tabindex="0">
+          <a on="tap:amp-access.login" class="button button-primary space-left">Login</a>
```
rowan-m: Add a little explanatory text, something like "Please login to comment".

static/login.html
kul3r4: I could do it but I would need to duplicate the css and I was not really in favor of that. Do you have any idea for not duplicating it? The only way I thought would be to have have this page under the src folder and be an AMP page so that the relevant css could be added via the template. Any idea would be appreciated!

backend/comment-section.go
kul3r4: @rowan-m Have a look at the latest changes on the server side, especially hotel.go and amp-access.go

sebastianbenz:
Some high-level comments:

* the logout button leaves the preview page -> it should stay there
* I find it confusing that it's not possible to add multiple comments. I assume this is because we don't want to implement the necessary server logic. Could we instead then hide the input field and comment button?
* Aren't comments usually appended at the end? 
* s/Previous Comments/Comments/
* Can you fake a normal comment field with author name and profile picture?
* Explain your approach and how you're using the different components in the doc section.

kul3r4:
@sebastianbenz I am trying to use server side templating to get the path of the sample (if it's either comment_section or comment_section/preview). 
It works well for setting up the logout href, but it doesn't work for the authorization endpoint, where the string is inside a json.
This is what happens:
"authorization": "/samples_templates/comment_section\/preview/authorization?rid=R....
where, server side, I am setting Path= "comment_section/preview" and in the template 
        "authorization": "/samples_templates/[[.Path]]/authorization?rid=READER_.....
As you can see, a backslash is added and the endpoint is not correct.
Have you got any idea how to solve this? I have been playing with it a bit with no luck; let me know if you have any idea, the latest code is committed.

sebastianbenz:
Is it possible that the backslash is added to escape the trailing slash?

sebastianbenz:
You can try if that's the cause by marking the path as HTML, e.g.  `template.HTML(path)`

kul3r4:
@sebastianbenz tried template.html and template.URL, both same problem, look at here: https://play.golang.org/p/akuGK1VyVJ

sebastianbenz:
Go's security model doesn't trust template parameters, you can either:

*  pass the whole string as a parameter: https://play.golang.org/p/9iTS5Ss6tc

or 

* pass the whole json object: https://play.golang.org/p/q6udq8MiZc

kul3r4:
Thanks, I was trying to keep as much code as possible in the HTML so that the sample would be clearer, but if that is not possible, then I will pick one of the 2 approaches.

kul3r4:
@sebastianbenz Addressed previous comments, PTAL

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	"fmt"
+	"html/template"
+	"net/http"
+	"path"
+	"time"
+)
+
+const (
+	COMMENT_SAMPLE_PATH                  = "/samples_templates/comment_section/"
+	COMMENT_SAMPLE_PATH_PREVIEW          = "/samples_templates/comment_section/preview/"
```
sebastianbenz: COMMENT_SAMPLE_PATH_PREVIEW = COMMENT_SAMPLE_PATH + "preview/"

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	UserImg  string
+}
+
+type AccessData struct {
+	ReaderID  string
+	ReturnURL string
+}
+
+func InitCommentSection() {
+	http.HandleFunc(COMMENT_SAMPLE_PATH+"submit-comment-xhr", func(w http.ResponseWriter, r *http.Request) {
```
sebastianbenz: Extract "submit-comit-xhr" and "submit-comment" into const.

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d, public, must-revalidate", MAX_AGE_IN_SECONDS))
+	t.Execute(w, CommentPage{Comments: comments, Path: sampleName,
+		AuthorizationEndpoint: template.JSStr(path.Join(SAMPLE_TEMPLATE_FOLDER, sampleName, AUTHORIZATION_ENDPOINT_SPECIFICATION)),
+		LoginEndpoint:         template.JSStr(path.Join(SAMPLE_TEMPLATE_FOLDER, sampleName, LOGIN_ENDPOINT_SPECIFICATION))})
+}
+
+func listComment(sampleName string, r *http.Request) []Comment {
+	var comments []Comment
+	firstComment := Comment{"This is the first comment", "user1", time.Now().Add(time.Duration(MINUS_TEN_SECONDS) * time.Second).Format("15:04:05"), USER_IMG}
+	lastComment := Comment{"This is the second comment", "user2", time.Now().Format("15:04:05"), USER_IMG}
```
sebastianbenz: nit: s/lastComment/secondComment/

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	comments = append(comments, firstComment)
+	comments = append(comments, lastComment)
+	return comments
+}
+
+func submitCommentXHR(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("AMP-Access-Control-Allow-Source-Origin", buildSourceOrigin(r.Host))
```
sebastianbenz: Somewhere we have a helper function to enable CORS.

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,135 @@
...
+            "access": false
+        }
+    }
+    </script>
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
+  <!--
+    Use the `amp-access` to realize a login flow. Learn more about amp-access [here](/components/amp-access). 
+    Use the `amp-form` component to realize a comment section. Learn more about amp-fomr [here](/components/amp-form).
```
sebastianbenz: s/fomr/form/

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,135 @@
...
+        "login": [[.LoginEndpoint]],
+        "authorizationFallbackResponse": {
+            "error": true,
+            "access": false
+        }
+    }
+    </script>
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
```
sebastianbenz: Can you please explain in more detail what's happening.

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	"time"
+)
+
+const (
+	COMMENT_SAMPLE_PATH                  = "/samples_templates/comment_section/"
+	COMMENT_SAMPLE_PATH_PREVIEW          = "/samples_templates/comment_section/preview/"
+	MINUS_TEN_SECONDS                    = -10
+	COMMENT_COOKIE_NAME                  = "ABE_EMAIL"
+	AUTHORIZATION_ENDPOINT_SPECIFICATION = "authorization?rid=READER_ID&url=CANONICAL_URL&ref=DOCUMENT_REFERRER&_=RANDOM"
+	LOGIN_ENDPOINT_SPECIFICATION         = "login?rid=READER_ID&url=CANONICAL_URL"
```
sebastianbenz: why not directly define this in the sample file?

backend/comment-section.go
```diff
@@ -0,0 +1,196 @@
...
+	USER_IMG                             = "/img/ic_account_box_black_48dp_1x.png"
```
sebastianbenz: why not directly define this in the sample file?

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,135 @@
...
+    <span amp-access="NOT user = 'test-user'" role="button" tabindex="0">
+          <h5>Please login to comment</h5>
+          <a on="tap:amp-access.login" class="button button-primary space-left">Login</a>  
+    </span>
+    <span amp-access="user = 'test-user'" amp-access-hide>    
+      <form method="post" action="<%host%>/samples_templates/comment_section/submit-comment" action-xhr="<%host%>/samples_templates/comment_section/submit-comment-xhr"  target="_top" >
+      <div class="full-width">
+        <h5 class="nopadding">Comments</h5>
+        <div>
+          [[range .Comments]]
```
sebastianbenz: Why don't you directly define the sample components in the sample? We don't need the template bit here.

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,135 @@
...
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
+  <!--
+    Use the `amp-access` to realize a login flow. Learn more about amp-access [here](/components/amp-access). 
+    Use the `amp-form` component to realize a comment section. Learn more about amp-fomr [here](/components/amp-form).
+  -->
+  <div>
+    <span amp-access="user = 'test-user'" amp-access-hide>
+          <a href="<%host%>/samples_templates/[[.Path]]/logout" class="button button-primary space-left">Logout</a>
```
sebastianbenz: amp-access trick: define a [separate login endpoint](https://github.com/ampproject/amphtml/blob/master/extensions/amp-access/amp-access.md#login-page) for the logout. This way you can use variable substitutions for the logout as well and provide a return URL. This means you don't need to separate between preview and sample page in the backend.

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: if i remove the template, how will a get comment timestamps?

sebastianbenz: Why do you need comment timestamps? Can't you just manually define them in the template?

backend/comment-section.go
kul3r4: ok

backend/comment-section.go
kul3r4: done

backend/comment-section.go
kul3r4: done

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: done

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: done

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: done

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: done

backend/comment-section.go
kul3r4: done

backend/comment-section.go
kul3r4: done

kul3r4:
@sebastianbenz , answers belows:
_Show comments to logged out users_
I moved the comments on top, so that they are outside the form component and can be shown to the logged out users. 
_margin-top: 16px to_
done
_I'd prefer to have a single amp-access server endpoint which we can use across all samples_
at the moment amp-access contains already some logic for the authorization endpoint related to the Hotel sample. The logic there is different from this sample. Me and @rowan-m already agreed that the Hotel sample could use the same code I am adding, but this will require some changes in the hotel sample code: for this reason this could be part of another PR.

sebastianbenz:
"at the moment amp-access contains already some logic for the authorization endpoint related to the Hotel sample. The logic there is different from this sample. Me and @rowan-m already agreed that the Hotel sample could use the same code I am adding, but this will require some changes in the hotel sample code: for this reason this could be part of another PR."

A start would be to use the hotel sample and add an additional flag for your sample: `"loggedId": true`. This has the other advantage that you can simplify your amp-access contraints to:

```
<span amp-access="loggedId" amp-access-hide>
...
</span>
```

kul3r4:
I don't understand the suggestion. Do you mean removing `span amp-access="user = 'test-user'"` and using `<span amp-access="loggedId"`?

sebastianbenz:
Yes. Testing for a specific user doesn't really make sense - comments should be open to everyone that is loggedIn.

kul3r4:
@sebastianbenz I made the suggested change and now using <span amp-access="loggedId" amp-access-hide>

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,144 @@
...
+          </div>
+          <div class="comment-user">
+            <amp-img width="44" height="44" alt="user" src="/img/ic_account_box_black_48dp_1x.png"></amp-img>
+            <div class="card comment">
+              <p class= "date">user2 20:26:50</p>
+              <p>This is the second comment</p>
+            </div>
+          </div>
+    </div> 
+    <span amp-access="loggedId" role="button" tabindex="0" amp-access-hide>
```
sebastianbenz: s/loggedId/loggedIn/

sebastianbenz: Why do we need a separate span here? And the others?

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,144 @@
...
+            "error": true,
+            "access": false
+        }
+    }
+    </script>
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
+  <!--
+    Use the `amp-access` to realize a login flow. After clicking on the `login` button, you can use the user `test-user` to perform a successfull login. You can then logout by clicking on the `logout` button. Learn more about amp-access [here](/components/amp-access).
```
sebastianbenz: Can you comment the different elements?

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,144 @@
...
+    </script>
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
+  <!--
+    Use the `amp-access` to realize a login flow. After clicking on the `login` button, you can use the user `test-user` to perform a successfull login. You can then logout by clicking on the `logout` button. Learn more about amp-access [here](/components/amp-access). 
+    Use the `amp-form` component to realize a comment section. Learn more about amp-form [here](/components/amp-form).
+  -->
+  <div>
+    <h5 class="padding-top">Comments</h5>
```
sebastianbenz: padding-top and padding-left are bad names for css rules. -> wrong level of abstraction

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,144 @@
...
+    <div class="padding-left">
```
sebastianbenz: why do we need this div?

backend/comment-section.go
```diff
@@ -0,0 +1,173 @@
+// Copyright Google Inc.
```
sebastianbenz: Why do we still need this file? Can't we get away without it? Or restrict to only the comment form endpoints.

kul3r4: Yes, the Hotel sample has a different logic and I think each sample should be able to specify its own logic. What do you mean "restrict to only the comment form endpoint"? It's handling only requests for "/samples_templates/comment_section/", so I believe it's already restricted.

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: Which structure would you suggest if I remove the span?

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: removed

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: changed

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: Not sure what else should be added, please suggest anything you have in mind.

kul3r4:
@sebastianbenz PTAL

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,142 @@
...
+              <p class= "date">{{user}} {{datetime}}</p>
+              <p>{{text}}</p>
+            </div>
+          </div>
+        </template>
+      </amp-list>
+
+    <!-- In order to add comments users need to be logged in. We use `amp-access` to integrate login. `on="tap:amp-access.login-sign-in"` specifies which 
+    action should be taken when clicking on the login button: `login` defines the property inside the `amp-access` json configuration, while `sign-in` defines
+    the endpoint. After clicking on the login button, a popup will appear; after clicking the displayed login button, the popup itself will disappear and the user will come back to sample page. We realize this behaviour by specifying an hidden input `return` with the value of [RETURN_URL](https://github.com/ampproject/amphtml/blob/master/extensions/amp-access/amp-access.md#access-url-variables) in the [login](https://github.com/ampproject/amp-by-example/tree/master/static/login.html) page; the AMP runtime will take care of the rest. -->
```
sebastianbenz: This bit is not needed here: 

```
After clicking on the login button, a popup will appear; after clicking the displayed login button, the popup itself will disappear and the user will come back to sample page. We realize this behaviour by specifying an hidden input `return` with the value of [RETURN_URL](https://github.com/ampproject/amphtml/blob/master/extensions/amp-access/amp-access.md#access-url-variables) in the [login](https://github.com/ampproject/amp-by-example/tree/master/static/login.html) page; the AMP runtime will take care of the rest.`
```

This is normal amp-access behavior. That's not something you need to explain here. What you need to explain is how we're using amp-access here. Namely that we use it to show and hide the login button depending on whether the user is logged in.

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,142 @@
...
+    the endpoint. After clicking on the login button, a popup will appear; after clicking the displayed login button, the popup itself will disappear and the user will come back to sample page. We realize this behaviour by specifying an hidden input `return` with the value of [RETURN_URL](https://github.com/ampproject/amphtml/blob/master/extensions/amp-access/amp-access.md#access-url-variables) in the [login](https://github.com/ampproject/amp-by-example/tree/master/static/login.html) page; the AMP runtime will take care of the rest. --> 
+    <span amp-access="NOT loggedIn" role="button" tabindex="0" amp-access-hide>
+          <h5>Please login to comment</h5>
+          <a on="tap:amp-access.login-sign-in" class="button button-primary comment-button">Login</a>  
+    </span>    
+      <!-- Similar to the sign-in, the sign-out flow is specified by `on="tap:amp-access.login-sign-out"`. On the contrary of the sign-in, after clicking on the logout button the user stays on the same page: we realize this by redirecting to the URL specified by the `return` query parameter added by AMP runtime and adding `#success=true`. -->
```
sebastianbenz: same as above: " Similar to the sign-in, the sign-out flow is specified by `on="tap:amp-access.login-sign-out"`. On the contrary of the sign-in, after clicking on the logout button the user stays on the same page: we realize this by redirecting to the URL specified by the `return` query parameter added by AMP runtime and adding `#success=true`." is not needed here.

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,142 @@
...
+    <style amp-custom>
+    .full-width {
+      width: 100%;
+    }
+    form.amp-form-submit-success [submit-success] {
+        width: 100%;
+        color: black;
+    }
+    form.amp-form-submit-success .data-input,
+    form.amp-form-submit-success .img-data-input,
```
sebastianbenz: Is there a way you can avoid declaring img-data-input?

kul3r4:
@sebastianbenz PTAL.
I noticed that the after login-insert comment-logout-login the previous comment appears, while reloading the page will get the right behaviour.
The login.html is loading something like:
`
<input name="returnurl" type="hidden" value="https://cdn.ampproject.org/v0/amp-login-done-0.1.html?url=http%3A%2F%2Flocalhost%3A8080%2Fsamples_templates%2Fcomment_section%2Fpreview%2F">
`
will I need a way to flag that this should not be cached?

backend/comment-section.go
```diff
@@ -0,0 +1,141 @@
...
+	http.HandleFunc(COMMENT_SAMPLE_PATH_PREVIEW+SUBMIT_COMMENT_XHR, func(w http.ResponseWriter, r *http.Request) {
+		handlePost(w, r, submitCommentXHR)
+	})
+	http.HandleFunc(COMMENT_SAMPLE_PATH_PREVIEW+SUBMIT_COMMENT, func(w http.ResponseWriter, r *http.Request) {
+		handlePost(w, r, submitComment)
+	})
+
+	http.HandleFunc(COMMENT_SAMPLE_PATH+"authorization", handleCommentAuthorization)
+	http.HandleFunc(COMMENT_SAMPLE_PATH_PREVIEW+"authorization", handleCommentAuthorization)
+	http.HandleFunc(COMMENT_SAMPLE_PATH+"login", handleCommentLogin)
```
sebastianbenz: why do you need different routes for login/logout/auth?

backend/comment-section.go
```diff
@@ -0,0 +1,141 @@
...
+	Datetime string
+	UserImg  string
+}
+
+type AccessData struct {
+	ReturnURL  string
+}
+
+func InitCommentSection() {
+	http.HandleFunc(COMMENT_SAMPLE_PATH+SUBMIT_COMMENT_XHR, func(w http.ResponseWriter, r *http.Request) {
```
sebastianbenz: Do we still need these?

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,139 @@
...
+        "authorizationFallbackResponse": {
+            "error": true,
+            "loggedIn": false
+        }
+    }
+    </script>
+    </head>
+    <body>
+    <!-- #### Add a Comment Section -->
+    <h5 class="comments-heading">Comments</h5>
```
sebastianbenz: Add `<!-- -->` above this line. Otherwise the headline will be shown as a code sample in mobile view. 

Also: why do you define a new rule `comments-heading`? I'd prefer if you use the default styling for things like headlines etc.

backend/comment-section.go
kul3r4: yes

backend/comment-section.go
kul3r4: Done

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: Done

backend/comment-section.go
sebastianbenz: But why for previews as well?

src/50_Samples_%26_Templates/Comment_Section.html
```diff
@@ -0,0 +1,137 @@
...
+    }
+    .comment {
+        background: #f2f2f2;
+        margin: 0 16px 16px 0;
+        width: 100%;
+    }
+    .button-primary {
+      margin: 16px
+    }
+    .user-image {
```
sebastianbenz: s/image/avatar

backend/comment-section.go
```diff
@@ -0,0 +1,137 @@
...
+	"time"
+)
+
+const (
+	COMMENT_SAMPLE_PATH         = "/samples_templates/comment_section/"
+	COMMENT_SAMPLE_PATH_PREVIEW = COMMENT_SAMPLE_PATH + "preview/"
+	COMMENT_COOKIE_NAME         = "ABE_LOGGED_IN"
+	SUBMIT_COMMENT              = "submit-comment"
+	SUBMIT_COMMENT_XHR          = "submit-comment-xhr"
+	USER                   			= "Charlie"
```
sebastianbenz: can you run go-format

src/50_Samples_%26_Templates/Comment_Section.html
kul3r4: done

backend/comment-section.go
kul3r4: done

kul3r4:
Applied changes, I noticed that there was some extra padding when a comment was added introduced by the form, so I added a css rule.

sebastianbenz:
LGTM - great sample!

PR #335, opened 2016-08-22
Removes amp-experiment draft

kul3r4:
After the fix of [this bug](https://github.com/ampproject/amphtml/pull/4498/files), we should add this sample to the main page.


sebastianbenz:
Lgtm

PR #338, opened 2016-08-24
Made class name consistent with experiment name.

tomayac:
Consistently using “button-color-experiment” for class name *and* experiment name instead of an additional “button-experiment” class name.


kul3r4:
LGTM

kul3r4:
Thank you very much!

PR #341, opened 2016-08-30
Correct attribute names and default values

rowan-m:
- Corrects a typo: `alue` → `value`
- Correctly sets default value for `datetime-local`

@kul3r4 please have a peek!


kul3r4:
LGTM

PR #342, opened 2016-08-31
Separate experiments from component names

rowan-m:
Currently there's an assumption built in that there is a 1:1 relationship between the `component` of a page and the experiment required to properly run it. This breaks the `amp-form` example as it needs both the `amp-form` and `form-submit` experiments enabled to prevent form submission with invalid data.

These changes:
- Add a new "experiments" metadata key to specify the AMP experiments to be enabled for a given page.
- Update the experiment page to enable all specified experiments.
- Update the samples to specify individual experiments required.

Open to discussion on this one, because having the `experiment` and the `experiments` metadata feels a bit redundant, but it makes the makes the mustache a bit cleaner. I was fiddling around trying to just set `experiment` to `true` if `experiments` was set, but `gulp` wasn't cooperating.

@kul3r4 @sebastianbenz please dive in and see what you think.

### DIFF
--- a/templates/experiment.html
+++ b/templates/experiment.html
@@ -1,9 +1,9 @@
 <div id="experimental-mode" class="card info">
   <h4>Experimental Mode</h4>
-  <p>The <code>{{component}}</code> component is currently in experimental mode and needs to be manually enabled (<a href="https://www.ampproject.org/docs/reference/experimental.html">more info</a>).</p>
+  <p>This example uses the following experimental feature{{#metadata.experiments.1}}s{{/metadata.experiments.1}}: <code class="amp-experiment-list">[{{#metadata.experiments}}<span>{{.}}</span>{{/metadata.experiments}}]</code><br>Enable them below and <a href="https://www.ampproject.org/docs/reference/experimental.html">learn more here</a>.</p>
 
   <div id="experiment-container">
-    <button id="experiment-toggle" class="button-inactive">Enable Experiment</button>
+    <button id="experiment-toggle" class="button-inactive">Enable Experiment{{#metadata.experiments.1}}s{{/metadata.experiments.1}}</button>
     <a id="canary-toggle" class="button button-inactive" href="https://cdn.ampproject.org/experiments.html" target="_blank">Enable Dev Channel</a>
   </div>
 </div>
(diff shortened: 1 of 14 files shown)


templates/experiment.html
```diff
@@ -1,9 +1,9 @@
 <div id="experimental-mode" class="card info">
   <h4>Experimental Mode</h4>
-  <p>The <code>{{component}}</code> component is currently in experimental mode and needs to be manually enabled (<a href="https://www.ampproject.org/docs/reference/experimental.html">more info</a>).</p>
+  <p>This example uses the following experimental features: <code class="amp-experiment-list">[{{#metadata.experiments}}<span>{{.}}</span>{{/metadata.experiments}}]</code><br>Enable them below and <a href="https://www.ampproject.org/docs/reference/experimental.html">learn more here</a>.</p>
```
kul3r4: features is always plural here

rowan-m:
@kul3r4 cleaned up after discussion.

templates/experiment.html
```diff
@@ -1,9 +1,9 @@
 <div id="experimental-mode" class="card info">
   <h4>Experimental Mode</h4>
-  <p>The <code>{{component}}</code> component is currently in experimental mode and needs to be manually enabled (<a href="https://www.ampproject.org/docs/reference/experimental.html">more info</a>).</p>
+  <p>This example uses the following experimental feature{{#metadata.experiments.1}}s{{/metadata.experiments.1}}: <code class="amp-experiment-list">[{{#metadata.experiments}}<span>{{.}}</span>{{/metadata.experiments}}]</code><br>Enable them below and <a href="https://www.ampproject.org/docs/reference/experimental.html">learn more here</a>.</p>
```
kul3r4: We are doing the check for the plural ({{#metadata.experiments.1}}s{{/metadata.experiments.1}}) but we are not doing it for "Enable them". If it's too complex check everytime before write the string, then using the array style would be better, otherwise let's be consistent.

rowan-m: @kul3r4 updated.

kul3r4:
LGTM, could you squash the commits?

templates/experiment.html
```diff
@@ -17,21 +17,34 @@
...
-    function isExperimentEnabled() {
-      return AMP.isExperimentOn('{{component}}');
+    function isExperimentEnabled(experimentName) {
+      return AMP.isExperimentOn(experimentName);
     }
 
-    function enableExperiment() {
-      if (!isExperimentEnabled()) {
-        AMP.toggleExperiment('{{component}}');
+    function areAllExperimentsEnabled() {
```
sebastianbenz: Can we use Array.prototype.every()?

sebastianbenz: Doesn't make sense as we're code generating...

rowan-m: I could inline the `AMP.isExperimentOn()` call to remove `isExperimentEnabled()`. Depends what you feel is more readable.

sebastianbenz:
Can you update the readme as well?

rowan-m:
Previous commits squashed, `README` updates on the end.

kul3r4:
LGTM

PR #346, opened 2016-09-08
drawer menu: don't show default tap color on chrome

sebastianbenz:
Also: remove invalid css statement from first line & reformatting

### DIFF
--- a/templates/css/styles.css
+++ b/templates/css/styles.css
@@ -1,5 +1,4 @@
 /* reset */
-  everflow-wrap: break-word;
 figure, ul, ol {
   margin: 0;
   padding: 0;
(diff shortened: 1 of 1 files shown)


sebastianbenz:
@rowan-m can you take a look?

templates/css/styles.css
```diff
@@ -1,5 +1,4 @@
 /* reset */
-  everflow-wrap: break-word;
```
rowan-m: Is the intent to remove this or add the correct property back in?

sebastianbenz: the correct property is still there...this just needs to be removed

rowan-m:
LGTM

sebastianbenz:
Cheers!

Issue #355, opened 2016-09-13
ABE munging some CSS classes


jmadler:
ABE seems to be munging some CSS classes, such as those in #354

sebastianbenz:
Can you please provide samples for the conflicts.

jmadler:
I'm not sure that it's a rule in style.css that's munging -- it might be part of the templating process.  

Have a look at the "helper css class" sample from PR #354 and compare to [this gist](https://gist.github.com/jmadler/e9182dcf202dc4ec9058784d6dbf4bdb)

I'm unfamiliar with the templating process but if you'd like I can dive into this deeper to address it.

jmadler:
Closed as unreproducible as of now.

PR #357, opened 2016-09-13
Add initial amp layout sample

### LINKED ISSUE #352: The AMP Layout System

jmadler:
Resolves #352


jmadler:
Blocked by #355

jmadler:
Made some updates to the sample.

#355 is closed, but https://github.com/ampproject/amphtml/issues/4947 is still blocking this

sebastianbenz:
Would this work better if we wouldn't be using amp-fit-text? I'd like to
get this out as soon as possible as it is super valuable. I might also move
it to the introduction section.

jmadler:
I'm not sure this is a bug in amp-fit-text so much as a bug in the fill layout itself.  It might be possible to omit layout=fill for this round until the other bugs are fixed -- should I try that?

sebastianbenz:
Can you try without amp-fit-text? Maybe I simply didn't understand your
intention.

jmadler:
How would you propose I do this without amp-fit-text?

sebastianbenz:
e.g. use an amp-img. This also makes more sense, as all layouts should work well with images.

jmadler:
This sample uses both amp-img and amp-fit-text.  There are meaningful differences in how text and images resize according to different layout strategies.

I'm unaware of any reason to remove amp-fit-text.  If you're concerned about the display issue, we should cut the layout=fill section or push on the AMP team to fix it.  If you're not, we should move forward.

sebastianbenz:
Reasons why I don't think amp-fit-text is necessary:

* amp-fit-text is rarely used
* amp-fit-text brings it's own semantics which makes things more confusing
* amp-fit-text makes only sense for a limited set of layouts

This sample should be about understanding the different layout types. I think a simple amp-img works well for this use case (container might need a different type though). 

The interesting question is: how can we demonstrate the effect of different viewport sizes on the layout demos? In my experience resizing the browser window works best for understanding the layouts. Can we somehow simulate this?

Sidenote: I removed the fill layout sample, but most demos were still broken.

jmadler:
I don't agree.  The only semantics that amp-fit-text brings is that it applies the layout system to textual content; and while it's true that amp-fit-text only makes sense for a limited set of layouts, and is rarely used,  neither of those facts meaningfully impact its usefulness in this demo.  I think there's still value in demonstrating how text and images respond to different layout types and sizes.  In any event, I'll defer to your judgement here.

In demonstrating different viewport sizes, I'm not sure what the best approach might be.  Does amp-bind allow for some magic here (i.e. buttons that toggle different sizes for an outer container div)?  If not, maybe we can just suggest the user view on different screens.

Yeah, it's possible the bug isn't just with layout=fill -- there's a lot going on here and I haven't really dived into it too deeply.

jmadler:
To clarify: while images can scale in height, width, or both, text is a more fluid content that can change shape to fit its container without concern for its aspect ratio, which is a valid concern from selecting a layout strategy from a UI perspective.

sebastianbenz:
I'd prefer if we leave amp-fit-text out for now. Let's focus on demoing the basic layouts first. We can later add more sophisticated scenarios. 

Awesome idea with amp-bind! You could trigger a css class which scales & animates the container to a larger size.

jmadler:
SGTM.  I'll tweak it later this week.

jmadler:
Added the binding and moved the fill bits over.  PTAL

kul3r4:
@jmadler LGTM but before merging please take a look at the errors from the CI

PR #375, opened 2016-09-21
Fix amp-ad example to include amp-ad.js

### LINKED ISSUE #364: Fix amp-ad example to include amp-ad.js
https://ampbyexample.com/components/amp-ad/

Does not include amp-ad-0.1.js as an extension. It's currently a soft warning requirement, but we should update docs to reflect:
https://github.com/ampproject/amphtml/blob/master/builtins/amp-ad.md

sebastianbenz:
Fixes #364


sebastianbenz:
@kul3r4 PTAL

PR #377, opened 2016-09-23
remove unneeded animation prefixes

estelle:
Removed unused and unneeded browser prefixes. left -webkit- for older android and UC browser. 
updated text of alt attribute to be more descriptive.


sebastianbenz:
See my comment on the previous PR regarding the AMP boilerplate. I'm happy to reopen in case the boilerplate gets simplified.

PR #378, opened 2016-09-23
Updated CSS

estelle:
Removed unneeded prefix versions.
animation: name 0s steps(1,end) 1 0s normal both 
is the equivalent of 
animation: name step-end both
as iteration count defaults to 1, delay defaults to 0s, duration defaults to 0s, and direction defaults to normal. Using the shorthand, you get the defaults for undeclared values. step-end is a shorter way of writing steps(1, end)

left webkit prefix in for android and uc browser.


kul3r4:
@jridgewell What do you think?

jridgewell:
This will fail validation.

/cc @Gregable

sebastianbenz:
Closing this PR as the AMPHTML project is a better place to discuss improvements to the boilerplate. @estelle can you please file an issue for improving the AMP boilerplate there. Thanks!

PR #388, opened 2016-10-03
Fix ABE munging Body

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

sebastianbenz:
@jmadler FYI

kul3r4:
LGTM, just one thing I noticed: is it gonna work if `body` is in capital letters?

sebastianbenz:
nope - but we don't support uppercase tags anyway.

Issue #373, opened 2016-09-19
Make "View on Github" button more discoverable on component pages


spacedino:
Use case:
As a developer who is looking to directly copy and paste the examples, given I cannot highlight the source code because it is in a table, I need another way to get it. 

Current behavior:
I don't notice "View on Github" button is in the top right corner of the page. I have to ask someone. 

Expected behavior:
Make the CTA more prominent and closer to the source code where a user would be looking for it. Maybe change the text as well to something like "download source code from Github" ; "get source code" ; etc. 

I've attached what it could look like on a large screen and small.
![getcode_cta_lg](https://cloud.githubusercontent.com/assets/7339798/18652145/d4913038-7e84-11e6-9f98-40015a1bbac1.png)
![getcode_cta_sm](https://cloud.githubusercontent.com/assets/7339798/18652144/d4900258-7e84-11e6-916c-055e9e3f6548.png)

sebastianbenz:
@abeck000 I've (hopefully) improved the UI to make it easier to get to the full source code: there is now a "Edit in Playground" button. Sample: https://amp-by-example-staging.appspot.com/advanced/how_to_create_interactive_amp_pages/

Let me know what you think. Thanks!

kul3r4:
This is now solved with the last Ui update where "Edit on github" is immediately below the title.

PR #392, opened 2016-10-03
add preview button on cards

sebastianbenz:


sebastianbenz:
@kul3r4 what do you think?

templates/index.html
```diff
@@ -193,7 +197,12 @@ <h4 class="{{#experiment}}amp-experiment-medium{{/experiment}}">
               {{title}}
             </h4> 
             {{#description}} <p class="description">{{{ description }}}</p> {{/description}}
-            <a class="button" href="{{url}}">View Sample</a>
+            <div class="buttons">
+              <a class="button" href="{{url}}">View Sample</a>
+              {{#metadata.preview}}
+              <a class="button" href="{{urlPreview}}">View Preview</a>
```
kul3r4: Should we use "show" instead of "view" to be consistent to what we are already using on the top right of the samples with preview?

sebastianbenz:
Updated PR: 

* change wording: preview -> demo
* change wording: view on github -> view source (#373)

templates/header.html
```diff
@@ -5,15 +5,15 @@
...
       <amp-img srcset="/img/ic_play_circle_filled_white_24dp_1x 1x, /img/ic_play_circle_filled_white_24dp_2x.png 2x" width="24" height="24" alt="preview"></amp-img>
-      <span>Show Preview</span>
+      <span>View Demo</span>
     </a>
     {{/metadata.preview}}
     {{#github}}
     <a href="{{github}}" class="button" target="_blank">
       <amp-img srcset="/img/GitHub-Mark-Light-32px.png 1x, /img/GitHub-Mark-Light-64px.png 2x" width="24" height="24" alt="Github" attribution="github" ></amp-img>
-      <span class="hide-on-mobile">View on Github</span>
+      <span class="hide-on-mobile">View Source</span>
```
kul3r4: View Source sounds fine for me. We are also using a similar wording "View annotated Source" when we switch between preview and complete sample, so maybe we can change that to "View demo with code" or something similar?

sebastianbenz:
Fixed the build.

PR #398, opened 2016-10-07
match live blog sample with documentation

jkingyens:


kul3r4:
@jkingyens The `BlogPosting` are generated server side via template, your PR contains encoded `BlogPosting`. The required change is to something like:
`
"liveBlogUpdate": [
          {{range $index, $value := .BlogItems}}
         
            {
              "@type": "BlogPosting",
              "headline": "{{$value.Heading}}",
              "url": "<%host%>/samples_templates/live_blog/#{{$value.ID}}",
              "datePublished": "{{$value.MetadataTimestamp}}",
              "articleBody": {
                "@value": "{{$value.Text}}",
                "@type": "rdf:HTML"
              }
            }
          {{end}} //Add a check here to decide if comma is required`

but I can't get the implementation of checking if it's the last value (this is golang code in the template)

kul3r4:
LGTM

PR #403, opened 2016-10-08
Change start_url to index.html to work locally as well

dandv:
During local development, if start_url points to https://ampbyexample.com, Chrome will ignore it:

image

The common practice is to only point to [`index.html`](https://developers.google.com/web/updates/2014/11/Support-for-installable-web-apps-with-webapp-manifest-in-chrome-38-for-Android) or [`.`](https://developer.mozilla.org/en-US/docs/Web/Manifest).

### DIFF
--- a/static/manifest.json
+++ b/static/manifest.json
@@ -2,7 +2,7 @@
   "name": "AMP by Example",
   "short_name": "AMP Examples",
   "description": "Learn AMP by Examples",
-  "start_url": "https://ampbyexample.com",
+  "start_url": "index.html",
   "display": "standalone",
   "background_color": "#607D8B",
   "theme_color": "#607D8B",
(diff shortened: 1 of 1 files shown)


static/manifest.json
```diff
@@ -2,7 +2,7 @@
   "name": "AMP by Example",
   "short_name": "AMP Examples",
   "description": "Learn AMP by Examples",
-  "start_url": "https://ampbyexample.com",
+  "start_url": "index.html",
```
sebastianbenz: the sample in the docs uses `/index.html` - is this a problem?

sebastianbenz:
One question - otherwise LGTM.

static/manifest.json
dandv: Fixed, thanks.

sebastianbenz:
Cheers!

PR #408, opened 2016-10-10
Fix markup liveblog

kul3r4:
Creates metadata from liveBlogUpdate array from the server.
@sebastianbenz PTAL
@jkingyens FYI: This would fix the array problem that was making the metadata unvalid. I also added some of your suggestions from #398

### DIFF
--- a/backend/amp-live-list.go
+++ b/backend/amp-live-list.go
@@ -21,6 +21,7 @@ import (
 	"path"
 	"strconv"
 	"time"
+	"log"
 )
 
 const (
@@ -129,7 +130,10 @@ func createPage(newStatus int, timestamp time.Time) Page {
 }
 
 func renderSample(w http.ResponseWriter, r *http.Request, filePath string) {
-	t, _ := template.ParseFiles(filePath)
+	// t := template.New("template test")
+	// t = template.Must(template.ParseFiles(filePath))
+	t, error := template.ParseFiles(filePath)
+	log.Println(error)
 	w.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d, public, must-revalidate", MAX_AGE_IN_SECONDS))
 	newStatus := updateStatus(w, r)
 	t.Execute(w, createPage(newStatus, time.Now()))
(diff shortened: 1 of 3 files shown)


backend/amp-live-list.go
```diff
@@ -119,20 +149,38 @@ func readStatus(r *http.Request) int {
...
+}
+
+func createPage(newStatus int, timestamp time.Time, r *http.Request) Page {
 	if newStatus > len(blogs) {
 		newStatus = len(blogs)
 	}
 	blogItems := getBlogEntries(newStatus, timestamp)
 	score := createScore(newStatus, 0)
-	return Page{BlogItems: blogItems, FootballScore: score}
+	metadata, _ := json.Marshal(createMetadata(r))
```
sebastianbenz: I think you can simply pass the json directly to the template.

backend/amp-live-list.go
```diff
@@ -119,20 +149,38 @@ func readStatus(r *http.Request) int {
...
+	return Page{BlogItems: blogItems, FootballScore: score, BlogMetadata: string(metadata)}
 }
 
 func renderSample(w http.ResponseWriter, r *http.Request, filePath string) {
-	t, _ := template.ParseFiles(filePath)
+	t, error := template.ParseFiles(filePath)
+	log.Println(error)
```
sebastianbenz: correctly handle the error, `panic(error)` is OK if you parse the template on startup -> don't parse the files with every request, instead cache the parsed template.

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -52,20 +53,7 @@
...
-              "url": "<%host%>/samples_templates/live_blog/#{{.ID}}",
-              "datePublished": "{{.MetadataTimestamp}}",
-              "articleBody": {
-                "@value": "{{.Text}}",
-                "@type": "rdf:HTML"
-              }
-            }
-          ]
-          {{end}}
+          "liveBlogUpdate": {{.BlogMetadata}}
```
sebastianbenz: nice! I was afraid you'd needed to convert the whole json.

jkingyens:
@kul3r4 Thank you for the metadata fix. But with regards to formatting suggestions in #398 but please keep PRs separate in the future.

PR #412, opened 2016-10-10
fix server side rendering of meta-data in live blog

jkingyens:
The live blog sample is rendering some metadata as a string which makes it invalid clientside. This renders the metadata correctly so that it is valid JSON-LD.


dandv:
LGTM, can you post before&after screenshots of the JSON+LD section?

jkingyens:
screen shot 2016-10-10 at 3 38 19 pm

screen shot 2016-10-10 at 3 39 04 pm

PR #414, opened 2016-10-11
Add file for amp-google-vrview-image sample

amedina:
Simple example with single amp-google-vrview-image component.
PTAL.

### DIFF
--- a/src/20_Components/amp-google-vrview-image.html
+++ b/src/20_Components/amp-google-vrview-image.html
@@ -1,0 +1,47 @@
+<!---{
+    "experiments": ["amp-google-vrview-image"],
+    "preview": "default"
+}--->
+<!--
+  #### Introduction
+  The [amp-google-vrview-image](https://www.ampproject.org/docs/reference/components/amp-google-vrview-image) extension allows embedding 360 degree VR media into AMP pages.
+-->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <link rel="canonical" href="<%host%>/components/amp-google-vrview-image/">
+  <!-- #### Setup -->
+  <!--
+    Import the AMP library.
+  -->
+  <script 
+    async 
+    src="https://cdn.ampproject.org/v0.js">    
+  </script>
+  <!--
+    Import the AMP component.
+  -->  
+  <script 
+    async 
+    custom-element="amp-google-vrview-image" 
+    src="https://cdn.ampproject.org/v0/amp-google-vrview-image-0.1.js">        
+  </script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+<!--
+  Include the AMP boilerplate.  
+-->
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+</head>
+<body>
+<article>
+    <!-- ##### Define the VR view component -->
+    <amp-google-vrview-image
+        src="https://storage.googleapis.com/vrview/examples/coral.jpg"
+        stereo
+        width="400" height="300" layout=responsive>
+    </amp-google-vrview-image>
+</article>
+</body>
+</html>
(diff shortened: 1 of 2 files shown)


amedina:
Makes sense; I will add other samples in the same file, along the lines of the amp-form sample.

src/20_Components/amp-google-vrview-image.html
```diff
@@ -0,0 +1,47 @@
...
+  </script>
+  <!--
+    Import the AMP component.
+  -->  
+  <script 
+    async 
+    custom-element="amp-google-vrview-image" 
+    src="https://cdn.ampproject.org/v0/amp-google-vrview-image-0.1.js">        
+  </script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
```
kul3r4: where amp-mustache is used?

amedina: Where is it needed?

src/20_Components/amp-google-vrview-image.html
```diff
@@ -0,0 +1,64 @@
...
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+<!--
+  Include the AMP boilerplate.  
+-->
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+</head>
+<body>
+<article>
+    <!-- **VR view component.** [**stereo**, **yaw=**0 (default)]-->
+    <h4>Google VR View Image [stereo, yaw=0]</h4>
```
sebastianbenz: Don't include headings in the code, use doc comments instead. In comments you can use markdown formattinge, e.g.:

```
<!-- #### Basic Usage -->
<!-- By default `amp-google-vrview-image` expects a mono image. The yaw is 0. -->
<amp-google-vrview-image
        src="https://storage.googleapis.com/vrview/examples/coral.jpg"
        width="400" height="300" layout=responsive>

<!-- Add the `stereo` attribute for stereo images. The yaw (what does it mean?) can be defined via the `yaw` attributes. -->

...
```

It's a good approach to have a first sample with the defaults and then add one or two other samples demonstrating different attribute values.

amedina: The image used is the same for the three instances.

Will make the changes to use markup instead of headings.

amedina: Going to ask Anirudh to see if a different image should be used between the two modes.

amedina: You are right; need to find two versions. Will do. Learning VR along the way! :)

amedina: Hey guys, I am confused; I think the image used is the same and the component should set appropriate metadata to differentiate the mode. In the regular vrview approach, this is done in the src attribute as this:

**src="vrview/index.html?image=/images/converted.jpg&is_stereo=true"**

I thought the amp component should do this under the hood. Will dive a little bit into the implementation of the component to figure it out.

amedina:
Conclusions so far: using the same image from the VR Viewers documentation; depending on the attributes used the image is treated as monoscopic or stereoscopic.  There are four cases with combinations of the parameters. In order to truly see the difference between the specific cases, the sample has to be experienced with a cardboard viewer (it is vrview sample after all). I specified that in the intro section to the component.

src/20_Components/amp-google-vrview-image.html
```diff
@@ -0,0 +1,76 @@
...
+</head>
+<body>
+<article>
+    <!-- #### Basic Usage -->
+    <!-- By default `amp-google-vrview-image` expects a mono image and the yaw value is 0. -->
+    <amp-google-vrview-image
+        src="https://storage.googleapis.com/vrview/examples/coral.jpg"
+        width="400" height="300" layout=responsive>
+    </amp-google-vrview-image>
+    <!-- Add the `stereo` attribute for stereoscopic image -->
```
sebastianbenz: s/image/images./

src/20_Components/amp-google-vrview-image.html
```diff
@@ -0,0 +1,76 @@
...
+    <amp-google-vrview-image
+        src="https://storage.googleapis.com/vrview/examples/coral.jpg"
+        stereo
+        width="400" height="300" layout=responsive>
+    </amp-google-vrview-image>
+    <!-- Add the `stereo` attribute for stereoscopic image and an initial yaw value of 90 degrees -->
```
sebastianbenz: s/image/images/
s/and an/. The/
s/of/is/

src/20_Components/amp-google-vrview-image.html
```diff
@@ -0,0 +1,76 @@
...
+    <amp-google-vrview-image
+        src="https://storage.googleapis.com/vrview/examples/coral.jpg"
+        stereo
+        yaw=90
+        width="400" height="300" layout=responsive>
+    </amp-google-vrview-image>
+    <!-- Add the `stereo` attribute for stereoscopic image and restrict motion around the yaw axis only -->
```
sebastianbenz: suggestion: " Use the `yaw-only` attribute to restrict motion to the yaw axis only."

amedina: Done!

sebastianbenz:
Thanks!

Issue #416, opened 2016-10-12
Remove amp-form non-XHR post sample


sebastianbenz:
non-xhr post is not allowed: 

"non-XHR POST requests are not allowed in AMP documents. This is due to inconsistencies of setting Origin header on these requests across browsers. And the complications supporting it would introduce in protecting against XSRF. This might be reconsidered and introduced later, please file an issue if you think this is needed."

https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#security-considerations

sebastianbenz:
We have to re-add the action attribute until #5607 is fixed.

PR #418, opened 2016-10-12
AMP form security improvements

kul3r4:
- Removes action where POST method is required
- improves samples to use GET only for info retrieval and search
This addresses #416, we still need to work on improving the shopping cart implementation for the Product sample, but will do in a separate PR.

@sebastianbenz PTAL
@rowan-m FYI In the Hotel sample,  I moved POST -> GET  for Check availability and it seems it is an info retrieval. I added an XHR endpoint for the BOOK button given the action is a POST.

### DIFF
--- a/src/20_Components/amp-form.html
+++ b/src/20_Components/amp-form.html
@@ -54,7 +54,7 @@
 
   <!-- #### Form Submission with Page Reload -->
   <!--
-    Use the `action` attribute to specify a server-endpoint which should handle the form input. The URL must be HTTPS.
+    Use the `action` attribute to specify a server-endpoint which should handle the form input. This works only for `GET` requests. The URL must be HTTPS.
     The following sample uses `type="search"` to emulate a search. The search button will trigger a page reload.
 
     Please find [security considerations](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#security-considerations) to check best practices when choosing `GET` or `POST` methods.
@@ -76,7 +76,7 @@
     The `amp-form` component displays `submit-success` or `submit-error` elements based on the response and
     renders the template data inside these two elements.
   -->
-  <form method="post" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr"  target="_top">
+  <form method="post" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr">
     <input type="text" class="data-input" name="name" placeholder="Name..." required>
     <input type="email" class="data-input" name="email" placeholder="Email..." required>
     <input type="submit" value="Subscribe" class="button button-primary">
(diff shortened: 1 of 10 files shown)


src/20_Components/amp-form.html
```diff
@@ -75,7 +76,7 @@
     The `amp-form` component displays `submit-success` or `submit-error` elements based on the response and
     renders the template data inside these two elements.
   -->
-  <form method="post" action="<%host%>/components/amp-form/submit-form-input-text" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr"  target="_top">
+  <form method="post" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr"  target="_top">
```
sebastianbenz: action attribute is required. applies to other changes as well.

kul3r4: Hi, look at [validator](https://github.com/ampproject/amphtml/blob/553247b6a408a06e144e05eeb7dcfa96bfce0335/src/document-submit.js), it seems that if it's not a GET, action shouldn't be specified.

sebastianbenz: Hm - I was looking at the documentation. 

@mkhatib: POST + action - who is right? Documentation or validator?

sebastianbenz: @kul3r4 you're right. In that case we can get rid of target as well.

src/20_Components/amp-form.html
```diff
@@ -102,7 +103,7 @@
   }
   ```
 -->
-<form class="hide-inputs" method="post" action="<%host%>/components/amp-form/submit-form-input-text" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr" target="_top">
+<form class="hide-inputs" method="post" action-xhr="<%host%>/components/amp-form/submit-form-input-text-xhr" target="_top">
```
sebastianbenz: remove target (and for all the following samples)

src/20_Components/amp-form.html
```diff
@@ -346,27 +347,11 @@
...
-        Success! Thanks for trying the <code>amp-form</code> demo!
-      </template>
-    </div>
-    <div submit-error>
-      <template type="amp-mustache">
-        Error!
-      </template>
-    </div>
-  </form>
+
```
sebastianbenz: why did you remove this sample?

src/20_Components/amp-form.html
```diff
@@ -55,14 +55,15 @@
   <!-- #### Form Submission with Page Reload -->
   <!--
     Use the `action` attribute to specify a server-endpoint which should handle the form input. The URL must be HTTPS.
```
sebastianbenz: Add: "This works only for `GET` requests.

src/20_Components/amp-form.html
kul3r4: Because search doesn't make sense with POST. Line 64 contains search with GET.

kul3r4:
@sebastianbenz PTAL

src/20_Components/amp-form.html
sebastianbenz: my fault

Issue #394, opened 2016-10-03
Link to docs from example


cramforce:
E.g. this page https://ampbyexample.com/components/amp-video/ should link to the main docs.

adewale:
Good idea.

For amp-video should we link to: https://github.com/ampproject/amphtml/blob/master/builtins/amp-video.md or https://www.ampproject.org/docs/reference/components/amp-video

Once you've decided which URL to use we can do that in all component-specific examples.

sebastianbenz:
Migrated all existing links. We should make sure that each sample also links to the official docs.

kul3r4:
@sebastianbenz can we close this?

sebastianbenz:
yes

PR #423, opened 2016-10-12
Link to official docs instead of Github. #394

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

kul3r4:
Hi, I think some are missing:
- [A4A_Banner_Ad](https://github.com/ampproject/amp-by-example/blob/3f774161d3764f2c6fe4ea4cda97cd0f96e2afd4/src/60_AMP_for_Ads_(A4A)/A4A_Banner_Ad.html) has still (this)[https://github.com/ampproject/amphtml/blob/master/builtins/amp-pixel.md] reference  (line 42) and [doc](https://www.ampproject.org/docs/reference/components/amp-pixel) could be used
- [amp-install-serviceworker](https://github.com/ampproject/amp-by-example/blob/3f774161d3764f2c6fe4ea4cda97cd0f96e2afd4/src/20_Components/amp-install-serviceworker.html) at line 7 could use the [doc](https://www.ampproject.org/docs/reference/components/amp-install-serviceworker)
- [amp-font](https://github.com/ampproject/amp-by-example/blob/3f774161d3764f2c6fe4ea4cda97cd0f96e2afd4/src/20_Components/amp-font.html) at line 17 could use [doc](https://www.ampproject.org/docs/reference/components/amp-font)

sebastianbenz:
fixed - and some more references I've missed. PTAL

kul3r4:
LGTM

PR #427, opened 2016-10-15
Favorite sample using <amp-form> and SVG icon

dandv:
Feedback welcome.

The heart can only be toggled once for now. Will take another stab at it once [this PR](https://github.com/ampproject/amphtml/pull/5540) lands, so we can render another `<input>` in the `<div submit-success>` template.

### DIFF
--- a/src/30_Advanced/favorite.html
+++ b/src/30_Advanced/favorite.html
@@ -1,0 +1,117 @@
+<!---{
+    "experiments": ["amp-form", "form-submit"],
+    "preview": "default"
+  }--->
+<!--
+  #### Introduction
+
+  This example shows a clever use of [amp-form](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md)
+  for displaying a favorite ("heart") control, letting the user click/tap on it, and recording the event in server-side
+  code. For now, the "favorite" icon can only be tapped once. Toggling may be added [after AMP implements
+  support](https://github.com/ampproject/amphtml/pull/5540) for rendering `<input>` elements in
+  [amp-mustache](https://github.com/ampproject/amphtml/blob/master/extensions/amp-mustache/amp-mustache.md) templates.
+
+-->
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Nice hack!

I filed [this](https://github.com/ampproject/amphtml/issues/4272) a while ago to make this use case even easier to implement. The approach described there doesn't require rendering multiple form inputs.

src/30_Advanced/favorite.html
```diff
@@ -0,0 +1,117 @@
...
+    }
+  </style>
+</head>
+
+<body>
+  <!-- #### Submitting the form via XMLHttpRequest
+
+  Since we've loaded `<amp-form>`, we can use a regular `<form>` element. It only has one `input` control,
+  the submit button, and an `action-xhr` attribute set to the server endpoint. This prevents the form submission from
+  loading a new page. The endpoint result will be egnored for the purposes of this example, as well as [the
```
kul3r4: egnored -> ignored

src/30_Advanced/favorite.html
```diff
@@ -0,0 +1,117 @@
...
+  target attribute](https://github.com/ampproject/amphtml/issues/5607#issuecomment-253894602).
+
+  #### Replacing the "favorite" input field with a "favorite<b>d</b>" icon
+
+  The form's [on
+  attribute](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#events) is set to
+  call the `hide` built-in method of the `input` (identified by its id, `favorite-btn`). If the submission succeeds,
+
+  1. The [`<div submit-success>` element will be
+     rendered](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#successerror-response-rendering).
```
kul3r4: we are trying to link to the docs as much as possible instead of linking to the github project. In here [link](https://www.ampproject.org/docs/reference/components/amp-form) can be used, I don't think it supports the anchor though

sebastianbenz:
We should wait until #4272 is fixed before creating a sample for this. Will close this for now.

PR #428, opened 2016-10-15
Fix invalid link to next sample

### LINKED ISSUE #422: Fix invalid link to next sample
Example here: https://ampbyexample.com/amp_for_ads_a4a/a4a_carousel/ -> Next up: amp-form-input-text-error

uncompiled:
Fixes #422.

This PR adds a metadata field called hideExample that can be used to prevent specific docs from getting picked up in tasks/compile-example.js.  This prevents those documents from appearing the in "Next up" link.

### DIFF
--- a/tasks/compile-example.js
+++ b/tasks/compile-example.js
@@ -154,7 +154,7 @@ module.exports = function(config, updateTimestamp) {
         return null;
       }
       const metadata = next.document.metadata;
-      if (metadata && metadata.draft) {
+      if (!next.category() || (metadata && metadata.draft)) {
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Thanks for looking into this! 

The invalid samples appearing in "next up" is actually a bug. What's missing is a check in findNextExample: 

```
if (!next.category() || (metadata && metadata.draft)) {
   return findNextExample(examples, index);
}
```

Can you use this instead?

uncompiled:
No problem!  I somehow missed the `category()` method in `ExampleFile` and that is much cleaner fix.

Issue #430, opened 2016-10-17
Add pagination to live blog example


kul3r4:
I am currently working on adding pagination to the live blog sample.
This is gonna use the tag `<div pagination>` as explained in the [doc](https://github.com/ampproject/amphtml/blob/master/extensions/amp-live-list/amp-live-list.md).
At the moment we are showing at max 10 items, I am planning on limit to 5 for each page and then make the remaining 5 as part of a second page.
I am planning to add, at the bottom at the page, Prev and Next clickable URL.

PR #431, opened 2016-10-17
Add pagination

kul3r4:
Solves #430


backend/amp-live-list.go
```diff
@@ -157,13 +163,56 @@ func createMetadata(r *http.Request) []BlogPosting {
...
+func createLiveBlogSample(newStatus int, timestamp time.Time, r *http.Request, page Page) LiveBlogSample {
+
+	firstBlogID := r.URL.Query().Get("from")
 	if newStatus > len(blogs) {
 		newStatus = len(blogs)
 	}
 	blogItems := getBlogEntries(newStatus, timestamp)
 	score := createScore(newStatus, 0)
-	return LiveBlogSample{BlogItems: blogItems, FootballScore: score, BlogMetadata: createMetadata(r)}
+	firstItemIndex, _ := getProductIndexFromID(firstBlogID, blogItems)
```
sebastianbenz: don't ignore the error?

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -173,11 +187,19 @@
     </div>
     <!-- #### Blog Posts -->
     <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list) -->
```
sebastianbenz: Can you explain the pagination feature? Or do you plan to add this to the amp-live-list sample?

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -200,6 +222,21 @@ <h4 class="title">[[.Heading]]</h4>
           </div>
         [[end]]
     </div>
+
+    <div pagination>
+      <nav aria-label="amp live list pagination">
+      <ul class="pagination">
+        [[if eq .PrevPageURL ""]][[else]]
```
sebastianbenz:
```
[[if eq .PrevPageURL ""]][[else]]
 +          <li><a href="[[.PrevPageURL]]">Prev</a></li>
 +        [[end]]
```
can be replaced with:
```
[[if .PrevPageURL ]]
       <li><a href="[[.PrevPageURL]]">Prev</a></li>
 [[end]]
```

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -173,11 +187,19 @@
     </div>
     <!-- #### Blog Posts -->
     <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list) -->
+    [[if eq .PrevPageURL ""]]
```
sebastianbenz: use instead:
 
```
[[if .PrevPageURL]] 

[[else]]
...
```

backend/amp-live-list.go
```diff
@@ -157,13 +163,56 @@ func createMetadata(r *http.Request) []BlogPosting {
...
+		return buildSourceOrigin(r.Host) + page.Route + "?from=" + prevPageFirstItemID
+	}
+	return ""
+}
+
+func getPageNumberFromProductIndex(index int) int {
+	return (index / MAX_BLOG_NUMBER) + 1
+}
+
+func getProductIndexFromID(id string, blogItems []BlogItem) (int, error) {
```
sebastianbenz: Why not store the blogs in an array and use the index as ID?

backend/amp-live-list.go
```diff
@@ -157,13 +163,56 @@ func createMetadata(r *http.Request) []BlogPosting {
...
+		BlogMetadata:  createMetadata(r),
+		NextPageURL:   nextPageURL,
+		PrevPageURL:   prevPageURL,
+		PageNumber:    getPageNumberFromProductIndex(firstItemIndex)}
+}
+
+func getNextPageUrl(r *http.Request, page Page, blogItems []BlogItem, nextPageFirstItemIndex int) string {
+	if nextPageFirstItemIndex < len(blogItems) {
+		nextPageFirstItemID := blogs[nextPageFirstItemIndex].ID
+		return buildSourceOrigin(r.Host) + page.Route + "?from=" + nextPageFirstItemID
```
sebastianbenz: you build the same URL in both methods getNextPageUrl and getPrevPageUrl. Make this method only return the new index and build the URL somewhere else.

backend/amp-live-list.go
```diff
@@ -23,6 +25,7 @@ import (
 const (
 	AMP_LIVE_LIST_COOKIE_NAME = "ABE_AMP_LIVE_LIST_STATUS"
 	FIFTEEN_SECONDS           = 15
+	MAX_BLOG_NUMBER           = 5
```
sebastianbenz: What does MAX_BLOG_NUMBER mean? Can you find a better name?

backend/amp-live-list.go
kul3r4: Maybe I should remove the error, if the item cannot be found at the index provided, it should simply show the first page. What do you think?

sebastianbenz: sgtm

kul3r4:
@sebastianbenz PTAL

backend/amp-live-list.go
```diff
@@ -146,7 +154,7 @@ func createMetadata(r *http.Request) []BlogPosting {
 	for i := 0; i < len(blogs); i++ {
 		result = append(result, BlogPosting{"BlogPosting",
 			blogs[i].Heading,
-			buildSourceOrigin(r.Host) + "/" + CATEGORY_SAMPLE_TEMPLATES + "/live_blog/#" + blogs[i].ID,
+			buildSourceOrigin(r.Host) + "/" + CATEGORY_SAMPLE_TEMPLATES + "/live_blog/#" + fmt.Sprintf("%s%d", BLOG_ID_PREFIX, i+1),
```
sebastianbenz: I find it weird to combine concatenation via `+` and `fmt.Sprintf`. Please use only one.

backend/amp-live-list.go
```diff
@@ -157,13 +165,71 @@ func createMetadata(r *http.Request) []BlogPosting {
...
+		NextPageURL:   getNextPageUrl(firstItemIndex, r, page, blogItems),
+		PrevPageURL:   getPrevPageUrl(firstItemIndex, r, page, newStatus),
+		PageNumber:    getPageNumberFromProductIndex(firstItemIndex)}
+}
+
+func getNextPageUrl(firstItemIndex int, r *http.Request, page Page, blogItems []BlogItem) string {
+	nextPageFirstItemIndex := firstItemIndex + MAX_BLOG_ITEMS_NUMBER_PER_PAGE
+	nextPageId := getNextPageId(r, page, blogItems, nextPageFirstItemIndex)
+	if nextPageId != "" {
+		return buildPaginationURL(r, page, nextPageId)
```
sebastianbenz: That's not what I meant - this method shouldn't build the URL, build the URL before calling this method. This method should simply calculate the index.

backend/amp-live-list.go
```diff
@@ -157,13 +165,71 @@ func createMetadata(r *http.Request) []BlogPosting {
...
+
+func getPageNumberFromProductIndex(index int) int {
+	return (index / MAX_BLOG_ITEMS_NUMBER_PER_PAGE) + 1
+}
+
+func getBlogEntryIndexFromID(id string, blogItems []BlogItem) int {
+	blogEntryId, error := strconv.Atoi(id)
+	//default to the first page
+	if error != nil || blogEntryId > len(blogItems) {
+		return 0
```
sebastianbenz: why return 0 if they start from 1 according to the following comment?

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -200,6 +221,21 @@ <h4 class="title">[[.Heading]]</h4>
           </div>
         [[end]]
     </div>
+
+    <div pagination>
+      <nav aria-label="amp live list pagination">
+      <ul class="pagination">
+        [[if.PrevPageURL]]
```
sebastianbenz: nit: s/if.PrevPageURL/if .PrevPageURL/

backend/amp-live-list.go
sebastianbenz: Also, `buildSourceOrigin(r.Host) + "/" + CATEGORY_SAMPLE_TEMPLATES + "/live_blog/#` should be done outside the loop and not repeated every iteration.

backend/amp-live-list.go
```diff
@@ -157,13 +165,71 @@ func createMetadata(r *http.Request) []BlogPosting {
...
 
-func createLiveBlogSample(newStatus int, timestamp time.Time, r *http.Request) LiveBlogSample {
+func createLiveBlogSample(newStatus int, timestamp time.Time, r *http.Request, page Page) LiveBlogSample {
 	if newStatus > len(blogs) {
 		newStatus = len(blogs)
 	}
 	blogItems := getBlogEntries(newStatus, timestamp)
 	score := createScore(newStatus, 0)
-	return LiveBlogSample{BlogItems: blogItems, FootballScore: score, BlogMetadata: createMetadata(r)}
+	firstBlogID := strings.TrimPrefix(r.URL.Query().Get("from"), BLOG_ID_PREFIX)
```
sebastianbenz: perform this `strings.TrimPrefix(r.URL.Query().Get("from")` outside this method. Creating a blog entry should not depend on the request.

backend/amp-live-list.go
kul3r4: This is an helper method I added.
we have now:
getNextPageId -> which only builds the id
buildPaginationURL -> which builds the URL
getNextPageUrl -> which combines the 2 above
Should I remove the last one?

sebastianbenz: You build the URL twice instead of:

```
urlPrefix := buildURL(...)
nextIndex := nextId(...)
previousIndex := prevId(...)
nextUrl := urlPrefix + nextIndex
prevUrl := urlPrefix + prevIndex
```

backend/amp-live-list.go
kul3r4: Because I am returning the index, not the id.
So for item 1, the ID would be post1 and the index would be 0.

backend/amp-live-list.go
kul3r4: I see it now, thanks! I pushed some changes,
PTAL

sebastianbenz:
Can you fix the build?

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -172,12 +185,20 @@
       <p id="summary">This is a sample article demonstrating how to write a live blog in AMP. It demonstrates the usage of amp-live-list component which allows to create live blogs.</p>
     </div>
     <!-- #### Blog Posts -->
-    <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list) -->
+    <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list). Use `<div pagination>` tag to include pagination, find more [here](https://www.ampproject.org/docs/reference/components/amp-live-list) -->
```
sebastianbenz: sug: "Use `<div pagination>` tag to include pagination, find more [here](https://www.ampproject.org/docs/reference/components/amp-live-list)" -> "You can implement pagination for live-blogs using [`<div pagination>`](https://www.ampproject.org/docs/reference/components/amp-live-list)

spec/compiler/CodeSectionSpec.js
```diff
@@ -160,6 +160,12 @@ describe("CodeSection", function() {
       expect(section.hideColumns()).toEqual(true);
     });
   });
+  describe('removes uncorrectly escaped templates', function(){
```
sebastianbenz: Can you please cover the most common cases where we have to clean the code.

tasks/validate-example.js
```diff
@@ -40,7 +40,7 @@ module.exports = function() {
       // skip over experiments which will fail validation
       if (file.metadata &&
         (file.metadata.experiments || file.metadata.skipValidation) ||
-         !file.path.endsWith('.html')) {
+         !file.path.endsWith('.html') || file.path.includes('live_blog')) {
```
sebastianbenz: please don't hardcode samples - use metadata instead. If you look at the code you will see that there is already a flag for skipping sample validation...

kul3r4:
@sebastianbenz I added 2 more tests.
I also fixed the spaces for the json, have a look: [here](https://chiara-ampbyexample.appspot.com/samples_templates/live_blog/)
@jkingyens Would you mind taking a look at https://chiara-ampbyexample.appspot.com/samples_templates/live_blog/ from the point of view of metadata?

spec/compiler/CodeSectionSpec.js
```diff
@@ -160,6 +160,18 @@ describe("CodeSection", function() {
...
+    it('contains a template', function() {
+      expect(section.cleanUpCode("[[<span class=\"hljs-attr\">.Disabled</span>]]")).toEqual("[[.Disabled]]")
+    });
+    it('does not alter valid escaped templates', function() {
+      expect(section.cleanUpCode("<span class=\"hljs-string\">\"[[.Timestamp]]\"</span>")).toEqual("<span class=\"hljs-string\">\"[[.Timestamp]]\"</span>")
+    });
+    it('does not alter valid escaped templates which contains if clause', function() {
+      expect(section.cleanUpCode("</span>[[if .PrevPageURL]]<span class=\"hljs-tag\">")).toEqual("</span>[[if .PrevPageURL]]<span class=\"hljs-tag\">")
```
sebastianbenz: Not sure this test makes sense. Much more interesting would be whether  escaped ifs, range & co work.

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -172,12 +186,14 @@
       <p id="summary">This is a sample article demonstrating how to write a live blog in AMP. It demonstrates the usage of amp-live-list component which allows to create live blogs.</p>
     </div>
     <!-- #### Blog Posts -->
-    <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list) -->
-    <amp-live-list
+    <!-- Use amp-live-list to implement a live blog. The amp-live-list component regularly polls the host document for updated content and updates the end user's browser as new items become available. This means that every time a new post needs to be added, the host document should be updated by the CMS to include the update both the body and the metadata section. Find an amp-live-list example [here](/components/amp-live-list).
```
sebastianbenz: Can you please add a line break after this paragraph.

src/50_Samples_%26_Templates/Live_Blog.html
```diff
@@ -172,12 +186,14 @@
       <p id="summary">This is a sample article demonstrating how to write a live blog in AMP. It demonstrates the usage of amp-live-list component which allows to create live blogs.</p>
     </div>
     <!-- #### Blog Posts -->
-    <!-- Use amp-live-list to implement a live blog. Find an amp-live-list example [here](/components/amp-live-list) -->
-    <amp-live-list
+    <!-- Use amp-live-list to implement a live blog. The amp-live-list component regularly polls the host document for updated content and updates the end user's browser as new items become available. This means that every time a new post needs to be added, the host document should be updated by the CMS to include the update both the body and the metadata section. Find an amp-live-list example [here](/components/amp-live-list). 
+    Long blogs could use pagination to improve performance by limiting the number of blog items to be displayed on a page. To implement pagination, amp-live-list allows to add the `<div pagination></div>` element, to then insert any markup needed for pagination, for example the page number, a link to the next and previous page. -->
```
sebastianbenz: Perfect! Thanks!

kul3r4:
@sebastianbenz PTAL

tasks/lib/CodeSection.js
```diff
@@ -133,5 +132,10 @@ module.exports = class CodeSection {
     }
     return startIndex;
   }
+
+  cleanUpCode(input) {
+    const encodedTemplateRegexp = /\[\[\s*\n*<span\sclass="hljs\-attr">(range|if|else|end)?(<\/span><span\sclass="hljs\-attr">)?(\.[A-Za-z]*)?<\/span>\s*\n*\]\]/g
```
sebastianbenz: Can you move the regex declaration to a constant at the top of the file. 

* `/s*/n*` -> the /n will never match as /s matches all whitespaces
* can you simplify `<span\sclass="hljs\-attr">` to `<.*?>` ? (I'm not sure)
* can you simplify `(range|if|else|end)?` to `(.)*?` ? (I'm not sure)

spec/compiler/CodeSectionSpec.js
```diff
@@ -160,6 +160,24 @@ describe("CodeSection", function() {
...
+      expect(section.cleanUpCode("<span class=\"hljs-string\">\"[[.Timestamp]]\"</span>")).toEqual("<span class=\"hljs-string\">\"[[.Timestamp]]\"</span>")
+    });
+    it('contains an escaped template with range clause', function() {
+      expect(section.cleanUpCode("[[<span class=\"hljs-attr\">range</span><span class=\"hljs-attr\">.BlogItems</span>]]")).toEqual("[[range .BlogItems]]")   
+    });
+    it('contains an escaped template with if clause', function() {
+      expect(section.cleanUpCode("[[<span class=\"hljs-attr\">if</span><span class=\"hljs-attr\">.BlogItems</span>]]")).toEqual("[[if .BlogItems]]")   
+    });
+    it('contains an escaped template with end clause', function() {
+      expect(section.cleanUpCode("[[<span class=\"hljs-attr\">end</span>]]")).toEqual("[[end ]]")
```
sebastianbenz: What happens if templates contain whitespace? 

```
[[  if   .Test   ]]
[[  .Test   ]]
...
```

kul3r4:
@sebastianbenz PTAL

src/20_Components/amp-live-list.html
```diff
@@ -75,7 +75,7 @@
...
 
-    This sample shows an implementation with a server rendering a series of 10 elements. `amp-live-list` will poll the page every 15 seconds. Every time `amp-live-list` polls the page, the server adds a new item to the page. Refer to [server.go](https://github.com/ampproject/amp-by-example/blob/master/server.go) and [amp-live-list.go](https://github.com/ampproject/amp-by-example/blob/master/backend/amp-live-list.go) for implementation details.
+    This sample shows an implementation with a server rendering a series of 5 elements. `amp-live-list` will poll the page every 15 seconds. Every time `amp-live-list` polls the page, the server adds a new item to the page. Refer to [server.go](https://github.com/ampproject/amp-by-example/blob/master/server.go) and [amp-live-list.go](https://github.com/ampproject/amp-by-example/blob/master/backend/amp-live-list.go) for implementation details.
 
     This sample stops after ten updates. If you want to start over, delete cookies from ampbyexample.com or open this page again in an incognito/private tab.
```
sebastianbenz: Is this still correct? Doesn't it stop after 5 items?

PR #420, opened 2016-10-12
Make code sections expandable

sebastianbenz:
Use a generic algorithm to collapse code sections. Collapse if:

- code section has no comment
- code section has > 4 lines

PR #435, opened 2016-10-19
Ensure the <amp-list> overflow demo works

dandv:
The last example in `<amp-list>`, exercising the overflow feature, worked intermittently. Sometimes the display looked like this, due to the fixed height begin set to 10:

image

Other times, the list was already fully expanded, because there was sufficient vertical space thanks to the code section staying expanded (NB: according to [the heuristic](https://github.com/ampproject/amp-by-example/pull/420#issue-182528225), this section should be collapsed because it has > 4 lines and no comment. However, it is fully expanded on reload (also not showing the overflow button):

image

This fix instructs the user to go to the preview mode in case they don't see the overflow button:

image

### DIFF
--- a/src/20_Components/amp-list.html
+++ b/src/20_Components/amp-list.html
@@ -1,3 +1,7 @@
+<!---{
+  "default": "preview",
+  "preview": "default"
+  }--->
 <!--
   #### Introduction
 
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Good point. A problem is the preview link which doesn't work in [embed mode](https://ampbyexample.com/components/amp-list/embed/). Unfortunately, there is no workaround yet.

sebastianbenz:
Fixed by https://github.com/ampproject/amp-by-example/pull/708

Issue #407, opened 2016-10-09
Update images 


kul3r4:
This is a placeholder for a task of replacing images from samples:
- amp-live-list
- news-article
- live-blog

PR #438, opened 2016-10-20
Uses different images

kul3r4:
Solves #407 
@sebastianbenz  PTAL


kul3r4:
Images are under CC0, they don't require attribution.
I updated the other samples, yes the images have different resolutions, but the carousel still works.

sebastianbenz:
Some of the images have way too much whitespace in the carousel due to different aspect ratios. Can you please only use images with roughly the same aspect ratio in the NewsArticle carousel. Thanks!

screen shot 2016-10-21 at 08 19 37

kul3r4:
@sebastianbenz I picked other 2 images with more similar height.

sebastianbenz:
LGTM - thanks for doing this!

kul3r4:
@ericlindley-g FYI

ericlindley-g:
Looks great—thank you!

PR #443, opened 2016-10-21
Don't use list tags in amp-list. Fixes #441

### LINKED ISSUE #441: amp-list markup needs improvment.
https://ampbyexample.com/components/amp-list/ has invalid HTML setup and there is no need for the wrapper UL and LI children.

They should just be:
```
    <amp-list width=auto
        height=100
        layout=fixed-height
        src="https://ampbyexample.com/json/examples.json">
      <template type="amp-mustache"
          id="amp-template-id">
        <div>
          <a href={{url}}>{{title}}</a>
        </div>
      </template>
    </amp-list>
```
We should add some text to the example that ensures the author we are using proper `aria` roles to make these divs list items

If we really like to allow folks to use 'li' as children, then amp-list needs to become smarter and not create the container until it has a child and can inspect the child to see if it is a li or not and based on that either create a div or an ul

See https://github.com/ampproject/amphtml/issues/5513#issuecomment-255264801

@lannka @kul3r4

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

kul3r4:
LGTM

PR #449, opened 2016-10-25
Uses [[ for template, removes / for templates route

kul3r4:
Shopping cart for the product listing is broken at the moment, this is a fix.


kul3r4:
@sebastianbenz PTAL

PR #453, opened 2016-10-27
Adds the target to form

kul3r4:
Unfortunately I missed a target for a form sample. This PR add the target.


kul3r4:
@sebastianbenz PTAL

PR #455, opened 2016-10-28
Removes experiment from amp form

kul3r4:


kul3r4:
@sebastianbenz PTAL

sebastianbenz:
How about the other samples using amp-form?

kul3r4:
You are right! done

sebastianbenz:
What about housing & product_listing?

kul3r4:
Thanks, removed from there as well

kul3r4:
@sebastianbenz In the housing sample, I was using the attribute type for the element li which is obsolete, I added css instead.

PR #459, opened 2016-10-31
Optimize script order. 

### LINKED ISSUE #458: Optimize script order
The AMP JavaScript should ideally be referenced high up the HTML near the opening `<head>`.

Also `v0.js` should be the first script tag to be referenced.

sebastianbenz:
Fixes #458


sebastianbenz:
@kul3r4 PTAL

PR #473, opened 2016-11-08
Updates amp-sticky-ad library

### LINKED ISSUE #472: Update amp-sticky-ad example to use v1.0
amp-sticky-ad has upgraded to version 1.0. Would you please update the example? Thanks.
CC @kul3r4 @zhouyx

kul3r4:
This PR is in preparation of amp-sticky-ad library update, this update will go in the AMP next release. Please wait before merging as I would like to test once the update is live.
Fixes #472

### DIFF
--- a/src/20_Components/amp-sticky-ad.html
+++ b/src/20_Components/amp-sticky-ad.html
@@ -15,23 +15,10 @@
   <!--
     Import the `amp-sticky-ad` component in the header.
   -->
-  <script async custom-element="amp-sticky-ad" src="https://cdn.ampproject.org/v0/amp-sticky-ad-0.1.js"></script>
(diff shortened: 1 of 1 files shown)


kul3r4:
@jasti @zhouyx I was trying the sample with the new library, but I don't see the css class "amp-sticky-ad-loaded" anymore. This class should still be available according to https://github.com/ampproject/amphtml/blob/master/extensions/amp-sticky-ad/amp-sticky-ad.md.

zhouyx:
@kul3r4 Yes that css class is removed from our code. The doc is for v0.1 version. I will update the doc and add a new doc for v1.0.

zhouyx:
You can still set style through `amp-sticky-ad`

kul3r4:
@sebastianbenz PTAL

PR #494, opened 2016-11-26
Fixed Social Share example (data-param)

peterplus1:
I also cleaned it up on several places, there is no need for a width=60 if the default is 60px for example (maybe at the time this example was created there was no default?). Displaying less attributes makes the examples easier to read, especially if they aren't needed for a particular example.

The Basic Usage is a lot simpler this way, ready to go.

The Configuration paragraph now clearly tells the viewer that "text" is the same as the attribute, and I've fixed the data-attributes, because they're (now) supposed to be "data-param-*".

I've started cleaning up the "Customized Views", just stripping away the configuration that isn't needed in this example for now (the example looked quite scary with about 10-15 tags and attributes.). The example bugs out on my screen, so I'm thinking this example is deprecated.

### DIFF
--- a/src/20_Components/amp-social-share.html
+++ b/src/20_Components/amp-social-share.html
@@ -35,7 +35,7 @@
 
   <!-- #### Basic Usage -->
   <!--
-    Embed the [`amp-social-share`](https://www.ampproject.org/docs/reference/components/amp-social-share) widget choosing a share type from the [supported types](https://www.ampproject.org/docs/reference/components/amp-social-share#types), ensuring you use the relevant width and height.
+    Embed the [`amp-social-share`](https://www.ampproject.org/docs/reference/components/amp-social-share) widget choosing a share type from the [supported types](https://www.ampproject.org/docs/reference/components/amp-social-share#types).
 
     `amp-social-share` with `type="facebook"` requires to specify the [facebook app id](https://developers.facebook.com/apps) via `data-attribution`.
   -->
@@ -55,8 +55,9 @@
     * `width`, default 60px.
     * `height`, default 44px.
     * `data-param-text` is the text to include in the share.
-    * `data-param-url` is the URL to share.
+    * `data-param-url` is the URL to share, current URL by default.
     * `data-param-attribution` is where the share is attributed to.
+
 
     All `data-param-*` prefixed attributes will be turned into URL parameters and passed to the share endpoint.
   -->
@@ -70,7 +71,7 @@
 
   <!-- #### Custom Style -->
   <!--
-    Whenever you want to provide your own style, use CSS properties within an `<style amp-custom>` ([Modify the presentation](https://www.ampproject.org/docs/get_started/create/presentation_layout)) element. This will make sure the image stays responsive and centered, and the default styling is overwritten.
+    Whenever you want to provide your own style, use CSS properties within the `<style amp-custom>` element ([Modify the presentation](https://www.ampproject.org/docs/get_started/create/presentation_layout)). This will make sure the image stays responsive and centered, and the default styling is overwritten.
 
     * `background-color`, if you want the element to be a different color.
     * `background-image`, provide another image if you want to change the icon.
@@ -81,11 +82,6 @@
   <amp-social-share type="linkedin"
     width="76"
     height="22"
-    class="custom-style">
-  </amp-social-share>
-  <amp-social-share type="linkedin"
-    width="41"
-    height="64"
     class="custom-style">
   </amp-social-share>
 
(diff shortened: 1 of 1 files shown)


plvdmeer:
I signed the CLA! I'm guessing this is where I reply to @googlebot? :')

src/20_Components/amp-social-share.html
```diff
@@ -32,37 +40,53 @@
...
 
-    * `text` is the text to include in the share
-    * `url` is the URL to share
-    * `attribution` is where the share is attributed to.
+    * `width`, default 60px.
+    * `height`, default 44px.
+    * `data-param-text` is the text to include in the share.
+    * `data-param-url` is the URL to share.
+    * `data-param-attribution` is where the share is attributed to.
+
```
sebastianbenz: Can you please add another empty line here. Otherwise the markdown processor messes up the next paragraph.

src/20_Components/amp-social-share.html
```diff
@@ -32,37 +40,53 @@
...
+    data-param-text="Check out these AMP Examples!"
+    data-param-url="https://ampbyexample.com/"
+    data-param-attribution="AMPhtml">
   </amp-social-share>
 
-  <!-- #### Customized Views -->
+  <!-- #### Custom Style -->
   <!--
-    Sometimes you want to provide your own style. In this instance, you can **embed an anchor without a `href`** and it will be populated. This provides you the flexibility to build your own UI for the share button.
+    Whenever you want to provide your own style, use CSS properties within an `<style amp-custom>` ([Modify the presentation](https://www.ampproject.org/docs/get_started/create/presentation_layout)) element. This will make sure the image stays responsive and centered, and the default styling is overwritten.
```
sebastianbenz: nit: can you move the link to the end of the sentence. Makes it easier to read:

```
the `<style amp-custom>` element ([Modify the presentation](https://www.ampproject.org/docs/get_started/create/presentation_layout)).
```

src/20_Components/amp-social-share.html
```diff
@@ -32,37 +40,53 @@
...
-    <a id="customized-social-share" class="custom">
-      <amp-img src="https://raw.githubusercontent.com/google/material-design-icons/master/social/1x_web/ic_share_white_48dp.png" width="18" height="18" layout="responsive" alt="an image">
-      </amp-img>
-    </a>
+  <amp-social-share type="linkedin"
+    width="76"
+    height="22"
+    class="custom-style">
+  </amp-social-share>
+  <amp-social-share type="linkedin"
```
sebastianbenz: Is there a particular reason why you've added two custom samples?

src/20_Components/amp-social-share.html
plvdmeer: Agreed, done.

src/20_Components/amp-social-share.html
plvdmeer: Not really, wasn't sure I was gonna commit the two examples. I was thinking it would make it more clear to the viewer that it's responsive and works the same as the "original" styling, but I don't think it's our job to teach people common CSS. Removed the second example, it's implied that it works the same as the out-of-the-box styling.

plvdmeer:
@sebastianbenz PTAL :)

src/20_Components/amp-social-share.html
```diff
@@ -21,48 +21,68 @@
...
+    * `background-size` (when using `background-image`), set to `contain`.
   -->
-  <amp-social-share type="linkedin" width="60" height="44" data-text="The AMP Project" data-url="https://www.ampproject.org/" data-attribution="amphtml">
-    <a id="customized-social-share" class="custom">
-      <amp-img src="https://raw.githubusercontent.com/google/material-design-icons/master/social/1x_web/ic_share_white_48dp.png" width="18" height="18" layout="responsive" alt="an image">
-      </amp-img>
-    </a>
+  <amp-social-share type="linkedin"
+    width="76"
+    height="22"
```
sebastianbenz: One small nit: can you resize the sample to 40x40 as the icon is square.

peterplus1: No problem, but maybe I should remove the height+width completely? The point of that example is to show the custom styling, which doesn't interfere with the default behavior. I used those dimensions to show that the image scales and stays within the container, default is a square icon within a rectangle shape as well. I agree the buttons look better when they are square, but I don't think that should be the point of that example. Making smaller examples that people can quickly understand and pick and choose to their desired result should be the focus. So I'd vote for just the type and class attributes for this example.

sebastianbenz: +1 even better

peterplus1: Awesome! Committed the change.

PR #497, opened 2016-11-30
Fix runtime errors in amp-banner-sample

sebastianbenz:
Don't install serviceworker as canonical domain does not match
serviceworker domain.


sebastianbenz:
@kul3r4 PTAL

kul3r4:
LGTM

PR #507, opened 2016-12-03
Improve ServiceWorker

sebastianbenz:
* instantly claim current page
* return error if offline page is missing
* cache first page


sebastianbenz:
@ithinkihaveacat I've been shamelessly copying from Gulliver's sw.js. PTAL

sebastianbenz:
Thanks!

Issue #372, opened 2016-09-19
Poll sample using amp-form


rudygalfi:
- a single question with 4 options or so
- our server could store responses anonymously, so the results displayed would just show the aggregate totals to-date
- would be awesome to have an option to just see the results without casting a vote

mkhatib:
> a single question with 4 options or so

Checkout the `examples/form.amp.html` for a sample poll I created to get started on this.

kul3r4:
@rudygalfi we deployed the [sample](https://ampbyexample.com/samples_templates/poll/preview/).

rudygalfi:
Awesome!

rudygalfi:
cc @ericlindley-g

mkhatib:
@kul3r4 this is pretty cool! Thanks for putting it out there! If you guys have a simple datastore on the ampbyexample server it would be cool to actually collect and store results 😄 - I played around the example to see if my voting is changing the result.

Can we get more examples like these, specifically an example that uses `amp-slidescroll.goToSlide` action to advance to the next question when the current one is answered - so a multi-step quiz/poll. Something like the following:

```html
<amp-carousel id="quiz" type="slides" >
<form id=q1 ... on="submit:quiz.goToSlide(index=1)">
   ... fields for the first question with on=change:q1.submit ...
</form>
<form id=q2 ... on="submit:quiz.goToSlide(index=2)">
   ... fields for the second question with on=change:q2.submit ...
</form>
<form id=q3 ...on="submit:quiz.goToSlide(index=3)">
   ... fields for the third question with on=change:q3.submit ...

  ... templates to render results for the entire quiz ...
</form>
</amp-carousel>
```

I have not tried this but it should work in theory.

rudygalfi:
We should discuss quiz in #371. :)

kul3r4:
@mkhatib the results are already stored, this is how you are able to see the percentage of answers increasing.

kul3r4:
Close via #510

PR #510, opened 2016-12-05
Adds a poll

kul3r4:
Solves #372
This is based on [amp sample](https://github.com/ampproject/amphtml/blob/master/examples/forms.amp.html) (with form id "poll1").
At the moment the server is returning fixed values for showing the percentage of answers, my plan was to persist the answer together with the client id in order to have a mapping on who has answered or not.
Find preview [here](https://chiara-ampbyexample.appspot.com/samples_templates/poll/) 

@sebastianbenz What do you think?

### DIFF
--- a/backend/housing.go
+++ b/backend/housing.go
@@ -55,13 +55,8 @@ func parseForm(r *http.Request) (MortgageForm, error) {
 
 	// can't return nil in place of a struct, so creating one in case there are errors
 	mortgageForm := MortgageForm{price, deposit, interest, period}
-
 	err := parseFormErrors([]error{priceErr, depositErr, interestErr, periodErr})
-	if err != nil {
-		return mortgageForm, err
-	}
-
-	return mortgageForm, nil
+	return mortgageForm, err
 }
 
 func calculateMortgageXHR(w http.ResponseWriter, r *http.Request) {
--- a/backend/housing.go
+++ b/backend/housing.go
@@ -19,7 +19,6 @@ import (
 	"math"
 	"net/http"
 	"strconv"
-	"strings"
 )
 
 const (
@@ -57,31 +56,12 @@ func parseForm(r *http.Request) (MortgageForm, error) {
 	// can't return nil in place of a struct, so creating one in case there are errors
 	mortgageForm := MortgageForm{price, deposit, interest, period}
 
-	err := parseFormErrors(priceErr, depositErr, interestErr, periodErr)
+	err := parseFormErrors([]error{priceErr, depositErr, interestErr, periodErr})
 	if err != nil {
 		return mortgageForm, err
 	}
 
 	return mortgageForm, nil
-}
-
-func parseFormErrors(priceErr error, depositErr error, interestErr error, periodErr error) error {
-	var errors []string
-	appendError(priceErr, errors)
-	appendError(depositErr, errors)
-	appendError(interestErr, errors)
-	appendError(periodErr, errors)
-
-	if errors != nil {
-		return fmt.Errorf(strings.Join(errors, "\n"))
-	}
-	return nil
-}
-
-func appendError(anError error, errors []string) {
-	if anError != nil {
-		errors = append(errors, anError.Error())
-	}
 }
 
 func calculateMortgageXHR(w http.ResponseWriter, r *http.Request) {
--- a/backend/poll.go
+++ b/backend/poll.go
@@ -1,0 +1,159 @@
+// Copyright Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend
+
+import (
+	"appengine"
+	"appengine/datastore"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+)
+
+const (
+	ERROR_CASE_POLL  = "error"
+	POLL_SAMPLE_PATH = "/" + CATEGORY_SAMPLE_TEMPLATES + "/poll/"
+)
+
+type PollForm struct {
+	ClientId string
+	Answer   string
+}
+
+type Result struct {
+	Percentage []int
+	Answer     string
+}
+
+type Answer struct {
+	AnswerText string
+	Votes      int
+}
+
+func InitPollSample() {
+	http.HandleFunc(POLL_SAMPLE_PATH+"submit-xhr", submitXHR)
+}
+
+func getAnswersFromDatastore(context appengine.Context) ([]Answer, error) {
+	query := datastore.NewQuery("Answer")
+	answers := make([]Answer, 0, 4)
+	if _, err := query.GetAll(context, &answers); err != nil {
+		return answers, err
+	}
+	return answers, nil
+}
+
+func saveAnswerIntoDatastore(context appengine.Context, answerText string, voteCount int) error {
+	answer := &Answer{
+		AnswerText: answerText,
+		Votes:      voteCount,
+	}
+	answerKey := datastore.NewKey(context, "Answer", answerText, 0, nil)
+	_, error := datastore.Put(context, answerKey, answer)
+	if error != nil {
+		return error
+	}
+	return nil
+}
+
+func updateAnswerCount(answers []Answer, answerText string) int {
+	voteCount := 1
+	found := false
+	for i := 0; i < len(answers); i++ {
+		if answers[i].AnswerText == answerText {
+			voteCount = answers[i].Votes + 1
+			answers[i].Votes = voteCount
+			found = true
+			break
+		}
+	}
+	if !found {
+		answers = append(answers, Answer{
+			AnswerText: answerText,
+			Votes:      voteCount,
+		})
+	}
+	return voteCount
+}
+
+func calculateResult(w http.ResponseWriter, context appengine.Context, pollForm PollForm) ([]Result, error) {
+	var results []Result
+	answerText := pollForm.Answer
+	answers, error := getAnswersFromDatastore(context)
+	if error != nil {
+		return results, error
+	}
+	voteCount := updateAnswerCount(answers, answerText)
+	error = saveAnswerIntoDatastore(context, answerText, voteCount)
+	if error != nil {
+		return results, error
+	}
+
+	for _, value := range answers {
+		results = append(results, Result{Answer: value.AnswerText, Percentage: make([]int, value.Votes)})
+	}
+
+	return results, nil
+}
+
+func parsePollForm(w http.ResponseWriter, r *http.Request, context appengine.Context) (PollForm, error) {
+	answerText, answerErr := parseNotEmptyFormValue(r, "answer")
+	clientId, clientIdErr := parseNotEmptyFormValue(r, "clientId")
+
+	pollForm := PollForm{clientId, answerText}
+
+	error := parseFormErrors([]error{answerErr, clientIdErr})
+	if error != nil {
+		return pollForm, error
+	}
+	return pollForm, nil
+}
+
+func parseNotEmptyFormValue(r *http.Request, input string) (string, error) {
+	value := r.FormValue(input)
+	if value == "" {
+		return "", errors.New(input + "has empty value")
+	}
+	return value, nil
+}
+
+func submitXHR(w http.ResponseWriter, r *http.Request) {
+	EnableCors(w, r)
+	SetContentTypeJson(w)
+	response := ""
+	context := appengine.NewContext(r)
+	pollForm, error := parsePollForm(w, r, context)
+	if error != nil {
+		handleError(error, w)
+	}
+	result, error := calculateResult(w, context, pollForm)
+	if error != nil {
+		handleError(error, w)
+	}
+	pollResult, _ := json.Marshal(result)
+	response = fmt.Sprintf("{\"result\":%s}", pollResult)
+	w.Write([]byte(response))
+}
+
+func handleError(error error, w http.ResponseWriter) {
+	w.WriteHeader(http.StatusBadRequest)
+	response := fmt.Sprintf("{\"err\":\"%s\"}", error)
+	w.Write([]byte(response))
+}
+
+func submit(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(http.StatusBadRequest)
+}
--- a/backend/poll.go
+++ b/backend/poll.go
@@ -22,6 +22,7 @@ import (
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
 	"net/http"
+	"strconv"
 )
 
 const (
@@ -33,7 +34,7 @@ const (
 //holds the answer chosen by the user and the client id, used for storing data coming from the UI
 type PollForm struct {
 	ClientId string
-	Answer   string
+	Answer   int
 }
 
 //holds an answers and votes represented as an array, used for displaying
@@ -44,41 +45,39 @@ type PollEntryResult struct {
 
 //holds the poll results and a message, used for displaying
 type PollResult struct {
-	PollResults []PollEntryResult
-	Message     string
+	PollEntryResults []PollEntryResult
+	Message          string
 }
 
 //holds an answer and votes, used for storing
 type PollAnswer struct {
-	Answer string
+	Answer int
 	Votes  int
 }
 
 //holds the questions
 type PollQuestions struct {
-	Questions  map[string]string
+	Questions []string
 }
 
-//holds stored data
+//holds stored Poll
 type Poll struct {
 	ClientIds   []string
 	PollAnswers []PollAnswer
 }
 
-var questions map[string]string
+var questions []string
+var pollQuestions PollQuestions
 
 func InitPollSample() {
-	questions = make(map[string]string)
-	questions["question1"] = "Penguins"
-	questions["question2"] = "Ostriches"
-	questions["question3"] = "Kiwis"
-	questions["question4"] = "Wekas"
+	questions = []string{"Penguins", "Ostriches", "Kiwis", "Wekas"}
+	pollQuestions = PollQuestions{questions}
 	http.HandleFunc(POLL_SAMPLE_PATH+"submit", submitPoll)
 	RegisterSample(CATEGORY_SAMPLE_TEMPLATES+"/poll", handlePoll)
 }
 
 func handlePoll(w http.ResponseWriter, r *http.Request, page Page) {
-	page.Render(w, PollQuestions{questions})
+	page.Render(w, pollQuestions)
 }
 
 func createPollResult(answers []PollAnswer, message string) PollResult {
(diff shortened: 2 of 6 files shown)


src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,138 @@
+<!---{
+  "preview": "default"
+  }--->
+  <!--
+  #### Introduction
+
+  This is a sample template for a poll in AMP. After selecting one of the answers, you will see the percentage of chosen answers.
```
sebastianbenz: Can you provide a bit more overview and context? What are the challenges you're sovling.

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,138 @@
...
+-->
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+<meta charset="utf-8">
+<script async src="https://cdn.ampproject.org/v0.js"></script>
+<!-- #### Setup -->
+<!--
+Import the additional AMP components. This sample uses a form and mustache templates.
```
sebastianbenz: s/a form/amp-form/

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+</style>
+</head>
+<body>
+
+<h4 id="header1">on=change and form.submit examples</h4>
+<!-- #### Poll -->
+<!--
+Use `amp-form` to realize a poll. Use `clientId` variable to identify the user that has voted, more info [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md#client-id).
```
sebastianbenz: s/Use/We use/ (there might be other options)

backend/poll.go
```diff
@@ -0,0 +1,97 @@
...
+	CanonicalUrl  string
+	Answer string
+}
+
+type Result struct {
+	Percentage    []int
+	Answer string
+}
+
+func InitPollSample() {
```
sebastianbenz: Why do we need both endpoints?

backend/poll.go
```diff
@@ -0,0 +1,97 @@
...
+	http.HandleFunc(POLL_SAMPLE_PATH+"submit-xhr", submitXHR)
+	http.HandleFunc(POLL_SAMPLE_PATH+"submit", submit)
+}
+
+func calculateResult(pollForm PollForm) []Result {
+
```
sebastianbenz: delete empty line

kul3r4:
@sebastianbenz PTAL
About your previous question:
how would you implement a poll that spans multiple pages? ->Do you have any idea?The current system of showing one line will not work.
One alternative would be just showing the number next to the answer.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	Answer   string
+}
+
+type Result struct {
+	Percentage []int
+	Answer     string
+}
+
+type Answer struct {
+	AnswerText string
```
sebastianbenz: Can you maybe find better names?

* what does PollForm represent?
* what is Result?
* inconsistent string naming AnswerText vs Answer

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	Votes      int
+}
+
+func InitPollSample() {
+	http.HandleFunc(POLL_SAMPLE_PATH+"submit-xhr", submitXHR)
```
sebastianbenz: I wouldn't use the `-xhr` postfix as all requests are xhr.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func getAnswersFromDatastore(context context.Context) ([]Answer, error) {
+	query := datastore.NewQuery("Answer")
```
sebastianbenz: Extract `"Answer"` into a constant - you use the same string at least three times.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	answers := make([]Answer, 0, 4)
```
sebastianbenz: What is 4?

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	if _, err := query.GetAll(context, &answers); err != nil {
+		return answers, err
```
sebastianbenz: s/answer/nil/ to make it more explicit that this is the error case.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	answer := &Answer{
+		AnswerText: answerText,
+		Votes:      voteCount,
+	}
+	answerKey := datastore.NewKey(context, "Answer", answerText, 0, nil)
+	_, error := datastore.Put(context, answerKey, answer)
+	if error != nil {
+		return error
+	}
+	return nil
```
sebastianbenz: You don't need the if, you can simply return error.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+
+func parsePollForm(w http.ResponseWriter, r *http.Request, context context.Context) (PollForm, error) {
+	answerText, answerErr := parseNotEmptyFormValue(r, "answer")
+	clientId, clientIdErr := parseNotEmptyFormValue(r, "clientId")
+
+	pollForm := PollForm{clientId, answerText}
+
+	error := parseFormErrors([]error{answerErr, clientIdErr})
+	if error != nil {
+		return pollForm, error
```
sebastianbenz: s/pollForm/nil/ (same as above)

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,140 @@
...
+
+</style>
+</head>
+<body>
+
+<h4 id="header1">Flightless bird poll</h4>
+<!-- #### Poll -->
+<!--
+We use `amp-form` to realize a poll.
+We use the `CLIENT_ID` variable to identify the user, this enables storing answers between repeated visits of an AMP page. This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
```
sebastianbenz: Why do you need to store answers between repeated visits? Isn't this simply about being able to tell if someone already submitted an answer?

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	return voteCount
+}
+
+func calculatePollResults(w http.ResponseWriter, ctx context.Context, pollForm PollForm) ([]Result, error) {
+	var results []Result
+	err := datastore.RunInTransaction(ctx, func(ctx context.Context) error {
+		//create keys
+		clientIdKey := datastore.NewKey(ctx, "PollForm", pollForm.ClientId, 0, nil)
+		answerKey := datastore.NewKey(ctx, "Answer", pollForm.Answer, 0, nil)
```
sebastianbenz: I think this is what IncompleteKeys are for - if you just want to query for an object key. Also replace for other queries.

backend/poll.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if error != nil {
+		handleError(error, w)
+	}
+	result, error := calculatePollResults(w, context, pollForm)
+	if error != nil {
+		handleError(error, w)
+	}
+	pollResult, _ := json.Marshal(result)
+	response = fmt.Sprintf("{\"result\":%s}", pollResult)
```
sebastianbenz: An easier way to write json to the response:

```
json.NewEncoder(w).Encode(result)
```

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,140 @@
...
+            left: 0;
+            right: 0;
+            bottom: 0;
+        }
+        .one-pc {
+            width: 1%;
+            height: 100%;
+            display: inline-block;
+        }
+        .Wekas .one-pc {
```
sebastianbenz: Having separate colors for answers looks good, but bloats our css. I'd be fine with using a single color.

src/50_Samples_%26_Templates/Poll.html
sebastianbenz: You start two subsequent sentences with 'We use'.

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,140 @@
...
+We use `amp-form` to realize a poll.
+We use the `CLIENT_ID` variable to identify the user, this enables storing answers between repeated visits of an AMP page. This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
+
+<form method="post" action-xhr="<%host%>/samples_templates/poll/submit-xhr" target="_blank"
+      id="poll1"
+      custom-validation-reporting="as-you-go">
+    <input name="clientId" type="hidden" default-value="CLIENT_ID(poll)">
+    <fieldset>
+        <p>What is your favorite flightless bird?</p>
+        <p>Choose your favourite flightless bird and you will be redirected to a page showing what other users have chosen.
```
sebastianbenz: There is no redirect happening...

sebastianbenz:
Two more things after running the sample:

* you don't show the results for answers without votes
* I think it would look better without the space between each percentage div:

screen shot 2016-12-14 at 20 37 10

backend/poll.go
kul3r4: I can't use nill, I get the error:
"cannot use nil as type PollForm in return argument"

kul3r4:
@sebastianbenz PTAL

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	return value, nil
+}
+
+func submitPoll(w http.ResponseWriter, r *http.Request) {
+	EnableCors(w, r)
+	SetContentTypeJson(w)
+	context := appengine.NewContext(r)
+	pollForm, error := parsePollForm(w, r, context)
+
```
sebastianbenz: nit: delete empty line

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+	if error != nil {
+		handleError(error, w)
```
sebastianbenz: does it make sense to continue the execution here?

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+	answer, answerErr := parseNotEmptyFormValue(r, "answer")
+	clientId, clientIdErr := parseNotEmptyFormValue(r, "clientId")
+
+	pollForm := PollForm{clientId, answer}
+
+	error := parseFormErrors([]error{answerErr, clientIdErr})
+	if error != nil {
+		return pollForm, error
+	}
+	return pollForm, nil
```
sebastianbenz: Simply write:

```
return pollForm, error
```

instead of:
```
if error != nil {
 +		return pollForm, error
 +	}
 +	return pollForm, nil
```

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+	var poll Poll
+	err := datastore.Get(ctx, pollKey, &poll)
+
+	if err != nil {
+		return PollResult{}, err
+	}
+
+	//check if the user has already voted
+	existingClientIds := poll.ClientIds
+	for i := 0; i < len(existingClientIds); i++ {
```
sebastianbenz: Do you need the index? Otherwise: 

```
for _,element := range poll.ClientIds {
...
}
```

might be easier.

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+
+	//if the user hasn't already voted, return a welcome message
+	message = "Thanks for answering the poll!"
+	//add the clientId
+	poll.ClientIds = append(existingClientIds, clientId)
+
+	pollExistingAnswers := poll.PollAnswers
+	//get the lastest vote count for the answer
+	answerFound := false
+	for i := 0; i < len(pollExistingAnswers); i++ {
```
sebastianbenz: Same as above, do you need the index variable?

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return PollResult{}, err
+	}
+
+	//check if the user has already voted
+	existingClientIds := poll.ClientIds
+	for i := 0; i < len(existingClientIds); i++ {
+		if existingClientIds[i] == clientId {
+			//return a message to let the user know that has already voted
+			message = "You have already answered this poll. If you want to run this sample again, use an incognito window."
```
sebastianbenz: Good idea to mention the incognito window in the message!

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+//holds an answers and votes represented as an array, used for displaying
+type PollEntryResult struct {
+	Percentage []int
+	Answer     string
+}
+
+//holds the poll results and a message, used for displaying
+type PollResult struct {
```
sebastianbenz: I find it a bit confusing that a struct called PollResult contains a list of PollResults

backend/poll.go
```diff
@@ -0,0 +1,186 @@
...
+
+//holds an answer and votes, used for storing
+type PollAnswer struct {
+	Answer string
+	Votes  int
+}
+
+//holds the questions
+type PollQuestions struct {
+	Questions  map[string]string
```
sebastianbenz: Why do you need a map? Can't you simply use an array and identify questions via their array index?

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,121 @@
...
+                    </div>
+                    <span class="title">{{Answer}}</span>
+                </li>
+            {{/PollResults}}
+            </ul>
+        </template>
+    </div>
+    <div submit-error>
+      <template type="amp-mustache">
+        Error! Looks like something went wrong with your comment, please try to submit it again. {{error}}
```
sebastianbenz: `with your comment` ?

kul3r4:
@sebastianbenz  thanks for the code review, applied all the changes.

backend/housing.go
```diff
@@ -57,33 +56,14 @@ func parseForm(r *http.Request) (MortgageForm, error) {
 	// can't return nil in place of a struct, so creating one in case there are errors
 	mortgageForm := MortgageForm{price, deposit, interest, period}
 
-	err := parseFormErrors(priceErr, depositErr, interestErr, periodErr)
+	err := parseFormErrors([]error{priceErr, depositErr, interestErr, periodErr})
 	if err != nil {
 		return mortgageForm, err
 	}
 
 	return mortgageForm, nil
```
sebastianbenz: Simplify to:

```
return parseFormErrors([]error{priceErr, depositErr, interestErr, periodErr});
```

backend/poll.go
```diff
@@ -0,0 +1,180 @@
...
+	clientId := pollForm.ClientId
+
+	//get the poll
+	pollKey := datastore.NewKey(ctx, "Poll", "poll", 0, nil)
+	var poll Poll
+	//if it's the first time executing, there will be no poll yet
+	datastore.Get(ctx, pollKey, &poll)
+
+	//check if the user has already voted
+	existingClientIds := poll.ClientIds
```
sebastianbenz: it would be more efficient to store ClientIds in a set (or a map, as golang doesn't support sets out of the box)

backend/poll.go
```diff
@@ -0,0 +1,180 @@
...
+
+	//if the user hasn't already voted, return a welcome message
+	message = "Thanks for answering the poll!"
+	//add the clientId
+	poll.ClientIds = append(existingClientIds, clientId)
+
+	pollExistingAnswers := poll.PollAnswers
+	//get the lastest vote count for the answer
+	answerFound := false
+	for _, pollExistingAnswer := range pollExistingAnswers {
```
sebastianbenz: You could simplify this a bit by storing the vote count in an array -> the question id is the array index of the vote count. The advantage would be that you could replace the for loop with a single array access.

kul3r4:
@sebastianbenz PTAL

backend/poll.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func handlePoll(w http.ResponseWriter, r *http.Request, page Page) {
+	page.Render(w, pollQuestions)
+}
+
+func createPollResult(answers []int, message string) PollResult {
+	results := make([]PollEntryResult, len(questions))
+	totalAnswers := 0
+	 for _, num := range answers {
```
sebastianbenz: nit formatting

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,148 @@
...
+      }
+
+</style>
+</head>
+<body>
+
+<h4 id="header1">Flightless bird poll</h4>
+<!-- #### Poll -->
+<!--
+We use `amp-form` to realize a poll and the `CLIENT_ID` variable to identify the user,
```
sebastianbenz: Can you explain our approach here?

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,154 @@
...
+
+<h4 id="header1">Flightless bird poll</h4>
+<!-- #### Poll -->
+<!--
+We use `amp-form` to implement a poll. To avoid multiple votes by the same user, we need to identify which users already voted.
+We use a cookie called `POLL_USER_ID` to identify the user.
+Given in Safari the default behavior is to block third-party cookies from sites that have not been previously visited, this approach may not work if the page is visited via the AMP Cache. In this case we fallback to checking for duplicate entries via the `CLIENT_ID`.
+The `CLIENT_ID` is an unique identifier for the user when
+accessing a specific origin, for example the original page or the cached version of it, this means that the same user will generate two different
+values when accessing different origins.
```
sebastianbenz: Add: 'That's why we use a custom cookie to track cross origin access to the poll page.'

src/50_Samples_%26_Templates/Poll.html
```diff
@@ -0,0 +1,154 @@
...
+accessing a specific origin, for example the original page or the cached version of it, this means that the same user will generate two different
+values when accessing different origins. That's why we use a custom cookie to track cross origin access to the poll page.
+The `CLIENT_ID` variable can be passed via an amp-form by declaring an hidden input
+value with `default-value="CLIENT_ID(poll)`.
+Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/ab78ba87cbf0408ca0ddcfcba7e666e1c1152051/extensions/amp-form/amp-form.md#variable-substitutions). -->
+
+<form method="post" action-xhr="<%host%>/samples_templates/poll/submit" target="_blank"
+      id="poll1"
+      custom-validation-reporting="as-you-go">
+    <input name="clientId" type="hidden" value="CLIENT_ID(poll)" data-amp-replace="CLIENT_ID">
```
ithinkihaveacat: Should this be `CLIENT_ID(POLL_USER_ID)` instead of `CLIENT_ID(poll)`? I don't see where the `poll` cookie is created or read.

backend/poll.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	//check if the user has already voted by checking the user id cookie existance,
+	cookie, err := r.Cookie(POLL_COOKIE_NAME)
+	if err == nil {
+		//return the existing poll answers and message to let the user know that has already voted
+		return createPollResult(pollExistingAnswers, ALREADY_VOTED_MESSAGE), nil
+	}
+		//cookies don't work on Safari when accessing the cdn, check if the user has already
+		//voted by checking the client id
+		clientId := pollForm.ClientId
```
ithinkihaveacat: Why fall back to clientId if POLL_COOKIE_NAME doesn't work? Why not just use clientId in all cases?

src/50_Samples_%26_Templates/Poll.html
sebastianbenz: yes!

sebastianbenz: @kul3r4 can you fix?

backend/poll.go
sebastianbenz: The client id differs based on host -> that's why we use our own cookie. Pubs usually have already session ids in place -> that's why we recommend not to use the client id cookie.

ithinkihaveacat: Oh, I see. The name `POLL_USER_ID` makes it sounds as if the cookie is specific to the poll, rather than an existing user id of potentially long standing. (Also, it actually [gets set to the value of clientId](https://github.com/ampproject/amp-by-example/blob/master/backend/poll.go#L165)--that could be just a random number, right?)

sebastianbenz:
Good point. We should make this more explicit.

Issue #369, opened 2016-09-19
"Add to Cart" functionality using amp-form


jasti:
Show a sample page showing a bunch of items that have number input fields. Preferably using the existing e-commerce examples.

User can edit the items on the input field (either increment or decrement) and press "add to cart" to land on another page . (The non-amp that will display the state of the cart (for v1 this page doesn't itself need to be interactive)

sebastianbenz:
Have you got a non-AMP sample page?

jasti:
This is just from Amazon.com on mWeb. They have DD instead of an increment/ decrement and then "add to cart".
<img width="318" alt="screen shot 2016-09-21 at 6 10 21 pm" src="https://cloud.githubusercontent.com/assets/2366341/18734199/bdc141a6-8026-11e6-9155-36e1966009be.png">

Submitting results in updating the cart. 
<img width="336" alt="screen shot 2016-09-21 at 6 11 14 pm" src="https://cloud.githubusercontent.com/assets/2366341/18734205/e1738ae6-8026-11e6-8e8f-0f3f4fcf28ff.png">

sebastianbenz:
Thanks. DD makes it a lot easier.

jasti:
Cool, DD sounds good.

sebastianbenz:
Do we want a lightbox showing a dialog with two buttons: Go to Cart | Continue Shopping?

jasti:
That would be nice. I was thinking of something simple like navigating to the next page (AMP or non-AMP) showing the number of items added to the cart but yeah, lightbox sounds great.

mkhatib:
@sebastianbenz 

> Do we want the add to cart button to trigger a lightbox showing a dialog with two buttons: Go to Cart | Continue Shopping? (if that's possible)

That's possible. See our `forms.amp.html` for an example `on=submit-success:lightbox`. 

> simple like navigating to the next page

This is currently not supported for `POST` but can be done for `GET` requests. We're discussing adding a way to redirect user to another page after a submission but still not in.

sebastianbenz:
Lightbox is currently not possible as we can't implement a continue shopping functionality via the AMP Cache. On iOS, 3P cookies might be disabled, which means the seller can't create cookies to store a user id. 

A possible solution would be to support variable substitutions for amp-form actions. Then the seller could associate a shopping cart with the CLIENT_ID.

kul3r4:
This is also related to https://github.com/ampproject/amphtml/issues/5937

[...]

PR #518, opened 2016-12-07
Uses CLIENT_ID for shopping cart

kul3r4:
Solves #369 

@sebastianbenz PTAL
I noticed the call to the shopping cart get cached and the quantity update doesn't happen in the next call, is there a way of disabling the service worker for that call?

### DIFF
--- a/backend/product-listing.go
+++ b/backend/product-listing.go
@@ -15,8 +15,9 @@
 package backend
 
 import (
+	"container/list"
 	"encoding/json"
-	"html"
+	"fmt"
 	"html/template"
 	"io/ioutil"
 	"net/http"
@@ -25,9 +26,9 @@ import (
 )
 
 const (
-	SEARCH        = "search"
-	ADD_TO_CART   = "add_to_cart"
-	SHOPPING_CART = "shopping_cart"
+	SEARCH           = "search"
+	SHOPPING_CART    = "shopping_cart"
+	ADD_TO_CART_PATH = "/samples_templates/product/add_to_cart"
 )
 
 type ProductListingPage struct {
@@ -58,6 +59,10 @@ type ShoppingCartItem struct {
 	Quantity string `json:"quantity"`
 }
 
+type ShoppingCart struct {
+	ShoppingCart []ShoppingCartItem
+}
+
 func (p *Product) StarsAsHtml() template.HTML {
 	return template.HTML(p.Stars)
 }
@@ -67,6 +72,8 @@ type JsonRoot struct {
 }
 
 var products []Product
+var shoppingCarts map[string]ShoppingCart
+var clientIds *list.List
 
 func InitProductListing() {
 	initProducts(DIST_FOLDER + "/json/related_products.json")
@@ -74,6 +81,43 @@ func InitProductListing() {
 	RegisterSample("samples_templates/product_listing", renderProductListing)
 	RegisterSample("samples_templates/product", renderProduct)
 	RegisterSampleEndpoint("samples_templates/product_listing", SEARCH, handleSearchRequest)
+	http.HandleFunc(ADD_TO_CART_PATH, func(w http.ResponseWriter, r *http.Request) {
+		handlePost(w, r, addToCart)
+	})
+	shoppingCarts = make(map[string]ShoppingCart)
+	clientIds = list.New()
+}
+
+func addToCart(w http.ResponseWriter, r *http.Request) {
+	EnableCors(w, r)
+	response := ""
+	name := r.FormValue("name")
+	quantity := r.FormValue("quantity")
+	clientId := r.FormValue("clientId")
+	img := r.FormValue("img")
+	price := r.FormValue("price")
+
+	shoppingCartItem := ShoppingCartItem{
+		Img:      img,
+		Name:     name,
+		Price:    price,
+		Quantity: quantity,
+	}
+
+	if clientIds.Len() == 100 {
+		delete(shoppingCarts, clientIds.Front().Value.(string))
+		clientIds.Remove(clientIds.Front())
+	}
+	shoppingCartItems := []ShoppingCartItem{shoppingCartItem}
+	shoppingCarts[clientId] = ShoppingCart{ShoppingCart: shoppingCartItems}
+	clientIds.PushFront(clientId)
+
+	if clientId != "" {
+		response = fmt.Sprintf("{\"ClientId\":\"%s\"}", clientId)
+		w.Write([]byte(response))
+	} else {
+		w.WriteHeader(http.StatusBadRequest)
+	}
 }
 
 func initProducts(path string) {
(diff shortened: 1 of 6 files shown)


backend/product-listing.go
```diff
@@ -67,13 +72,52 @@ type JsonRoot struct {
...
+	price := r.FormValue("price")
+
+	shoppingCartItem := ShoppingCartItem{
+		Img:      img,
+		Name:     name,
+		Price:    price,
+		Quantity: quantity,
+	}
+
+	if clientIds.Len() == 100 {
```
sebastianbenz: s/==/!=/ in case there are concurrent requests

src/50_Samples_%26_Templates/Product.html
```diff
@@ -125,24 +136,35 @@
...
                   </select>
-                   <input type="submit" class="button button-primary" id="add-to-cart"value="ADD TO CART">
+                   <input type="submit" class="button button-primary" id="add-to-cart" value="ADD TO CART">
                    <input type="hidden" name="name" value="Apple">
                    <input type="hidden" name="price" value="$1.99">
                    <input type="hidden" name="img" value="/img/product1_1024x682.jpg">
+                   <input name="clientId" type="hidden" default-value="CLIENT_ID(cart)">
+                   <div class="submit-success" submit-success>
+                        <template type="amp-mustache">
+                            <a class="button button-primary go-to-cart" href="<%host%>/shopping_cart/?clientId={{ClientId}}">GO TO CART</a>
```
sebastianbenz: s/GO TO CART/go to cart/ the button rule should convert it to uppercase

src/50_Samples_%26_Templates/Product.html
```diff
@@ -125,24 +136,35 @@
...
+                        </template>
+                    </div>
+                    <div submit-error>
+                        <template type="amp-mustache">
+                            Error! Looks like something went wrong with your comment, please try to submit it again. {{error}}
```
sebastianbenz: Adjust error message to this sample

static/sw.js
```diff
@@ -90,8 +90,8 @@ function requestAccepts(request, contentType) {
 function ampByExampleHandler(request, values) {
   // for samples show offline page if offline and samples are not cached
   if (requestAccepts(request, 'text/html')) {
-    // never use cached version for AMP CORS requests (e.g. amp-live-list)
-    if (request.url.indexOf("__amp_source_origin") != -1) {
+    // never use cached version for AMP CORS requests (e.g. amp-live-list) or pages open from href like shopping_cart
+    if (request.url.indexOf("__amp_source_origin") != -1 || request.url.indexOf("shopping_cart") != -1) {
```
sebastianbenz: Are URL variable substitutions already supported? In this case you could simply use the RANDOM variable substitution in the URL, e.g.:

`href="<%host%>/shopping_cart/?clientId={{ClientId}}&"`

Otherwise: can you make this more generic so that we can use it in other samples as well, e.g. add a query parameter `shopping_cart/?cache=0`.

backend/product-listing.go
```diff
@@ -144,3 +183,8 @@ func handleSearchRequest(w http.ResponseWriter, r *http.Request, page Page) {
 	route := page.Route + "?" + SEARCH + "=" + r.FormValue(SEARCH)
 	http.Redirect(w, r, route, http.StatusSeeOther)
 }
+
+func handleAddToCartRequest(w http.ResponseWriter, r *http.Request, page Page) {
```
sebastianbenz: where is this used?

backend/product-listing.go
```diff
@@ -67,13 +72,52 @@ type JsonRoot struct {
...
+
+	shoppingCartItem := ShoppingCartItem{
+		Img:      img,
+		Name:     name,
+		Price:    price,
+		Quantity: quantity,
+	}
+
+	if clientIds.Len() == 100 {
+		delete(shoppingCarts, clientIds.Front().Value.(string))
```
sebastianbenz: You could use an existing LRU Cache implementation, e.g.: https://github.com/golang/groupcache/blob/master/lru/lru.go

backend/product-listing.go
```diff
@@ -67,13 +72,52 @@ type JsonRoot struct {
 }
 
 var products []Product
+var shoppingCarts map[string]ShoppingCart
+var clientIds *list.List
```
sebastianbenz: Why are you storing the client ids?

backend/product-listing.go
```diff
@@ -67,13 +72,52 @@ type JsonRoot struct {
...
+		Price:    price,
+		Quantity: quantity,
+	}
+
+	if clientIds.Len() == 100 {
+		delete(shoppingCarts, clientIds.Front().Value.(string))
+		clientIds.Remove(clientIds.Front())
+	}
+	shoppingCartItems := []ShoppingCartItem{shoppingCartItem}
+	shoppingCarts[clientId] = ShoppingCart{ShoppingCart: shoppingCartItems}
```
sebastianbenz: won't this always delete any existing shopping cart?

kul3r4:
@sebastianbenz I addressed the comments, PTAL

backend/product-listing.go
```diff
@@ -90,14 +126,10 @@ func initProducts(path string) {
...
 func renderShoppingCart(w http.ResponseWriter, r *http.Request, page Page) {
-	page.Render(w,
-		ShoppingCartItem{
-			Img:      html.UnescapeString(r.URL.Query().Get("img")),
-			Name:     r.URL.Query().Get("name"),
-			Price:    r.URL.Query().Get("price"),
-			Quantity: r.URL.Query().Get("quantity"),
-		},
-	)
+	clientId := r.FormValue("clientId")
```
kul3r4: The shopping cart page is an AMP page so I don't think I can use `history.replaceState(null, null, location.pathname)`, right?

kul3r4:
@sebastianbenz I added the cookie and the redirect, PTAL

backend/product-listing.go
```diff
@@ -89,15 +127,37 @@ func initProducts(path string) {
...
+func redirectToShoppingCart(w http.ResponseWriter, r *http.Request, page Page, clientId string) {
+	expireInOneDay := time.Now().AddDate(0, 0, 1)
+	cookie := &http.Cookie{
+		Name:    ABE_CLIENT_ID,
+		Expires: expireInOneDay,
+		Value:   clientId,
+	}
+	http.SetCookie(w, cookie)
+	route := page.Route
+	route = strings.Split(route, "?")[0]
```
sebastianbenz: can you add a comment `// remove CLIENT_ID from URL

sebastianbenz:
Forgot one thing, formatting for error messages is off:

screen shot 2016-12-09 at 12 58 15

kul3r4:
@sebastianbenz PTAL

src/50_Samples_%26_Templates/Product.html
```diff
@@ -124,36 +131,48 @@
...
+    <amp-carousel width="512" height="384" layout="responsive" type="slides">
+        <amp-img src="/img/product1_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
+        <amp-img src="/img/product1_alt1_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
+        <amp-img src="/img/product1_alt2_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
+    </amp-carousel>
+
+
+    <!-- #### Product Quantity -->
+    <!-- This dropdown is realized using [amp-form](/components/amp-form/) and allows to select how many items to be added to the shopping cart. Pressing the `ADD TO CART` button leads to a shopping cart page. We are using the `CLIENT_ID` variable to identify the user. This variable can be used inside an amp-form by declaring an input hidden value with `default-value="CLIENT_ID(cart)`.  In this sample, the server is using it to map the user to a shopping cart. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md)-->
```
sebastianbenz: * ' Pressing the `ADD TO CART` button leads to a shopping cart page. ' is no longer correct
* Could you maybe elaborate a bit more what 'identify the user' means? I think it is important to explain here why we need to identify the user and what's the problem we're solving. 
* s/input hidden value/hidden input value/
* Can you explain 'In this sample, the server is using it to map the user to a shopping cart.'?

*

PR #519, opened 2016-12-07
add mustache

erwinmombay:


kickstart/gulpfile.js
```diff
@@ -16,64 +16,56 @@
...
 const del = require('del');
+const mustache = require('gulp-mustache');
 const posthtml = require('gulp-posthtml');
 const postcss = require('gulp-postcss');
 const runSequence = require('run-sequence');
 
+function getData() {
+  return JSON.parse(require('fs').readFileSync('data.json'));
+}
+
```
camelburrito: will all the data be in one file? we would want different data for different pages/templates i dont think one file would be enough right?


will it be possible/useful to inline the data on the page? wdyt?

erwinmombay: currently its in one file and we just namespace it (which is why the template has to do {{index.message}} instead of just {{message}}). I'll see if we can split it up in a follow up PR

PR #536, opened 2016-12-17
add amp-selector sample

sebastianbenz:


sebastianbenz:
@kul3r4 @camelburrito PTAL

src/20_Components/amp-selector.html
```diff
@@ -0,0 +1,115 @@
+<!--
```
kul3r4: We need to add the experiment tag here.

sebastianbenz: It works and validates out of the box.

sebastianbenz:
Added experiment flag - ptal

kul3r4:
LGTM

PR #539, opened 2017-01-02
Fix CORS access via www.ampbyexample.com

### LINKED ISSUE #538: CORS error on Form Submission with Page Update
I'm not sure if this example was intentionally designed this way to allow users to see both error and success states, but...

When trying the **Form Submission with Page Update** sample on https://www.ampbyexample.com/components/amp-form/, it results in:

`Error! Thanks for trying the amp-form demo with an error response.`

This message is caused by:

`Uncaught Error: Form submission failed:: Returned AMP-Access-Control-Allow-Source-Origin is not equal to the current: https://ampbyexample.com vs https://www.ampbyexample.com​​​`

The example works at https://ampbyexample.com/components/amp-form/, which makes sense since the origin matches.

sebastianbenz:
This also removes the buildSourceOrigin helper method which is no longer
needed (using the origin header is usually a better choice).

Fixes #538

### DIFF
--- a/backend/amp-form.go
+++ b/backend/amp-form.go
@@ -50,15 +50,6 @@ func submitFormXHRInputText(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte(response))
 }
 
-func submitFormInputText(w http.ResponseWriter, r *http.Request) {
-	name := r.FormValue("name")
-	if isUserTryingTheInputTextErrorDemo(name) {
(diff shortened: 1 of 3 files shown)


sebastianbenz:
@kul3r4 PTAL

kul3r4:
Does it work with localhost?

sebastianbenz:
yes - the origin header contains the host + scheme

kul3r4:
LGTM

PR #547, opened 2017-01-19
Moves comment into doc section to not break the page

kul3r4:
@sebastianbenz PTAL

### DIFF
--- a/src/50_Samples_%26_Templates/Poll.html
+++ b/src/50_Samples_%26_Templates/Poll.html
@@ -91,7 +91,8 @@ accessing a specific origin, for example the original page or the cached version
 values when accessing different origins. That's why we use a custom cookie to track cross origin access to the poll page.
 The `CLIENT_ID` variable can be passed via an amp-form by declaring an hidden input
 value with `default-value="CLIENT_ID(poll)`.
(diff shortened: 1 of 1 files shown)


src/50_Samples_%26_Templates/Poll.html
```diff
@@ -91,7 +91,8 @@ <h4 id="header1">Flightless bird poll</h4>
 values when accessing different origins. That's why we use a custom cookie to track cross origin access to the poll page.
 The `CLIENT_ID` variable can be passed via an amp-form by declaring an hidden input
 value with `default-value="CLIENT_ID(poll)`.
-Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/ab78ba87cbf0408ca0ddcfcba7e666e1c1152051/extensions/amp-form/amp-form.md#variable-substitutions). -->
+Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/ab78ba87cbf0408ca0ddcfcba7e666e1c1152051/extensions/amp-form/amp-form.md#variable-substitutions).
+Currently, HTML tables are not supported in form responses. -->
```
sebastianbenz: add "that's why we use CSS to layout the result data."

kul3r4: @sebastianbenz added, thanks!

PR #551, opened 2017-01-23
add AMP playground

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

playground/serve.go
```diff
@@ -0,0 +1,88 @@
...
+		return
+	}
+	param := r.FormValue("url")
+	u, err := url.Parse(param)
+	if param == "" || err != nil || (u.Scheme != "http" && u.Scheme != "https") {
+		http.Error(w, "Bad request.", http.StatusBadRequest)
+		return
+	}
+	// only allow URLs from ampbyexample
+	if u.Host != "ampbyexample.com" &&
```
kul3r4: what about putting those 3 strings in an array of allowed URLs instead of enumerating those?

playground/serve.go
```diff
@@ -0,0 +1,88 @@
...
+	ctx := appengine.NewContext(r)
+	client := urlfetch.Client(ctx)
+	req, err := http.NewRequest("GET", u.String(), nil)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Bad gateway (%v)", err.Error()),
+			http.StatusBadGateway)
+		return
+	}
+	req.Header.Add("User-Agent",
+		"Mozilla/5.0 (Linux; Android 6.0.1; Nexus 5X Build/MTC19V) "+
```
kul3r4: why are these needed?

sebastianbenz: In case pages have a redirect for desktops.

playground/serve.go
sebastianbenz: good idea. done.

sebastianbenz:
PTAL

kul3r4:
LGTM

PR #553, opened 2017-01-24
Lint backend/*.go files

ithinkihaveacat:


sebastianbenz:
nice! would be great if you could move `test -z $(gofmt -l backend/*.go)` to the gulpfile and make it part of the validate task? 

```
gulp.task('validate:backend', 'Validate the go backend code', function(){
  return run('test -z $(gofmt -l backend/*.go)').exec();
});
```

sebastianbenz:
Oh - and there is another go file `server.go` in the root folder we need to check.

ithinkihaveacat:
Fixed, I didn't think to look in the gulpfile. It now also checks `server.go` and `api/server.go`.

PR #556, opened 2017-01-24
Show different items in the cart

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

backend/product-listing.go
```diff
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io"
```
kul3r4: if we import "io", could we remove "io/ioutil"?

kul3r4:
LGTM

PR #569, opened 2017-01-30
Fix sample formatting in playground

sebastianbenz:
- amp-form: spacing around inputs
- don't show playground link for samples without code
- amp-fit-text: but each box in a new line
- amp-selector: fix list item height


sebastianbenz:
@dandv PTAL

kul3r4:
LGTM

PR #577, opened 2017-02-01
Attribute values consistently quoted and spaced

strikemike2k:
- Consistently quoting HTML attribute values
- Consistent spacing before end of tags
- Consistent HTML5 tag endings
- Trim unnecessary trailing space of modified files

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ Use HTML comments (`<!-- ... -->`) to document your sample code:
 
 ```html
 <!-- Look! Images in AMP. -->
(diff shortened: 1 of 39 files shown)


strikemike2k:
I signed it!

sebastianbenz:
Thanks so much for doing this! Hope you didn't do it all by hand ;-). Which tool did you use?

LGTM.

strikemike2k:
I used a regular expression search and replace, going through each result to verify it was correct.

PR #590, opened 2017-02-02
Remove unused amp-sidebar

kul3r4:
Including unused libraries creates warnings, better to remove them.
@sebastianbenz PTAL

### DIFF
--- a/src/50_Samples_%26_Templates/Product.html
+++ b/src/50_Samples_%26_Templates/Product.html
@@ -25,7 +25,6 @@ This sample showcases how to build a product page in AMP HTML.
     <script async custom-element="amp-lightbox" src="https://cdn.ampproject.org/v0/amp-lightbox-0.1.js"></script>
     <script async custom-element="amp-youtube" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>
     <script async custom-element="amp-form" src="https://cdn.ampproject.org/v0/amp-form-0.1.js"></script>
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Afaik - the product sample uses a sidebar (see templates/product/header.html).

kul3r4:
Right, it's imported. I think the problem is in the preview. I was looking at [product preview](https://ampbyexample.com/samples_templates/product/preview/). If you inspect the page, I can't find the amp-sidebar being used, and I can see the warning in the console.

sebastianbenz:
This is actually a problem in the product sample: the sidebar should work.

kul3r4:
We may need a way to remove the amp-sidebar import when building the preview page, as it doesn't have a menu and we are still importing the library.

sebastianbenz:
Nope. The preview page should show the sidebar. It's defined in the sample source.

kul3r4:
oh yes, you are right, it's there. However when I click on it I get:
Uncaught Error: Action Error: target not found in [tap:drawermenu.toggle]​​​
    at na.f.assert (log.js:285)
    at Uf (action-impl.js:286)
    at Kf.f.trigger (action-impl.js:270)
    at HTMLDocument.<anonymous> (action-impl.js:146)
sounds like this need more investigation.

sebastianbenz:
Found the problem, this line `{{{elementsAfterBody}}}` needs to go directly after `<body>` in templates/preview.html.

sebastianbenz:
Fixed it as part of #612

Issue #557, opened 2017-01-25
Enhance product sample by using amp-bind


kul3r4:
Enhance current product sample by using amp-bind:

- select the corresponding preview thumbnail while scrolling the carousel
- update the carousel image when choosing the color

Moreover, we could:

- make some fields dependent on selection, like showing the size only for some color

kul3r4:
@sebastianbenz 
Regards: "update the carousel image when choosing the color", at the moment the color is implemented using amp-selector and the value for any option is a string describing the color, as 

`<amp-img option="green" selected="" src="/img/product1_60x40.jpg" width="60" height="40" ....`

If we want to select the corresponding color when we select an image from the carousel by using amp-bind, I would need to make the option a number, like:

`<amp-selector name="color_index" layout="container" [selected]="selectedSlide" on="select:AMP.setState(selectedSlide=event.targetOption)">
..
                <amp-img option="0" selected src="/img/product1_60x40.jpg" width="60" height="40"></amp-img>`

were `selectedSlide` is the variable that amp-bind sets and it's based on number.

Would it be ok if I change this and then I add in the backend the logic for mapping between numbers and a string describing the color?

sebastianbenz:
A more realistic use case would be to have different image sets for different colors, e.g. two images for a red apple, which only shows if color red is selected etc. Is this possible?

kul3r4:
I think so, I am trying this implementation currently.

choumx:
Will this example cover @cramforce's e-commerce request in https://github.com/ampproject/amphtml/pull/7127#issuecomment-274645079?

If not, I'll create a new tracking issue that either @kul3r4 or I could implement.

kul3r4:
Hi, I am following most of the suggestions.

From the list:
- an image gallery: Yes
- a color selector: Yes
- a size selector: Yes
- support for availability (some items are sold out): @cramforce Can you expand a bit more on this?
- support for varying prices and sales specific to colors: I was planning to have different sizes specific to colors. For example, for color red only size S would be available. For the price, I will add that as well.

cramforce:
@kul3r4 Basically when you select a color, the size selector should disable the sizes that are not available in that color.

kul3r4:
ok, got it. Yes, will include those in the sample.

ericlindley-g:
[...]

PR #593, opened 2017-02-03
Product amp bind

kul3r4:
Solves #557 
Current status is [here](https://chiara-ampbyexample.appspot.com/samples_templates/product-bind/preview/). Please ignore errors related to "AMP-Access-Control-Allow-Source-Origin is not equal to the current", it will disappear in master. Opt out from dev channel, it work with the latest build but not with the latest dev-channel.


kul3r4:
@sebastianbenz PTAL. You need to opt out from the dev channel if you are in and enable the relevant examples.

sebastianbenz:
Can you rebase against master?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -158,6 +165,34 @@
     form#order amp-selector:not([disabled]) [option]:not([selected]):not([disabled]) amp-img {
       opacity: 0.5;
     }
+    .hide{
+      display: none;
+    }
+    .show{
```
sebastianbenz: nit formatting.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -158,6 +165,34 @@
     form#order amp-selector:not([disabled]) [option]:not([selected]):not([disabled]) amp-img {
       opacity: 0.5;
     }
+    .hide{
```
sebastianbenz: nit formatting.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -158,6 +165,34 @@
...
+      display: none;
+    }
+    .show{
+      display: block;
+    }
+    .show-size-unavailable{
```
sebastianbenz: nit formatting.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -158,6 +165,34 @@
...
+      align-items: center;
+      justify-content: center;
+      font-size: 4em;
+      color: grey;
+      opacity: 0.3;
+      height: 40px;
+      min-width: 40px;
+      width: auto;
+      position: absolute;
+      filter: alpha(opacity=30); /* For IE8 and earlier */
```
sebastianbenz: remove this as AMP doesn't support IE8 anyway

src/50_Samples_%26_Templates/Product.html
```diff
@@ -195,8 +230,35 @@
     </script>
 
     <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+    <amp-state id="product">
+      <script type="application/json">
+      {
+        "selectedSlide": "0",
```
sebastianbenz: Is there a particular reasons why the value is a string and not an int?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -313,11 +415,20 @@ <h5 class="price-description">Price: $1.99</h5>
...
         <div>
           <label for="size">Size:</label>
-          <amp-selector name="size" layout="container">
+          <amp-selector name="size" layout="container" on="select:AMP.setState(selectedSize=event.targetOption)" [selected]="allColors[selectedCarousel].availableSizes.indexOf(selectedSize) >= 0 ? selectedSize : allColors[selectedCarousel].availableSizes[0]">
             <ul>
-              <li option="S">S</li>
-              <li option="M" selected>M</li>
+              <div class="size-container">
+                <li option="S">S</li>
+                <span class="show-size-unavailable" [class]="allColors[selectedCarousel].availableSizes.indexOf('S') >= 0 ? 'hide' : 'show-size-unavailable'">X</span>
```
sebastianbenz: Can we use CSS to show when a size is unavailable? This makes it also possible to format the X to better fit into the box. Another option is to simply remove the shadow and lower the opacity.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -313,11 +415,20 @@ <h5 class="price-description">Price: $1.99</h5>
         </div>
         <div>
           <label for="size">Size:</label>
-          <amp-selector name="size" layout="container">
+          <amp-selector name="size" layout="container" on="select:AMP.setState(selectedSize=event.targetOption)" [selected]="allColors[selectedCarousel].availableSizes.indexOf(selectedSize) >= 0 ? selectedSize : allColors[selectedCarousel].availableSizes[0]">
             <ul>
-              <li option="S">S</li>
-              <li option="M" selected>M</li>
+              <div class="size-container">
```
sebastianbenz: Do we need the div here? I'm not sure if it's valid html if an `ul` is not followed by a `li`, but in any case we should keep the number of divs to a minimum to maintain readability.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -249,53 +311,93 @@
         </div>
       </amp-lightbox>
     </div>
-
      <!-- #### Image Gallery -->
     <!-- The `amp-carousel` component works very well for product image galleries. We use the [amp-selector component](/components/amp-selector/) to implement a film strip like slide preview. Clicking on a preview image will reveal the image in the carousel via the tap action `on="tap:product-carousel-preview.goToSlide(index=0)"`. Note: the preview is only visible on desktops.  -->
```
sebastianbenz: Can you adjust the documentation to reflect amp-bind?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -249,53 +311,93 @@
...
+    </amp-carousel>
+
   </div>
 
-    <h5 class="price-description">Price: $1.99</h5>
     <!-- #### Product Configuration -->
-    <!-- The add to cart action is implemented using  [amp-form](/components/amp-form/). In our sample we use the [amp-selector](/components/amp-selector/) for selecting different product properties. Pressing the `ADD TO CART` button adds the product to a shopping cart page using the properties you have selected. Notice that the button URL contains the query `clientId={{ClientId}}`. If an item is successfully added to the shopping cart, we redirect the user to the shopping cart page. The redirect target is configured via a header in the server's form response ([redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)):
+    <!-- We are using [amp-bind](https://www.ampproject.org/docs/reference/components/amp-bind) to update the price of the product depending on the colour
+      selected. We configure the price for each colour in the `amp-state` component.
+      We bind to the `text` attribute: we specify that the value of text is "Price:"" concatented with the
```
sebastianbenz: Is there a way we can avoid concatenating the string?

src/50_Samples_%26_Templates/Product.html
kul3r4: What do you mean? The span containing the X is styled by the css class `show-size-unavailable`, depending on the value of the expressions the span is shown or hidden.

sebastianbenz: I usually try to avoid using html/text (which should be for content/structure) for styling and use css instead. The problem is, for example, that the X doesn't fit correctly into the square (e.g. the lines don't go into the corners).

src/50_Samples_%26_Templates/Product.html
kul3r4: we could put the whole string in the json instead of splitting the price number and the string "Price:"

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: this was not used, removed it

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: removed, thanks

src/50_Samples_%26_Templates/Product.html
kul3r4: I removed the span and styled it with css

kul3r4:
What's the problem you are having when running it? Make sure you are using amp version 1486153081489, which is the latest stable and not the one in dev channel.

src/50_Samples_%26_Templates/Product.html
sebastianbenz: why not put the price into a span?

kul3r4: Done

sebastianbenz: can you update the doc as well

src/50_Samples_%26_Templates/Product.html
```diff
@@ -250,53 +315,97 @@
...
+      value of expression `allColors[selectedCarousel].price`. When this expression will be evaluated,
+      it will read the price value from the json inside the `amp-state`.
+      Notice how we set a default value for the price: this is because the value defined inside `amp-state`
+      will not be accessed at the page load, but only after a mutation in the bind expression:
```
sebastianbenz: nit: two colons in one sentence

src/50_Samples_%26_Templates/Product.html
```diff
@@ -159,6 +169,43 @@
     form#order amp-selector:not([disabled]) [option]:not([selected]):not([disabled]) amp-img {
       opacity: 0.5;
     }
+    .hide {
+      display: none;
+    }
+    .show {
+      display: block;
+    }
+    .show-size-unavailable {
```
sebastianbenz: do we still need this?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -159,6 +169,43 @@
...
+      display: flex;
+    }
+    .size-container {
+    position: relative;
+    background: gray;
+    overflow: hidden; //hide overflow of pseudo elements
```
sebastianbenz: `//` is not a valid comment in css

src/50_Samples_%26_Templates/Product.html
```diff
@@ -198,6 +245,24 @@
     <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
     </head>
     <body>
+    <amp-state id="allColors">
```
ithinkihaveacat: This data structure feels a bit weird. It's ultimately tabular data, but stored as a associative array, which sometimes has key names ('availableSizes') and sometimes drops key names ('color'). Also, it has a very specific name: `allColors`. (In a production-ready version, since this will all need to be machine generated, it would probably end up something generic like `product` or `price`.)

I'm wondering if it would make things simpler if each dimension (color, size, price, etc.) were mapped to successive levels of the associative array:

```javascript
{
  'green': {
    'S': null,
    'M': '$1.99',
    'L': '$1.99'
  },
  'golden': {
    'S': '$9.99',
    ...
  }   
}
```

This can then be indexed as e.g. `price[color][size]`.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -2,7 +2,7 @@
   "preview": "default",
   "skipValidation": "true",
   "default": "preview",
-  "experiments": ["amp-form-var-sub"]
+  "experiments": ["amp-form-var-sub", "amp-bind"]
   }--->
```
sebastianbenz: amp-selector is already out of experiment. Most likely related to a stale cache. Fixed in #600.

src/50_Samples_%26_Templates/Product.html
kul3r4: Thanks for the suggestion, will update the code

src/50_Samples_%26_Templates/Product.html
kul3r4: please see my above comment: we could put the whole string in the json instead of splitting the price number and the string "Price:", is this what you meant

src/50_Samples_%26_Templates/Product.html
ithinkihaveacat: Huh, good question. The `amp-bind` expression language is a bit limited. It would be nice if `Object.keys` were available. Is there any chance of getting that whitelisted? (`Object.values`, too.) Is there any reason why various `String.*` and `Array.*` methods are available, but no `Object.*`?

So, to make sure I understand, you need to determine the first available size so that an available product is selected on page load? I don't completely understand the example nor `amp-bind` but perhaps the easiest way given that it doesn't seem to be possible to loop through the data structure is to define another variable that stores the "on load" state:

```javascript
{
  'size': 'S',
  'color': 'green'
}
```

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
@kul3r4 can you please fix the html lint errors.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -198,10 +276,40 @@
     <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
     </head>
     <body>
+    <amp-state id="product">
+      <script type="application/json">
+      {
+          "green": {
+            "sizes": {
+        			"S": "$5.99",
```
ithinkihaveacat: Some inconsistent indentation (tabs/spaces issue?).

kul3r4:
Thanks, html lint errors fixed.

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
There still seem to be html validation errors. Can you also rebase against master? This should make the build more stable.

sebastianbenz:
Here is some (mostly) high-level feedback. I haven't looked into the details yet, but it might make sense to work on the high-level issues first. 

**Formatting:** 

-  an you add a space after two single quotes, s/''"/'' "/ it took me a while to understand that it wasn't a typo.
- can you add empty lines between the different carousels / film strips? I think that would make it easier to visually separate these.
- As we disabled auto line breaks during compilation, we have to manually make samples more readable. Can you add linebreaks before attributes (when tags span multiple lines), e.g.:

```html
<amp-img on="tap:AMP.setState(selectedSlide=0)" src="/img/red_apple_1_60x40.jpg" width="60" height="40" [class]="selectedSlide == 0 ? 'selected' : ''"></amp-img>
```
->
```html
<amp-img src="/img/red_apple_1_60x40.jpg" 
         on="tap:AMP.setState(selectedSlide=0)" 
         [class]="selectedSlide == 0 ? 'selected' : '' "
         width="60" height="40"></amp-img>
```

**Documentation:**

- Image Gallery section: not sure this works, but can we split the docs into two code sections, the film strip and the carousel? We might have to remove the enclosing `<div class="product-gallery">` for this.
- Remove 'Additionally used AMP components must be imported in the header.' it doesn't provide any value as most imported compoenents are hidden in the annotated source view
- Maybe consider splitting the docs up into smaller paragraphs to make them easier to scan.
- You start explaining amp-bind, then explain the client id handling and go back to amp-bind. Maybe explain amp-bind first and then client id handling later. It might make sense to give each part a different headline (e.g. Product Configuration, User Identity).
- Can you add a doc section to the amp-state element? It might make sense to move the declaration to a place where it makes more sense in the doc flow.

**Design:**

- film strip preview: remove the green outline around use different elevations instead similar to the quantity buttons
- color selector
   - use the accent color (pink) for the outline around the colors. 
   - add opacity (maybe 0.8) to non-active buttons to more clearly distinguish the active state visually
- Whenever you show an outline, make sure the outline doesn't increase the size (which looks weird)
- disabled button, add `user-select: none;` to avoid accidentally adding a blue selection marker

kul3r4:
@sebastianbenz Applied the suggested changes.
Few notes:
- Image Gallery section: not sure this works, but can we split the docs into two code sections, the film strip and the carousel? We might have to remove the enclosing <div class="product-gallery"> for this. -> I splitted the two while I kept the `div`. The reason is that it helps me applying css classes to the children given that `amp-bind` removes the original `class` when you bind to `class`.
- You start explaining amp-bind, then explain the client id handling and go back to amp-bind. Maybe explain amp-bind first and then client id handling later. It might make sense to give each part a different headline (e.g. Product Configuration, User Identity). -> iI moved client id paragraph at the end. The reason that I didn't split them is because they are both part of the tag `form`. Every headline needs code and in this case I wouldn't be able to do that

src/50_Samples_%26_Templates/Product.html
```diff
@@ -2,7 +2,7 @@
   "preview": "default",
   "skipValidation": "true",
   "default": "preview",
-  "experiments": ["amp-form-var-sub"]
+  "experiments": ["amp-form-var-sub", "amp-bind"]
   }--->
 <!--
 #### Introduction
```
sebastianbenz: Add an empty comment `<!-- -->` after the introduction to make the intro print before the code. Currently the whole header + CSS is shown in the annotated source mode ... which is very long.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -144,21 +149,90 @@
...
+    -webkit-transform: rotate(45deg);
+    transform: rotate(45deg);
+}
+
+div.square {
+  width: 40px;
+  height: 40px;
+}
+
+div.square.green {
```
sebastianbenz: nit: it doesn't make sense to combine green with div and square. Green would make sense for other elements as well.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -278,53 +346,177 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+                layout="responsive">
+        </amp-img>
+        <amp-img src="/img/red_apple_2_1024x793.jpg"
+                width="1024" height="682"
+                layout="responsive">
+        </amp-img>
     </amp-carousel>
+
   </div>
+  <!-- #### Product Configuration -->
```
sebastianbenz: There are now two sections called Product Configuration

src/50_Samples_%26_Templates/Product.html
```diff
@@ -278,53 +346,177 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+      value of expression `product[selectedColor].price`. When this expression is evaluated,
+      it gets the price value from the json inside the `amp-state`.-->
+    <!--Notice how we set a default value for the price: this is because the value defined inside `amp-state`
+      will not be accessed at the page load. We set `green` and `S` as default color and default size
+      as the correspondent variables have null value at default.-->
+    <!--As soon as the state of this variable changes, the expression is evaluated.
+      See below to understand how we change the state of the 2 variables `selectedColor` and `selectedSize`. -->
+      <p class="price-description">Price: <span [text]="product[selectedColor || 'green'].sizes[selectedSize || 'S']">$1.99</span>
+    </p>
+    <!-- The add to cart action is implemented using  [amp-form](/components/amp-form/). In our sample we use the [amp-selector](/components/amp-selector/) for selecting different product properties.
```
sebastianbenz: s/add to cart action/add-to-cart action/

sebastianbenz: This feels like a different topic - it might make sense to start a new paragraph here.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -278,53 +346,177 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+    The redirect target is configured via a header in the server's form response ([redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)):
 
       access-control-expose-headers:AMP-Access-Control-Allow-Source-Origin,AMP-Redirect-To
-      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789
+      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789 -->
 
-    We are using the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
+      <!-- We are using `amp-selector` together with `amp-bind` to update the page for showing the price, the size availability, the gallery and preview of images.
```
sebastianbenz: This reads weird at this point - can we move it earlier to the other part where we explain amp-bind?

src/50_Samples_%26_Templates/Product.html
kul3r4: done

src/50_Samples_%26_Templates/Product.html
kul3r4: done

kul3r4:
@sebastianbenz I applied the changes and some small changes to improve readability

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+                  [class]="selectedSlide == 1 ? 'selected' : '' ">
+         </amp-img>
        </li>
      </ul>
-    <amp-carousel id="product-carousel" width="1024" height="682" layout="responsive" type="slides">
-        <amp-img src="/img/product1_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
-        <amp-img src="/img/product1_alt1_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
-        <amp-img src="/img/product1_alt2_1024x682.jpg" width="1024" height="682" layout="responsive"></amp-img>
+   </div>
+   <!-- #### Image Gallery -->
```
sebastianbenz: idea: we don't need to show all three carousels in the source code view. We could move the documentation before the first carousel.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789 -->
 
-    We are using the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
+    <!-- We use the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
 
       <form id="order" method="POST"
             action-xhr="<%host%>/samples_templates/product/add_to_cart"
             target="_top">
         <div>
```
kul3r4: Would a flex layout be ok for you when removing this div?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+             "M":"$7.99",
+             "L":"$7.99"
+          },
+          "defaultSize":"S"
+       }
+     }
+  </script>
+</amp-state>
+
+    <!-- #### Product price -->
```
kul3r4: done

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
                          data-videoid="lBTCB7yLs8Y">
             </amp-youtube>
           </div>
         </div>
       </amp-lightbox>
     </div>
+    <!-- #### Slides preview Gallery -->
+    <!--On desktop we've added a film strip like slide preview. Each product colour has its own image gallery and slide preview.
+    We use [amp-bind](https://www.ampproject.org/docs/reference/components/amp-bind) to update the page selecting the gallery and image preview
+    corresponding to the selected colour. We realized the color selection by using `amp-selector`. Find more the doc and info in The Product Page section below.
```
kul3r4: done

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
-                           height="270"
+              <amp-youtube width="480" height="270"
                          layout="responsive"
                          data-videoid="lBTCB7yLs8Y">
             </amp-youtube>
           </div>
         </div>
       </amp-lightbox>
     </div>
+    <!-- #### Slides preview Gallery -->
```
kul3r4: done

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+     }
+  </script>
+</amp-state>
+
+    <!-- #### Product price -->
+    <!-- We use [amp-bind](https://www.ampproject.org/docs/reference/components/amp-bind) to update the price of the product
+      depending on the selected colour.
+
+    We bind the `text` attribute to the
+      value of expression `product[selectedColor].price`. `product` is the id `amp-state` json described in the Product Configuration section.
```
kul3r4: done

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+
+    Notice how we set a default value for the price: values defined inside `amp-state`
+      will not be accessed at the page load. We set `green` and `S` as default color and default size
+      as the correspondent variables have `null` value at default.
+
+     We use `amp-selector` together with `amp-bind` to update the page for showing the price, the size availability, the gallery and preview of images.
+    Every time you select a color, we set the expression `selectedColor=event.targetOption`. This is triggering
+    all the attributes and elements binded to `selectedColor` to update: the price (binded with the attribute `[text]`), the gallery, the image preview and the size selector (all binded with `[class]`). -->
```
kul3r4: done

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +340,208 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+      <p class="price-description">Price:
+        <span [text]="product[selectedColor || 'green'].sizes[selectedSize || 'S']">$1.99</span>
+    </p>
 
-    <h5 class="price-description">Price: $1.99</h5>
-    <!-- #### Product Configuration -->
-    <!-- The add to cart action is implemented using  [amp-form](/components/amp-form/). In our sample we use the [amp-selector](/components/amp-selector/) for selecting different product properties. Pressing the `ADD TO CART` button adds the product to a shopping cart page using the properties you have selected. Notice that the button URL contains the query `clientId={{ClientId}}`. If an item is successfully added to the shopping cart, we redirect the user to the shopping cart page. The redirect target is configured via a header in the server's form response ([redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)):
+    <!-- #### Product page -->
```
kul3r4: What about "Product Features" or "Product Selection"?

src/50_Samples_%26_Templates/Product.html
kul3r4: we will hide the code, so this should be ok

src/50_Samples_%26_Templates/Product.html
sebastianbenz: Maybe use Product Configuration here and use another one above?

src/50_Samples_%26_Templates/Product.html
sebastianbenz: I don't understand how this relates to flex layout?

kul3r4:
@sebastianbenz PTAL

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +339,226 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+    Pressing the `ADD TO CART` button adds the product to a shopping cart page using the properties you have selected. Notice that the button URL contains the query `clientId={{ClientId}}`.
+    If an item is successfully added to the shopping cart, we redirect the user to the shopping cart page.
+    The redirect target is configured via a header in the server's form response ([redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)):
 
+      access-control-expose-headers:AMP-Access-Control-Allow-Source-Origin,AMP-Redirect-To
+      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789 -->
+
+    <!-- We use the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
```
sebastianbenz: Don't start a new comment section here -> this causes the configuration preview to show too late.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +339,226 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+    <!--
+    We use [amp-selector](/components/amp-selector/) together with [amp-bind](/components/amp-bind/) for selecting different product properties and to update the page for showing the price, the size availability, the gallery and preview of images.
 
-      access-control-expose-headers:AMP-Access-Control-Allow-Source-Origin,AMP-Redirect-To
-      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789
+    Every time you select a color, we set the expression `selectedColor=event.targetOption`.
+    This is triggering all the attributes and elements bound to `selectedColor` to update:
```
sebastianbenz: s/This is triggerin/This triggers/

src/50_Samples_%26_Templates/Product.html
```diff
@@ -269,62 +339,226 @@ <h4 class="category"><a href="/samples_templates/product_listing/preview/">More<
...
+    the price (bound with the attribute `[text]`), the gallery, the image preview and
+    the size selector (all bound with `[class]`).
 
-    We are using the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
+    The add-to-cart action action is implemented using  [amp-form](/components/amp-form/).
+    Pressing the `ADD TO CART` button adds the product to a shopping cart page using the properties you have selected. Notice that the button URL contains the query `clientId={{ClientId}}`.
```
sebastianbenz: ' Notice that the button URL contains the query `clientId={{ClientId}}`.' how can the reader notice this without seeing the code?

src/50_Samples_%26_Templates/Product.html
sebastianbenz: The CLIENT_ID is mentioned in the previous paragraph (without explanation). Can we somehow merge these two sections?

src/50_Samples_%26_Templates/Product.html
kul3r4: This was an old phrase, the button doesn't contain the client ID now. Will remove it.

src/50_Samples_%26_Templates/Product.html
kul3r4: Done

src/50_Samples_%26_Templates/Product.html
kul3r4: Done

kul3r4:
@sebastianbenz I removed couple of "We use".

kul3r4:
@ericlindley-g could you take a look at the [latest version](https://chiara-ampbyexample.appspot.com/samples_templates/product/preview/)?

ericlindley-g:
Looks great — I like the use of solid colors for the color selection, and by and large the interactions are really slick.

A few points of feedback:

• (bug) If the first thing I change is the size selection, the entire carousel disappears. It reappears if then I make a color selection.

• (feature request) Can you make it so the price multiplies according to different quantity selections? (I think this should be possible, and would be great to demo, to show that AMP can give the user the accurate price before the item goes in the cart)

• (bug) I'm still seeing the bug where the first carousel thumbnail doesn't affect the carousel slide index
To repro:
1. On desktop, pick either the red apple or green apple with the color selector
2. click on the second picture of the apple to advance to the second slide in the carousel
3. Click on the first picture of the apple to return to the first slide in the carousel
Expected: should return to the first slide in the carousel
Actual: nothing happens

• (bug) When I submit the form and land on the shopping cart, I notice that often it's not updated with the data added with that submission. Is it possible to make sure it includes the latest data I've submitted?

• (feature request) Carousel thumbnails hide on mobile (would be nice to keep these, to demo the functionality, but I'm open to discuss—I know there's a tradeoff here because the carousel would have to be smaller if the thumbnails are positioned to the side of the carousel)

kul3r4:
@ericlindley-g 
Thanks!
• (bug) If the first thing I change is the size selection, the entire carousel disappears. It reappears if then I make a color selection -> that was my fault, will deploy a new version with that fixed

• (feature request) Can you make it so the price multiplies according to different quantity selections? (I think this should be possible, and would be great to demo, to show that AMP can give the user the accurate price before the item goes in the cart) -> I checked Amazon, if you select a product and then change the quantity, the price shown doesn't change, only when you go to the cart you'll se the multiplied price

• (bug) I'm still seeing the bug where the first carousel thumbnail doesn't affect the carousel slide index -> which version of AMP are you using?is it 1486769961849?

• (bug) When I submit the form and land on the shopping cart, I notice that often it's not updated with the data added with that submission. Is it possible to make sure it includes the latest data I've submitted? -> I checked on [code on master](https://ampbyexample.com/samples_templates/product/preview/) and the same problem is there, I guess this should be a separate issue to fix

ericlindley-g:
Thanks!

• re: changing price — now that you mention it, I think the way you have it now is right. It could be confusing to change the price based on quantity selection, because users would have to back-solve to get the per-unit price. Sounds good to leave as-is for now. :)

• Re: AMP version — Looks like my version of AMP is behind:
1485459680127

Apologies for the naive question: I'm opted into the dev channel — is there something else I should do to use the right version to test this feature?

kul3r4:
Try to clear everything: Chrome console -> Application -> Clear Storage (after that you will need to reenable the experiment by clicking on the button provided in the sample) and see if the version updates

ericlindley-g:
Got on the right version and that one bug is fixed — thanks!

kul3r4:
@sebastianbenz Any other comments?

sebastianbenz:
@kul3r4 can you rebase and merge?

kul3r4:
I rebased and solved conflicts. As for merging, do you mean in master?

sebastianbenz:
whenever we're ready @ericlindley-g

ericlindley-g:
LGTM — excited for this to go live!

Issue #574, opened 2017-01-31
Request: additional examples for amp-bind (amp-video[src] + XHR)


ericlindley-g:
The initial example for amp-bind is great!

I have a request for two additional examples to include:

1. change amp-video[src] on tap (this would be like the example of changing amp-img[src] on tap, but would use amp-video instead)

2. Change content based on XHR triggered by tap
The example could be as simple as tapping a button to update text from "cat" to "rabbit" (or some other animal — whatever text is returned via XHR), but could alternately simulate tapping a "like" button (following up in a separate request for this specific example).

WDYT?

/to @kul3r4 
/cc @choumx

choumx:
For reference, there are existing examples for these in the amphtml repository:
1. https://github.com/ampproject/amphtml/blob/master/examples/bind/basic.amp.html#L34
2. https://github.com/ampproject/amphtml/blob/master/examples/bind/xhr.amp.html#L27

sebastianbenz:
Sgtm. Both are good candidates for the amp-bind sample.

ericlindley-g:
Awesome—thanks! Do you have a sense for how soon these two examples in particular could go up?

kul3r4:
Integration between `amp-bind` and `amp-video` is [here](https://ampbyexample.com/components/amp-bind/).

Capturing conversation with @sebastianbenz: when working on the form sample with bind, we could also verify if it's possible to add content to state via the submit button. Could you come up with a scenario?

ericlindley-g:
@kul3r4 , are you looking for a scenario where the state is changed on form submission and response? One could be optimistically incrementing the count of "likes" after tapping on a <3 button, and then updating with the current value after the response comes in. (captured in [this issue](https://github.com/ampproject/amp-by-example/issues/575))

kul3r4:
@choumx i gave this a try [here](https://github.com/kul3r4/amp-by-example/blob/amp-bind-form/src/20_Components/amp-bind.html#L131).
I wanted to optistimatically update the count in 
`<p> Number of likes: <span [text]="(likesNumber+1)">1</span></p>` 

and then, in case of error, decrementing it by 
`<form method="POST" action-xhr="<%host%>/components/amp-form/submit-form-xhr"  target="_top"
      on="submit-error:AMP.setState(likesNumber = likesNumber - 1)">`
but I get errors like:

> submit-error:AMP.setState(likesNumber = likesNumber - 1) ] Expected either [,] or [)]

[...]

PR #594, opened 2017-02-03
Video with bind

kul3r4:
Related to #574 
@sebastianbenz PTAL

### DIFF
--- a/src/20_Components/amp-bind.html
+++ b/src/20_Components/amp-bind.html
@@ -102,6 +102,7 @@
   -->
   <amp-video src="/video/dog-video.mp4" [src]="allAnimals[currentAnimal].videoUrl" width=480 height=270 autoplay controls=true></amp-video>
 
+  <!-- #### Text data bindings -->
   <!--
   The text of this &lt;p&gt; will change when the `currentAnimal` variable
   changes.
--- a/src/20_Components/amp-bind.html
+++ b/src/20_Components/amp-bind.html
@@ -33,6 +33,7 @@
   -->
   <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
   <script async custom-element="amp-video" src="https://cdn.ampproject.org/v0/amp-video-0.1.js"></script>
+  <script async custom-element="amp-youtube" src="https://cdn.ampproject.org/v0/amp-youtube-0.1.js"></script>
   <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
 </head>
 <body>
@@ -45,10 +46,12 @@
     <script type="application/json">
       {
         "dog": {
-          "url": "/img/Border_Collie.jpg"
+          "url": "/img/Border_Collie.jpg",
+          "videoUrl": "/video/dog-video.mp4"
         },
         "cat": {
-          "url": "/img/cat-looking-up-300x200.jpg"
+          "url": "/img/cat-looking-up-300x200.jpg",
+          "videoUrl": "/video/cat-video.mp4"
         }
       }
     </script>
@@ -80,6 +83,9 @@
 
   `[property]="expression"`
 
+-->
+<!-- #### Attributes data bindings -->
+<!--
   In this example, the `src` of this &lt;amp-img&gt; element will change when
   either the `allAnimals` or `currentAnimal` implicit state variables change.
 
@@ -92,19 +98,18 @@
   </amp-img>
 
   <!--
+  The src of the video will change when the `currentAnimal` variable
+  changes.
+  -->
+  <amp-video src="/video/dog-video.mp4" [src]="allAnimals[currentAnimal].videoUrl" width=480 height=270 autoplay controls=true></amp-video>
+
+  <!--
   The text of this &lt;p&gt; will change when the `currentAnimal` variable
   changes.
   -->
   <p [text]="'This is a ' + currentAnimal + '.'">
   This is a dog.
   </p>
-
-  <!--
-  The src of the vide will change when the button is clicked ad the `videoSrc` variable
-  changes.
-  -->
-  <amp-video src="https://ampbyexample.com/video/tokyo.mp4" [src]="videoSrc" width=480 height=270 autoplay [controls]="videoControls"></amp-video>
-
 
   <!-- #### Triggering updates -->
   <!--
@@ -116,6 +121,6 @@
     &lt;amp-state&gt; components
   -->
   <button class="button-primary"
-      on="tap:AMP.setState(currentAnimal='cat', videoSrc='https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerJoyrides.mp4')">Set to Cat</button>
+      on="tap:AMP.setState(currentAnimal='cat')">Set to Cat</button>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


sebastianbenz:
I'm not sure the video sample fits into the triggering updates section (which is more of a conceptual nature). Having too many samples there makes things harder to understand. 

How about adding a new section: Use Cases? Would it possible to implement a video playlist with this? Maybe this would be a more realistic sample.

kul3r4:
@sebastianbenz I added a section around "attributes binding" and it contains the sample of binding with video and image src and another for "text binding" We could later add another sections showing class binding. It works with the latest build, not with the one in Canary, so please opt out from Canary while keeping the experiment amp-bind ON.

@ericlindley-g I tried to use amp-youtube trying to bind to the data-videoid. I guess it's not supported at the moment, as I get Uncaught Error: <AMP-YOUTUBE [data-videoid]> binding is not allowed.​​​ I could mention in the sample that the only video supported is amp-video at the moment, WDYT?

@choumx I noticed that the autoplay does not seem to work when the src changes. (When switching from dog to cat)

ericlindley-g:
@kul3r4 Sounds good to me — it's expected that only amp-video works for now, so that's okay. Thanks for adding this so quickly!

src/20_Components/amp-bind.html
```diff
@@ -44,10 +45,12 @@
     <script type="application/json">
       {
         "dog": {
-          "url": "/img/Border_Collie.jpg"
+          "url": "/img/Border_Collie.jpg",
```
sebastianbenz: s/url/imageUrl/

src/20_Components/amp-sticky-ad.html
```diff
@@ -16,6 +16,7 @@
     Import the `amp-sticky-ad` component in the header.
   -->
   <script async custom-element="amp-sticky-ad" src="https://cdn.ampproject.org/v0/amp-sticky-ad-1.0.js"></script>
+  <script async custom-element="amp-ad" src="https://cdn.ampproject.org/v0/amp-ad-0.1.js"></script>
```
sebastianbenz: isn't this change part of aother PR?

src/50_Samples_%26_Templates/News_Article.html
```diff
@@ -24,6 +24,7 @@
     <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
     <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
     <script async custom-element="amp-analytics" src="https://cdn.ampproject.org/v0/amp-analytics-0.1.js"></script>
+    <script async custom-element="amp-ad" src="https://cdn.ampproject.org/v0/amp-ad-0.1.js"></script>
```
sebastianbenz: isn't this change part of aother PR?

src/20_Components/amp-bind.html
```diff
@@ -44,10 +45,12 @@
...
+          "videoUrl": "/video/dog-video.mp4"
```
sebastianbenz: Can we maybe find a better video? I'll try to find a better one (I know I did send you the link, but didn't watch it to the end).

src/20_Components/amp-bind.html
```diff
@@ -91,6 +97,13 @@
   </amp-img>
 
   <!--
+  The src of the video will change when the `currentAnimal` variable
```
sebastianbenz: Won't it also change when allAnimals changes as explained above? (I know you copied it from below, just noticed the inconsistency).

sebastianbenz: Maybe use [this](https://pixabay.com/en/videos/border-collies-dogs-pets-playing-7259/) (matches the border collie image above).

src/20_Components/amp-bind.html
kul3r4: done

src/20_Components/amp-bind.html
kul3r4: done

src/20_Components/amp-bind.html
kul3r4: done

src/20_Components/amp-sticky-ad.html
kul3r4: done

src/50_Samples_%26_Templates/News_Article.html
kul3r4: done

kul3r4:
@sebastianbenz PTAL

src/20_Components/amp-bind.html
```diff
@@ -44,10 +50,13 @@
...
         "dog": {
-          "url": "/img/Border_Collie.jpg"
+          "imageUrl": "/img/Border_Collie.jpg",
+          "videoUrl": "/video/dog-video.mp4"
         },
         "cat": {
-          "url": "/img/cat-looking-up-300x200.jpg"
+          "imageUrl": "/img/cat-looking-up-300x200.jpg",
+          "videoUrl": "/video/cat-video.mp4",
+          "styleClass": "redBackground"
```
sebastianbenz: For consistency, can we have a different style for the dogs and set it as the initial background. This will be removed when the class is changed - right? If yes, we should mention this behavior. 

s/styleClass/style

src/20_Components/amp-bind.html
```diff
@@ -79,6 +88,25 @@
 
   `[property]="expression"`
 
+-->
+
+<!-- #### Text for textContent -->
+<!--
+The text of this &lt;p&gt; will change when the `currentAnimal` variable
```
sebastianbenz: If you use the exact same wording, that you should copy the whole thing. Add 'either the `allAnimals` or '.

src/20_Components/amp-bind.html
```diff
@@ -79,6 +88,25 @@
...
+-->
+<p [text]="'This is a ' + currentAnimal + '.'">This is a dog.</p>
+
+<!-- #### class for CSS classes -->
+<!--
+The class of this &lt;p&gt; will change when
+either the `allAnimals` or `currentAnimal` implicit state variables change.
+-->
+
+<p [class]="allAnimals[currentAnimal].styleClass">This text will have have a red background color when cat is selected.</p>
```
sebastianbenz: s/This text will have have a red background color when cat is selected./Each animal has a different background color./

src/20_Components/amp-bind.html
```diff
@@ -79,6 +88,25 @@
...
+
+<!-- #### Element-specific attributes -->
+<!--
   In this example, the `src` of this &lt;amp-img&gt; element will change when
```
sebastianbenz: I don't really like using the same sentence over and over again. If there is no new information we should make it shorter. For example:
```
Some amp-component provide custom binding targets for `amp-bind`. 

The src URL for videos embedded with `amp-video` can be changed.

`amp-img` URLs can be changed (does this include srcset as well?).
```
Can you also change the descriptions in the previous paragraphs?

src/20_Components/amp-bind.html
kul3r4: It doesn't make sense to add 'either the `allAnimals` or '. because the expression is only using currentAnimal variable.

src/20_Components/amp-bind.html
kul3r4: Done

src/20_Components/amp-bind.html
sebastianbenz: yes - my mistake

src/20_Components/amp-bind.html
kul3r4: It doesn't include srcset as you can't bind to it at the moment.

src/20_Components/amp-bind.html
kul3r4: done

kul3r4:
@sebastianbenz applied all the suggestions!

src/20_Components/amp-bind.html
```diff
@@ -79,24 +92,36 @@
...
+The text of this &lt;p&gt; will change when the `currentAnimal` variable
+changes. `currentAnimal` is an uninitialized implicit state variable, which will be
+   set when a state change is triggered (see below).
+
+-->
+<p [text]="'This is a ' + currentAnimal + '.'">This is a dog.</p>
+
+<!-- #### Binding Styles -->
+<!--
+  Style class can be changed. It will override the existing element style class. `allAnimals` is the `id` of the &lt;amp-state&gt; component above.
```
sebastianbenz: s/Style class can be changed./Styling can be changed by applying css classes./
s/the existing element style class/all the element's style classes/

src/20_Components/amp-bind.html
```diff
@@ -79,24 +92,36 @@
...
+-->
+
+<p [class]="allAnimals[currentAnimal].style" class="greenBackground">Each animal has a different background color.</p>
+
+<!-- #### Binding Attribute Values -->
+<!--
+Various AMP components support binding attribute values with amp-bind. `amp-img` URLs can be changed.
```
sebastianbenz: s/`amp-img`/For example, `amp-img`/

kul3r4:
@sebastianbenz Thanks, nits applied.

PR #596, opened 2017-02-04
Add amp-list custom spinner sample

dandv:
Customizing the `amp-list` spinner is not documented yet. (The fact that a spinner exists is not actually mentioned in the docs [at all](http://stackoverflow.com/questions/42034524/display-loading-indicator-while-amp-list-loads-the-content).) This sample is the first step towards fixing that.

### DIFF
--- a/backend/slow-response.go
+++ b/backend/slow-response.go
@@ -18,6 +18,8 @@ package backend
 
 import (
 	"net/http"
+	"strconv"
+	"fmt"
 	"time"
 )
 
@@ -26,14 +28,15 @@ const (
 )
 
 func InitSlowResponseSample() {
-	http.HandleFunc(SLOW_RESPONSE_SAMPLE_PATH +"rates", roomRates)
+	http.HandleFunc(SLOW_RESPONSE_SAMPLE_PATH +"", sleep)
 }
 
-func roomRates(w http.ResponseWriter, r *http.Request) {
+func sleep(w http.ResponseWriter, r *http.Request) {
 	EnableCors(w, r)
 	SetContentTypeJson(w)
-	response := "{\"items\":[{\"title\": \"This response was delayed. Reload the page if you didn't see the spinner.\"}]}"
-	duration := time.Duration(10)*time.Second
+	delay, _ := strconv.ParseUint(r.URL.Query().Get("delay"), 0, 64)
+	response := fmt.Sprintf("{\"items\":[{\"title\": \"This response was delayed %v milliseconds. Reload the page if you didn't see the spinner.\"}]}", delay)
+	duration := time.Duration(delay)*time.Millisecond
 	time.Sleep(duration)
 	w.Write([]byte(response))
 }
--- a/backend/slow-response.go
+++ b/backend/slow-response.go
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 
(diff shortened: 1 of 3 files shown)


backend/slow-response.go
```diff
@@ -0,0 +1,39 @@
...
+	"net/http"
+	"time"
+)
+
+const (
+	SLOW_RESPONSE_SAMPLE_PATH = "/" + CATEGORY_SAMPLE_TEMPLATES + "/slow-response/"
+)
+
+func InitSlowResponseSample() {
+	http.HandleFunc(SLOW_RESPONSE_SAMPLE_PATH +"rates", roomRates)
```
sebastianbenz: Can we find a better name than roomRates?

src/20_Components/amp-list.html
```diff
@@ -25,6 +25,38 @@
     bottom: 0;
     right: 0;
   }
+  <!--
+    #### CSS for the amp-list loader
```
sebastianbenz: Commenting inside the css block is unfortunately not supported (but would be very nice to have).  We have to add the docs before the `<style amp-custom>...`.

dandv: Yup, saw `The text (CDATA) inside tag 'style amp-custom' contains 'html comments', which is disallowed. (see https://www.ampproject.org/docs/reference/spec.html#stylesheets)` in the Travis logs, will fix.

backend/slow-response.go
dandv: I'll add a parameter to specify the delay, in milliseconds.

dandv:
> does this change loading animation for all animations

Yes, the loader is the same for `amp-iframe`, it turns out (since it's, as far as I can tell, [undocumented](https://github.com/ampproject/amphtml/issues/4299#issuecomment-277431321)).

src/20_Components/amp-list.html
dandv: Great, now we have `The text (CDATA) inside tag 'style amp-custom' contains 'CSS i-amphtml- name prefix', which is disallowed.`.

dandv:
> it would make more sense to have an advanced sample, e.g. 'Custom loading animations'.

Turns out it's [against spec to customize the loader's CSS classes](https://www.ampproject.org/docs/reference/spec#class-and-tag-names):

> the user's stylesheet may not reference CSS selectors for `-amp-` classes, `i-amp-` IDs and `i-amp-` tags and attributes. These classes, IDs and tag/attribute names are not meant to be customized by authors.

Looks like this sample is dead in the water (and I should've used the playground, which would've highlighted the error early on!).

sebastianbenz:
Too bad! Would've been a nice sample...

dandv:
@mkhatib - any chance to revisit allowing styling of the loading indicator?

dandv:
The first check fails with `Error: Missing error message` - what am I doing wrong?

sebastianbenz:
Sweet! Just a thought: this applies to all components with a loading indicator. Would it make more sense to turn this into a advanced sample, e.g. 'Custom Loading Indicators'?

sebastianbenz:
The build is fine. It's a problem in how we run our travis build. Should be fixed if you rebase against master (wondering why this didn't pop up earlier...).

dandv:
Submitted a new PR in #623.

PR #600, opened 2017-02-06
Use networkFirst for samples & AMP Runtime

### LINKED ISSUE #599: ServiceWorker: Use networkFirst caching strategy for samples & AMP runtime
...to avoid problems like [this](https://github.com/ampproject/amphtml/issues/7296#issuecomment-277385399).

sebastianbenz:
Make sure page gets always the latest AMP Runtime and sample
code. Also: cache all other requests for offline access.

Fixes #599

### DIFF
--- a/static/sw.js
+++ b/static/sw.js
@@ -96,8 +96,8 @@ function ampByExampleHandler(request, values) {
     if (request.url.indexOf("__amp_source_origin") != -1 || shouldNotCache(request)) {
       return toolbox.networkOnly(request, values);
     }
-    // cache or network - whatever is fastest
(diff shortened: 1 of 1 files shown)


sebastianbenz:
@kul3r4 PTAL

kul3r4:
Fine for me. Now that we are using `return toolbox.cacheFirst(request, values)` for the `if` and `else` (line 112, 121), would it be possible to remove the `else` and just have in the `if` the code about placeholder images?

sebastianbenz:
The difference in the two branches is the image fallback. Or did I
misunderstood your suggestion?

kul3r4:
LGTM

PR #621, opened 2017-02-13
Adds dropdown to amp-form

### LINKED ISSUE #609: build out amp-form with more input examples
Hi there!

A question from Google Webmaster Forum to see if would be possible to build out the amp-form example to include different kinds of inputs (eg dropdowns, radio buttons etc).

https://productforums.google.com/d/msg/webmasters/GpWhrLkeGUY/gcUWqG-UBwAJ

kul3r4:
Fixes #609 
@sebastianbenz PTAL

### DIFF
--- a/src/20_Components/amp-form.html
+++ b/src/20_Components/amp-form.html
@@ -50,6 +50,9 @@
   form {
     margin: 16px;
   }
+  form > select {
+    margin: 0 16px 0 0;
+  }
   </style>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-form.html
```diff
@@ -369,5 +372,28 @@
       </template>
     </div>
   </form>
+    
+  <!--
+    Use `select` element to realize a dropdown.
```
sebastianbenz: s/to realize a dropdown./for dropdowns.

sebastianbenz:
* while we're at it, can you please cleanup the css/formatting and replace the existing one with (should be a bit more consistent:

```
  <style amp-custom>
  form.amp-form-submit-success [submit-success],
  form.amp-form-submit-error [submit-error]{
    margin-top: 16px;
  }
  form.amp-form-submit-success [submit-success] {
    color: green;
  }
  form.amp-form-submit-error [submit-error] {
    color: red;
  }
  form.amp-form-submit-success.hide-inputs > input {
    display: none
  }
  .horizontal-display {
    display: flex;
    align-items: center;
  }
  form > input {
    margin: 0 16px;
  }
  form label {
    margin: 8px;
  }
  form {
    margin: 0;
    padding: 16px;
  }
  </style>
```
* the <br> have no effect - I'd prefer to use css here instead of brs anyway.

kul3r4:
@sebastianbenz I applied the css from above, still the select neesd the `margin-top` to be set to 0 otherwise it does not align with the submit button.

sebastianbenz:
Thanks!

ckxion:
Thanks for your work on this. I was the Hacker (cannot call my self a programmer), who asked for an example form with all features on one form.

I see this is now at:
[](https://ampbyexample.com/components/amp-form/preview/)

Error report ( think) 
Name and Email subscribe section:
After successfully subscribing, the confirm message reads "Success! Thanks (name) for trying the amp-form demo! Try to insert the word "error" as a name input in the form to see how amp-form handles errors."

But
On mobile when I try entering name as Error I get same success message. On desktop I get "Error! Thanks error for trying the amp-form demo with an error response."

So something seems wrong on mobile (Android mobile)

Also, not sure the purpose of rejecting name as Error (error2 is accepted)

Thanks ColinK

PR #632, opened 2017-02-17
always generate a preview page

sebastianbenz:
This makes it possible to view the preview version even if it is not linked from the sample by appending /preview to the sample URL.

### DIFF
--- a/tasks/compile-example.js
+++ b/tasks/compile-example.js
@@ -221,43 +221,41 @@ module.exports = function(config, updateTimestamp) {
       });
 
       // compile example preview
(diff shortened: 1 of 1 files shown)


sebastianbenz:
@kul3r4 PTAL

kul3r4:
I think we found some cases in which previews didn't work or didn't make sense, should we disable some or check?

sebastianbenz:
They're not linked from anywhere so I wouldn't consider this a problem. This is more a hidden feature for creating videos / embeds.

kul3r4:
Sounds good. LGTM

PR #637, opened 2017-02-19
Star rating sample using amp-form and CSS

### LINKED ISSUE #558: Star rating sample using amp-selector and amp-form
Should be fairly straight forward to create a star rating sample using `amp-selector` and `amp-form ` with radio buttons. 

@dandv Is this something you can tackle?

dandv:
Please squash. Thanks!

Resolves #558.


src/50_Samples_%26_Templates/Star_Rating.html
```diff
@@ -0,0 +1,141 @@
...
+-->
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <link rel="canonical" href="<%host%>/samples_templates/star_rating/">
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <!-- #### Setup
```
sebastianbenz: nit: can you please place the import below the boilerplate and put the setup headline in a separate comment. This results in a better formatting in the sample view as both imports will be shown and no unrelevant code (this makes use of how the parser works to improve the formatting - I'd be happy to make this more explicit, but haven't got an idea how).

```
  <!-- #### Setup -->
  <!-- 
    Import the `amp-form` and `amp-mustache` components.
  -->
  <script async custom-element="amp-form" src="https://cdn.ampproject.org/v0/amp-form-0.1.js"></script>
  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>

  <!--
    #### CSS for the star rating
  -->
```

src/50_Samples_%26_Templates/Star_Rating.html
```diff
@@ -0,0 +1,141 @@
...
+    }
+    form.amp-form-submit-error [submit-error] {
+      color: red;
+    }
+  </style>
+</head>
+
+<body>
+
+  <h1>CSS-only keyboard-accessible rating widget</h1>
```
sebastianbenz: This will never show up as it is hidden by default (no associated docs). However, it'd be good to enable the preview mode via metadata where this would be really helpful.

src/50_Samples_%26_Templates/Star_Rating.html
dandv: Thanks for pointing that out. I must confess that the rules by which text is shown or hidden, elude me. The heading plus text after it, are an example. Will push a fix for that too.

src/50_Samples_%26_Templates/Star_Rating.html
dandv: How exactly do I prevent ABE from hiding my HTML?

```html
  <!--
  While this star rating widget is accessible via the keyboard, that doesn't happen without quirks — which we'll mention to the user.
  -->
  <p>Things to try on computers with keyboards and/or pointing devices:</p>
    <ul>
    <li>
      To test keyboard accessibility, press Tab twice, then arrows. Unfortunately, in Firefox and Safari, keyboard arrows work backwards.
      This issue appears to be <a href="https://github.com/LunarLogic/starability/issues/15">impossible to fix</a>
      on non-Chrome-based browsers without using JavaScript, so stay tuned for the
      <a href="https://www.ampproject.org/docs/reference/components/amp-rating">amp-rating component</a>.
    </li>
    <li>
      Mouse over the stars to see the style change
    </li>
  </ul>
```

Only the `<p>` is displayed.

dandv:
> I think this sample would better fit into the advanced category. Wdyt?

Agree. Feel free to move.

> Idea: would it make sense to initially highlight 2.5 stars to indicate that this is possible as well using server-side rendering?

The widget doesn't support half-stars.

sebastianbenz:
* Can you please move the sample.
* Regarding 2.5 stars, the point wasn't about the .5, but about showing an initial rating. 

Idea: how much effort is it to make the rating work? e.g. we have an initial rating of 2, once the user has rated we should the updated value.

dandv:
Setting the rating is as easy as setting `checked="checked"` on the desired radio button.

dandv:
> I think this is our best documented sample so far.

Thank you :) Let me know if there are other changed requested. Would love to see this live and be able to point partners to it (and move on to finalizing the proper component).

sebastianbenz:
Did you see my suggestion regarding whether it's possible to make it work like an actual rating? The initial rating is 2 and the value changes (reflected in the stars) after the rating is submitted. One approach would be to hide the initial stars on form submit success and render a new one with the form results. 

I'll merge this this now, but might be a good idea for another PR.

sebastianbenz:
Had another look at the sample - my proposal actually doesn't make sense. I think the current version is fine.

PR #639, opened 2017-02-19
fix-image-gallery

### LINKED ISSUE #631: Advanced carousels bind example: linked thumbnails updates w/ wrong number
Was excited to see the slide index UI for the advanced carousels use-case!

Just noticed a bug, though: Tapping on the thumbnails immediately below the carousel works well, but when tapping on the second row of thumbnails, the number updates strangely (seems to concatenate the two numbers):

screen shot 2017-02-17 at 8 36 58 am

/to @kul3r4

kul3r4:
Fixes #631 

When using amp-selector, the option is a string. When applying the mathematical a+1, given a is a string, the result will be "a+1".

Amp-bind doesn't support Number.parseInt so using *1 will ensure the cast to a number.

@sebastianbenz PTAL

### DIFF
--- a/src/30_Advanced/Image_Galleries_with_amp-carousel.html
+++ b/src/30_Advanced/Image_Galleries_with_amp-carousel.html
@@ -144,10 +144,10 @@ first described in text and paintings in medieval times.
   attribute in the `<p>` tag above the carousel to update as well.
 
   Only [whitelisted functions](https://www.ampproject.org/es/docs/reference/components/amp-bind)
-  are allowed when using `amp-bind`. We use the syntax `+ selectedSlide` to make sure the variable is casted into an integer.
+  are allowed when using `amp-bind`. We use the unary plus operator before selectedSlide to cast it into an integer.
   -->
   <div>
-    <p> Selected slide: <span [text]="(+ selectedSlide + 1)">1</span>/3</p>
+    <p> Selected slide: <span [text]="+ selectedSlide + 1">1</span>/3</p>
 
     <amp-carousel controls type="slides" width="400" height="300"
     [slide]="selectedSlide" on="slideChange:AMP.setState(selectedSlide=event.index)">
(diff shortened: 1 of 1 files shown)


src/30_Advanced/Image_Galleries_with_amp-carousel.html
```diff
@@ -144,7 +144,7 @@
   attribute in the `<p>` tag above the carousel to update as well.
   -->
   <div>
-    <p> Selected slide: <span [text]="(selectedSlide + 1)">1</span>/3</p>
+    <p> Selected slide: <span [text]="(selectedSlide * 1 + 1)">1</span>/3</p>
```
sebastianbenz: We should explain this. Out of curiosity, does this work: `+ selectedSlide + 1`? Might be a bit more idiomatic.

kul3r4:
@sebastianbenz it works with +. I updated the docs as well.

src/30_Advanced/Image_Galleries_with_amp-carousel.html
```diff
@@ -142,9 +142,12 @@
 
   Whenever the carousel slide changes, it updates `selectedSlide`. This causes the `text`
   attribute in the `<p>` tag above the carousel to update as well.
+
+  Only [whitelisted functions](https://www.ampproject.org/es/docs/reference/components/amp-bind)
+  are allowed when using `amp-bind`. We use the syntax `+ selectedSlide` to make sure the variable is casted into an integer.
```
sebastianbenz: s/We use the syntax `+ selectedSlide` to make sure the variable is casted into an integer./We use the unary plus operator before `selectedSlide` to cast it into an integer./

One more question: why are there parentheses around the expression?

kul3r4:
@sebastianbenz PTAL

src/30_Advanced/Image_Galleries_with_amp-carousel.html
```diff
@@ -142,9 +142,12 @@
 
   Whenever the carousel slide changes, it updates `selectedSlide`. This causes the `text`
   attribute in the `<p>` tag above the carousel to update as well.
+
+  Only [whitelisted functions](https://www.ampproject.org/es/docs/reference/components/amp-bind)
+  are allowed when using `amp-bind`. We use the unary plus operator before selectedSlide to cast it into an integer.
```
sebastianbenz: sorry, one last nit: can you place selectedSlide inside "`"

kul3r4:
@sebastianbenz done

PR #657, opened 2017-02-26
Ampstart samples

sebastianbenz:
Migrates the sample pages to the new layout:

* removes existing styles.css from sample page
* new one-column sample layout
* add sample outline (requires parsing headings in sample source -> no nested levels headings supported)
* preview and code listing are in one row
* preview no longer has a background color -> this means 
* fixed experiment card (no background color - still visible enough?)
* `hide: doc` is no longer supported. Added workaround for shopping card (the only samples that uses it)
* (unrelated) fix cards not wrapping on safari for landing page
* a few sample fixes (accordion, katura player, publishing amps)

Demo: https://amp-by-example-staging.appspot.com


sebastianbenz:
@kul3r4 PTAL

PR #658, opened 2017-02-27
Ampstart samples2

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL

https://amp-by-example-staging.appspot.com/introduction/hello_world/

templates/css/example.css
```diff
@@ -33,6 +33,8 @@ pre>code {
 .abe-code pre {
   overflow-x: auto;
   white-space: pre;
+  padding: 0;
```
kul3r4: Can we use basscss for these?

sebastianbenz: unfortunately not - I need to overwrite the more specific styling by ampstart's `www-component-desc .pre` rule.

PR #660, opened 2017-02-27
fix validation errors

sebastianbenz:
Disable amp-selector in example page template if the sample contains a
live-list or an amp-selector -> these can't be nested.


sebastianbenz:
@kul3r4 PTAL

sebastianbenz:
Also fixes a few styling related conflicts between amp-selector sample and main layout.

PR #670, opened 2017-03-02
Adds margin to amp-bind

kul3r4:
I add to add divs for the paragraphs as the class is overwritten by amp-bind. The alternative is to add a css rule using the parent but that would increase the css duplicating the style.

### DIFF
--- a/src/20_Components/amp-bind.html
+++ b/src/20_Components/amp-bind.html
@@ -41,6 +41,9 @@
     .greenBackground {
       background-color: green;
     }
+    .abe-preview > p {
+      padding: 1rem;
+    }
   </style>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-bind.html
```diff
@@ -101,27 +101,29 @@
...
+</div>
 
 <!-- ## Binding Styles -->
 <!--
   Styling can be changed by applying css classes. It will override all the element's style classes. `allAnimals` is the `id` of the &lt;amp-state&gt; component above.
 -->
-
-<p [class]="allAnimals[currentAnimal].style" class="greenBackground">Each animal has a different background color.</p>
-
+<div class="p1">
```
sebastianbenz: why add another div here?

src/20_Components/amp-bind.html
```diff
@@ -101,27 +101,29 @@
...
+  <p [class]="allAnimals[currentAnimal].style" class="greenBackground">Each animal has a different background color.</p>
+</div>
 <!-- ## Binding Attribute Values -->
 <!--
 Various AMP components support binding attribute values with amp-bind. For example, `amp-img` URLs can be changed.
 -->
-  <amp-img src="/img/Border_Collie.jpg" width="300" height="200"
+  <amp-img src="/img/Border_Collie.jpg" class="m1" width="300" height="200"
```
sebastianbenz: I'd prefer to keep the image full-bleed but use a responsive layout

src/20_Components/amp-bind.html
```diff
@@ -101,27 +101,29 @@
...
       [src]="allAnimals[currentAnimal].imageUrl">
   </amp-img>
 
   <!--
     The src URL for videos embedded with `amp-video` can be changed. At the moment, among all the AMP video components, only amp-video supports amp-bind.
   -->
-  <amp-video src="/video/dog-video.mp4" [src]="allAnimals[currentAnimal].videoUrl" width="300" height="200" autoplay controls="true"></amp-video>
+  <amp-video src="/video/dog-video.mp4" class="m1" [src]="allAnimals[currentAnimal].videoUrl" width="300" height="200" autoplay controls="true"></amp-video>
```
sebastianbenz: same as above, I'd prefer to keep the video full-bleed but use a responsive layout

src/20_Components/amp-bind.html
kul3r4: I explained this in the PR comment.

sebastianbenz: didn't notice - sorry!

I'm wondering whether it'd make sense to add a default margin for paragraphs in the preview, e.g.:

```
.abe-preview > p {
  margin: 1rem;
}

What do you think?

Comment (kul3r4) [inline src/20_Components/amp-bind.html]: Margin unfortunately doesn't work, snapshot attached.
screen shot 2017-03-02 at 3 27 48 pm
Adding a padding to the p itelf also doesn't work as it enlarges the background. I tonly works having a parent element with a padding

Comment (sebastianbenz) [inline src/20_Components/amp-bind.html]: I think `padding: 1rem;` is actually what we want.

**Comment by kul3r4** [comment]
@sebastianbenz PTAL

File: src/20_Components/amp-bind.html [hunk 4]
```diff
@@ -41,6 +41,9 @@
     .greenBackground {
       background-color: green;
     }
+    .abe-preview {
+      padding: 1rem;
```
sebastianbenz: I meant give parapgraphs inside the preview a padding (but only direct childs):

```
.abe-preview > p {
  padding: 1rem;
}
```

kul3r4: Sorry, my bad. all fixed.

src/20_Components/amp-bind.html
```diff
@@ -41,6 +41,9 @@
     .greenBackground {
       background-color: green;
     }
+    .abe-preview > p{
```
sebastianbenz: nit: formatting

kul3r4: thanks!

PR #671, opened 2017-03-02
Ampstart amp experiment and rebase

kul3r4:


kul3r4:
This code was doing a rebase from master, will open a different one for the amp-experiment change.

PR #674, opened 2017-03-02
Fixes margin and layout

kul3r4:
Fixes:

- amp-font
- amp-fx-flying-carpet
- amp-gfycat

### DIFF
--- a/src/20_Components/amp-font.html
+++ b/src/20_Components/amp-font.html
@@ -93,13 +93,13 @@
     on-load-add-class="tangerine-loaded">
   </amp-font>
     <!-- and use an existing font. -->
-  <p class="tangerine p1">
+  <p class="tangerine">
     This text is displayed in red if the chosen font is not supported.
   </p>
   <!--
     This time the font is not available.
   -->
-  <p class="unavailable-font p1">
+  <p class="unavailable-font">
     This text is displayed in red if the chosen font is not supported.
   </p>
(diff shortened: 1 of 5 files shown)


src/20_Components/amp-font.html
```diff
@@ -94,13 +93,13 @@
     on-load-add-class="tangerine-loaded">
   </amp-font>
     <!-- and use an existing font. -->
-  <p class="tangerine">
+  <p class="tangerine p1">
```
sebastianbenz: Shouldn't these be catched by our generic rule: 
```
.abe-preview p { 
  padding: 1rem;
}
```

src/20_Components/amp-fx-flying-carpet.html
```diff
@@ -100,14 +100,13 @@
 
      The text that follows is to ensure that the `amp-fx-flying-carpet` is placed before the last viewport.
   -->
-  <p>
+  <p class="p1">
```
sebastianbenz: same as above

kul3r4: I added that rule at the component sample level for amp-bind, should I move it to example.css?

sebastianbenz: yes

kul3r4: Just moved it, thanks

PR #677, opened 2017-03-03
amp-fx-parallax sample page

aghassemi:
please DO NOT MERGE until Canary is patched with https://github.com/ampproject/amphtml/pull/7942 
I will ping this PR when it is patched.

mcwjd8zt0pk

### DIFF
--- a/src/20_Components/amp-fx-parallax.html
+++ b/src/20_Components/amp-fx-parallax.html
@@ -1,0 +1,70 @@
+<!---{
+    "experiments": ["amp-fx-parallax"],
+    "preview": "default"
+}--->
+<!--
+  #### Introduction
+  The [amp-fx-parallax](https://www.ampproject.org/docs/reference/components/amp-fx-parallax) extension allows an element to move as if it is nearer or farther relative to the foreground of the page content.
+  As the user scrolls the page, the element scrolls faster or slower depending on the value assigned to the attribute.
+
+  Parallax effect can be enabled by adding `amp-fx-parallax=<factor>` **attribute** to any HTML or AMP.
+
+  The factor is a decimal that controls how much faster or slower the element would scroll relative to the scrolling speed.
+
+  A value greater than `1` scrolls the element upward (element scrolls faster) when the user scrolls down the page.
+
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-fx-parallax.html
```diff
@@ -0,0 +1,65 @@
...
+    "preview": "default"
+}--->
+<!--
+  #### Introduction
+  The [amp-fx-parallax](https://www.ampproject.org/docs/reference/components/amp-fx-parallax) extension allows an element to move as if it is nearer or farther relative to the foreground of the page content.
+  As the user scrolls the page, the element scrolls faster or slower depending on the value assigned to the attribute.
+
+  Parallax effect can be enabled by adding `amp-fx-parallax=<factor>` **attribute** to any HTML or AMP.
+
+  The factor is a decimal that controls how much faster or slower the element would scroll relative to the scrolling speed.
```
sebastianbenz: formatting it: Can you turn this into a list?

```
The factor is a decimal that controls how much faster or slower the element would scroll relative to the scrolling speed:

* A value greater than `1` scrolls the element upward (element scrolls faster) when the user scrolls down the page.
* ...
```

src/20_Components/amp-fx-parallax.html
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    .header {
+      position: relative;
+      overflow: hidden;
+    }
+  </style>
+</head>
+<body>
+  <!-- #### Basic Usage -->
```
sebastianbenz: It'd be great if you could provide samples for different factors. In that case it might make sense to rename the heading from `Basic Usage` to `Factor 1.7` etc...

aghassemi:
@sebastianbenz Done with comments. Also Canary is patched with the fix, so this can be merged.

FYI: https://github.com/ampproject/amphtml/projects/21 to track needed features and bugs for launch of parallax. Important feature we need is a component to create `scenes` where parallax effect can be "contained" within a specific parent (rather the whole document as it is now). Having that feature would allow us to put more parallax samples in the same page here.

sebastianbenz:
Woohoo!

PR #678, opened 2017-03-04
Update ampstart

sebastianbenz:
* add logo
* navbar highlight selected category
* migrate playground
* buttons on homepage have a transparent button now
* sidebar has improved spacing
* fix pre in doc right margin
* new footer divider

https://amp-by-example-staging.appspot.com/


sebastianbenz:
@kul3r4 PTAL - btw I've noticed that spacing in comment and amp-form sample is broken.

kul3r4:
Looks good, I was also thinking that on the main page we could remove "view" from the buttons so instead of "View code" and "View demo" we could simply have "Code" "Demo".
Will take a look at the broken spacing, it may be some domino effect!

sebastianbenz:
thanks!

PR #680, opened 2017-03-04
Ampstart font

sebastianbenz:


kul3r4:
LGTM

herschel666:
Is it intended to have the `templates/css/ampstart.css` file versioned in a minified state?

sebastianbenz:
@herschel666 this is a workaround until we have a better way to share the css with ampproject/ampstart.

PR #681, opened 2017-03-05
Update footer.snip.html

kul3r4:
This fixes the padding for the footer. There is a top-bottom padding rule inherited from ampstart-footer class.
Currently is: 
footer


sebastianbenz:
Just noticed the broken footer - and then saw your PR. Yay!

PR #691, opened 2017-03-11
Remove wrapping of links with buttons

herschel666:
I removed the button-elements, which makes imho no sense from a semantic point of view. Also it breaks the links in Firefox and makes them unusable.

Atm the `disabled`-styles are missing. I'll add them, when the [styles "issue"](https://github.com/ampproject/amp-by-example/pull/680#issuecomment-285877353) is solved.

### DIFF
--- a/templates/css/shared.css
+++ b/templates/css/shared.css
@@ -39,4 +39,7 @@ body {
 .ampstart-sidebar {
   width: 340px;
 }
-
+.ampstart-btn.disabled {
+  pointer-events: none;
+  opacity: 0.5;
+}
--- a/templates/css/shared.css
+++ b/templates/css/shared.css
@@ -43,3 +43,10 @@ body {
   pointer-events: none;
   opacity: 0.5;
 }
+.ampstart-btn {
+  color: #b60845;
+  border: 1px solid #b60845;
+}
+a.ampstart-btn-secondary {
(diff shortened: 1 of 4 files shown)


sebastianbenz:
Thanks for looking into this! The problem is that this should be directly fixed in ampstart. Until then the best approach would be to add the required rules in [templates/styles/shared.css](templates/styles/shared.css).

sebastianbenz:
(didn't intent to close the PR)

herschel666:
The styles for disabled `a[class=*"ampstart-btn"]`-elements are added. PTAL.

templates/css/shared.css
```diff
@@ -39,4 +39,7 @@ body {
 .ampstart-sidebar {
   width: 340px;
 }
-
+.ampstart-btn.disabled {
+  pointer-events: none;
+  opacity: 0.5;
+}
```
sebastianbenz: We also need to update the color which was previously handled by the button: 

```
.ampstart-btn {
  color: #b60845;
  border: 1px solid #b60845;
}
a.ampstart-btn-secondary {
  color: #fff;
}
```

sebastianbenz:
I took the liberty to add the suggested changes to your PR. 

Thanks a lot for looking into this!

PR #698, opened 2017-03-15
Add padding to menu

kul3r4:
This adds padding to the menu of the product-listing sample.
This is the current menu without padding:
menu-margin


sebastianbenz:
This looks broken overall - can you also style the links and title?

kul3r4:
@sebastianbenz The snapshot shows the current state, which looks broken and the PR is fixing it by adding the padding.
This is how it's looking with the fix:
menu-padding

sebastianbenz:
Yes. Can you remove the underline and adjust the text sizes?

kul3r4:
@sebastianbenz I added some style, it looks like this now:
menu-style

sebastianbenz:
Great - thanks for fixing this!

PR #701, opened 2017-03-16
fix cards overflowing in safari

sebastianbenz:
Bug introduced by #695


sebastianbenz:
@kul3r4 PTAL

sebastianbenz:
thanks for the review!

PR #702, opened 2017-03-16
Add sample for video ad

alanorozco:
Advanced example that uses the new multiple actions and hide/show features.

Fixes https://github.com/ampproject/amphtml/issues/7840


src/60_AMP_Ads/Video_Ad.html
```diff
@@ -0,0 +1,196 @@
...
+        Watch Video
+      </div>
+      <amp-img src="https://ampbyexample.com/img/tokyo.jpg"
+        width="300"
+        height="168">
+      </amp-img>
+    </div>
+    <div id="video-container"
+         hidden>
+      <amp-video id="my-video"
```
aghassemi: I'd put a `loop` on it

alanorozco: Done

src/60_AMP_Ads/Video_Ad.html
```diff
@@ -0,0 +1,197 @@
+<!---{
+ "preview": "a4a"
```
sebastianbenz: You can specify the ad dimensions for the preview here:

```
<!---{
 "preview": "a4a",
 "width": 300,
 "height": 168
}--->
```

src/60_AMP_Ads/Video_Ad.html
```diff
@@ -0,0 +1,197 @@
...
+  </style>
+</head>
+<body>
+
+  <!-- ## Full Markup -->
+  <!--
+  Create the necessary markup to show two layers that represent the two ad
+  states.
+
+  **Splash layer**
```
sebastianbenz: You can inline the code section to match the corresponding code blocks, e.g.:

```
<!--  A `<div>` that contains a static image, text, and a button to open the video  layer.-->
<div id="splash-container">
```

The sample doc view will then render the doc followed by a code section including the next html tag and it's content.

alanorozco: I did that initially. The example would then be broken since the two divs reference each other by id.

sebastianbenz: I see. This should still work if you put the comments above `splash-container` and `video-container`.

alanorozco: Not really, because the splash would hide on action and while it would "work" it would look broken" :)

sebastianbenz: I actually like it - it demonstrates the behavior of each section in isolation. -> You get the overall behavior in the demo view. You can make this more explicit by saying something like: 'the video layer is initially hidden, but will appear when you click 'watch video' above.'. 

I think it'd be nice to have the sample then split into four sections:

* Introduction
* Setup
* Splash Layer
* Video Layer

But I'm happy either way - pick whatever you think works best. It's your sample 😉.

sebastianbenz:
One other thing I've noticed: can we make the sample behave a bit more ad like and add a link to a landing page? e.g. to https://ampbyexample.com such as in the existing amp ad samples: https://ampbyexample.com/amp_ads/carousel_ad/preview/

src/60_AMP_Ads/Video_Ad.html
alanorozco: Done. I can change it back if it's deemed confusing :)

alanorozco:
@sebastianbenz ptal

sebastianbenz:
Looks great! One little problem: clicking the watch video button now opens the landing page link as well.

alanorozco:
Thanks for catching that! Fixed.

aghassemi:
:tada:

sebastianbenz:
Live: https://ampbyexample.com/amp_ads/video_ad/preview/ and FYI #714

PR #703, opened 2017-03-17
improve demo view for amp-bind sample

sebastianbenz:


sebastianbenz:
@kul3r4 PTAL - what do you. think about this approach?

kul3r4:
I agree, it doesn't look nice at the moment. LGTM

PR #708, opened 2017-03-20
fix amp list overflow

sebastianbenz:
* new button hasn't got enough space
* replace button with div to make it work

### DIFF
--- a/src/20_Components/amp-list.html
+++ b/src/20_Components/amp-list.html
@@ -47,7 +47,7 @@
   ```
   The list content is rendered via an [amp-mustache template](https://www.ampproject.org/docs/reference/components/amp-mustache). The template can be specified either by id, or by using a nested element.
   -->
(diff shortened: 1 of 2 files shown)


sebastianbenz:
@kul3r4 PTAL

kul3r4:
It is aligning on the right, is it correct? Wouldn't be better to be below the visible content?
Moreover, should we add some padding to the button?

sebastianbenz:
Below doesn't really work as it'll interfere with the list content. You're
right about the margin.

sebastianbenz:
Added margin. Fixed also that margins in preview didn't show.

kul3r4:
I think the margin is not enough for the button and we would need padding This is what I see:
button

sebastianbenz:
Sure? Works for me: https://amp-by-example-staging.appspot.com/components/amp-list/

kul3r4:
ops sorry, probably I was looking at an old version of the page.
LGTM

PR #720, opened 2017-03-27
AMP Ad: preview real DFP ads

sebastianbenz:
* update preview description to reflect the change
* add link to amp ad spec
* rephrase AMP Ad section overview
* render DFP ads in preview (Defined via metadata)


sebastianbenz:
@kul3r4 PTAL 

https://ampbyexample.com/#amp_ads

sebastianbenz:
Good point! Updated the preview to fallback to an iframe if there is no `adSlot`. Ad previews are also no longer copied to the api app engine instance, but served on the same domain without cookies.

PTAL

sebastianbenz:
thanks for review!

PR #721, opened 2017-03-27
Use an IE11-compatible font for amp-font examples

### LINKED ISSUE #718: Use IE11-compatible font in amp-font example
IE11 fails to load the [`Tangerine_Regular.ttf`](https://github.com/ampproject/amp-by-example/blob/master/src/fonts/Tangerine_Regular.ttf) font used in the [amp-font page](https://ampbyexample.com/components/amp-font/) with this error: `CSS3114: @font-face failed OpenType embedding permission Check. Permission must be Installable.`

Screenshot of Tangerine_Regular.tff failing to render in IE11

I recommend we replace Tangerine_Regular with another font that loads in IE11 so that users see amp-font works correctly in that browser. I know [`ComicAMP.ttf`](https://github.com/ampproject/amphtml/blob/master/examples/fonts/ComicAMP.ttf) works.

cvializ:
Closes #718

It works in IE11


cvializ:
Wonderful, thank you! I don't have merge permission so this can be merged any time.

PR #723, opened 2017-03-28
Use new AMP.setState() syntax

choumx:
Partial for https://github.com/ampproject/amphtml/issues/8390.

- Replace key-value syntax with object-literal syntax for `AMP.setState`
- Also some minor formatting fixes

/to @kul3r4

### DIFF
--- a/src/50_Samples_%26_Templates/Product.html
+++ b/src/50_Samples_%26_Templates/Product.html
@@ -341,7 +341,7 @@ a.item {
   <div class="product-gallery">
     <ul class="show" [class]="selectedColor == 'green' ? 'show' : 'hide'">
       <li tabindex="0" role="button">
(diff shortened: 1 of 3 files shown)


src/50_Samples_%26_Templates/Product.html
```diff
@@ -448,33 +448,33 @@ <h2 class="m1">Apple</h2>
...
-          "defaultSize":"M"
-       }
-     }
-  </script>
-</amp-state>
+      "green": {
+        "sizes": {
+          "S": "$5.99",
+          "M": "$5.99",
+          "L": null
```
kul3r4: I think 'unavailable' (as it was before) makes a lot of more sense for someone reading this configuration file.

choumx: Fair point, I was trying to reduce the length of the expression. Done!

PR #727, opened 2017-03-30
Add sample for amp-gist (#90)

### LINKED ISSUE #90: Embedding Gist code snippets

levidurfee:
Closes #90

### DIFF
--- a/src/20_Components/amp-gist.html
+++ b/src/20_Components/amp-gist.html
@@ -1,0 +1,42 @@
+<!--
+  ## Introduction
+
+  The amp-gist component allows embedding of an entire gist or a single file.
+-->
+<!-- -->
+<!doctype html>
+<html ⚡>
+<head>
+  <meta charset="utf-8">
+  <link rel="canonical" href="<%host%>/components/amp-gist/">
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <!-- ## Setup -->
+  <!-- Import the amp-gist component -->
+  <script async custom-element="amp-gist" src="https://cdn.ampproject.org/v0/amp-gist-0.1.js"></script>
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-gist.html
```diff
@@ -0,0 +1,42 @@
+<!--
+  ## Introduction
+
+  The amp-gist component allows embedding of an entire gist or a single file.
```
sebastianbenz: Can you please add links to the official documentation and to gist.github.com.

src/20_Components/amp-gist.html
```diff
@@ -0,0 +1,42 @@
...
+    "experiments": ["amp-gist"]
+  }--->
+</head>
+<body>
+
+  <!-- ## Basic Usage -->
+  <!-- Embed an entire gist. -->
+  <amp-gist
+    data-gistid="b9bb35bc68df68259af94430f012425f"
+    layout="responsive"
```
sebastianbenz: * What's the recommended way to layout a gist? Responsive seems wrong as the height is not likely to change. 
* Can you explain how to obtain the width and height?

levidurfee: After thinking about this for a little bit, `fill` seems to be more appropriate for this layout. That would also remove the need for `width` and `height` attributes.

I'll make these changes to the component and to the documentation later this evening.

Thanks! :)

sebastianbenz: Hm - just gave the sample a try. Fill is definitely the wrong choice. I'd assumed that fixed-height is the best one, as gist's usually have a fixed height (might even make sense to make this the only possible option). 

However, it seems like the layout is only relevant for the initial layout of the AMP, as amp-gist will resize to the actual height once the gist has been loaded (which can lead to quite bad layout jumps if you pick the wrong height).

levidurfee: ## Idea 1

What if we put a restriction on `amp-gist` that is similar to `amp-iframe`?

> Must be at least 600px or 75% of the first viewport away from the top.

Content developers may not know the actual height of their gist (I wouldn't) and the height would vary based on the width of their screen. Having that restriction would hopefully prevent the user from seeing any layout jumps.

## Idea 2

I also like how `amp-iframe` has the "Resize on user interaction" option. The user could specify the width and height and it would grow based on their interaction.

## Idea 3

I'll think on it and discuss it with some people on Slack.

jpettitt: I really dislike idea 1.  There is no way of knowing what 600 pixels will be.  If the user opens the doc on an iPad it's likely to be very different from an iPhone 4.

sebastianbenz: +1 to what John said. I'd be fine with keeping it as it is, but making it explicit in the docs that:

* fixed-height layout should be used
* providing an accurate height is important to avoid jumps (obtain height e.g. via Chrome Dev Tools on gist page - gists are usually used by devs, so I think it's fine to recommend Dev Tools)

levidurfee: Thanks @jpettitt and @sebastianbenz !

levidurfee: I've made those changes :)

levidurfee:
Yay :)

PR #730, opened 2017-03-30
Fix iframe embed layout

sebastianbenz:
Fixes the iframe breaking the layout:

screen shot 2017-03-30 at 8 22 21 pm


sebastianbenz:
@kul3r4 PTAL

PR #735, opened 2017-04-03
Support sub dirs in static

sebastianbenz:
This is to make it easier to structure static files associated with
samples.

Preparation for a sample demonstrating hreflangs + AMP.


sebastianbenz:
@kul3r4 PTAL

PR #736, opened 2017-04-03
Add example for amp-call-tracking

alanorozco:


src/20_Components/amp-call-tracking.html
```diff
@@ -0,0 +1,49 @@
...
+    }
+    ```
+
+    Note that `formattedPhoneNumber` is optional. See the [component documentation for more information](https://www.ampproject.org/docs/reference/components/ads/amp-call-tracking).
+
+    We also need to set a default phone number that will be shown before the endpoint responds. This is done via an `<a>` element inside the `amp-call-tracking` tag.
+  -->
+  <p>
+    <strong>Call us!</strong><br>
+    <amp-call-tracking config="https://ampbyexample.com/json/call-tracking-number.json">
```
sebastianbenz: Can you replace `https://ampbyexample.com/` with `<%host%>`. This way the sample will also work on the staging environment/localhost.

src/20_Components/amp-call-tracking.html
```diff
@@ -0,0 +1,49 @@
+<!--
+  ## Introduction
+
+  [amp-call-tracking](https://www.ampproject.org/docs/reference/components/ads/amp-call-tracking) allows to replace static phone numbers with dynamically generated phone numbers.
```
sebastianbenz: Can you add a sentence describing why this is useful? This is just to explain to someone like me who doesn't have a clue what this is good for.

src/20_Components/amp-call-tracking.html
alanorozco: Well, it will work partially since the element expects HTTPS :) `<%host%>` is better regardless, done.

src/20_Components/amp-call-tracking.html
alanorozco: Done. Let me know if it's still unclear.

PR #743, opened 2017-04-09
Add check to ensure variables have default value

### LINKED ISSUE #740: Clean up for product page demo
I noticed a few issues with the product page demo (https://ampbyexample.com/samples_templates/product/preview/):

- [ ] amp-form-var-sub is listed as a required experiment, but this is now launched, so that can be removed. (Presumably amp-bind is still needed?)
- [ ] From initial load: Click yellow, click green: Green shows an unavailable price
- [ ] From initial load: Click "S" and the page jumps
- [ ] From initial load: Click "S" [page jump, per above] and then click yellow and the page jumps
- [ ] From initial load: Click "S", then "M", then yellow, and the price says unavailable.

cc @ericlindley-g

kul3r4:
Fixes #740 
@sebastianbenz PTAL
@choumx FYI

### DIFF
--- a/src/50_Samples_%26_Templates/Product.html
+++ b/src/50_Samples_%26_Templates/Product.html
@@ -3,7 +3,7 @@
   "skipValidation": "true",
   "default": "preview",
   "hideCode": true,
-  "experiments": ["amp-form-var-sub", "amp-bind"]
+  "experiments": ["amp-bind"]
   }--->
 <!--
 ## Introduction
@@ -516,7 +516,7 @@ a.item {
                   on="select:AMP.setState({
                     selectedColor: event.targetOption,
                     selectedSlide: 0,
-                    selectedSize: product[event.targetOption].sizes[selectedSize] ? selectedSize : product[event.targetOption].defaultSize
+                    selectedSize: product[event.targetOption].sizes[selectedSize] != 'unavailable' ? selectedSize : product[event.targetOption].defaultSize
                   })">
             <ul class="p0 m1">
               <li>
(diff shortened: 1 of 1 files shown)


src/50_Samples_%26_Templates/Product.html
```diff
@@ -516,7 +516,7 @@ <h2 class="m1">Apple</h2>
                   on="select:AMP.setState({
                     selectedColor: event.targetOption,
                     selectedSlide: 0,
-                    selectedSize: product[event.targetOption].sizes[selectedSize] ? selectedSize : product[event.targetOption].defaultSize
+                    selectedSize: product[event.targetOption].sizes[selectedSize] != 'unavailable' ? selectedSize : product[event.targetOption].defaultSize
```
sebastianbenz: Is it possible to get rid of the `unavailable` state? What would happened if we'd include only the available sizes in the state?

kul3r4: The reason is there is to improve readability.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -544,7 +544,7 @@ <h2 class="m1">Apple</h2>
         <div class="items-center flex">
           <label for="size">Size:</label>
           <amp-selector name="size" layout="container"
-                  on="select:AMP.setState({selectedSize: event.targetOption})"
+                  on="select:AMP.setState({selectedSize: event.targetOption, selectedColor: selectedColor ? selectedColor: 'green'})"
```
choumx: Nit: Space before `:` in ternary operator.

PR #747, opened 2017-04-14
Fixed two small bugs in the product sample

kmh287:
- Fixed expressions for CSS classes on carousel and left-side images to prevent them from disappearing on bind's first apply.
- Fixed bug that prevented the image carousel from responding to `slideChange` events. (The event was renamed from `goToSlide` somewhat recently.)

/to @kul3r4

PR #749, opened 2017-04-17
Update Product.html

kul3r4:
This adds an additional note about default variables. It also updates the documentation.
@kmh287 I added this after #747

### DIFF
--- a/src/50_Samples_%26_Templates/Product.html
+++ b/src/50_Samples_%26_Templates/Product.html
@@ -339,7 +339,7 @@ a.item {
 
     See below for more informations on how we use`amp-bind`. -->
   <div class="product-gallery">
-    <ul class="show" [class]="selectedColor == 'green' ? 'show' : 'hide'">
+    <ul class="show" [class]="(!selectedColor || selectedColor == 'green') ? 'show' : 'hide'">
       <li tabindex="0" role="button">
         <amp-img on="tap:AMP.setState({selectedSlide: 0})"
                   src="/img/green_apple_1_60x40.jpg"
@@ -395,9 +395,9 @@ a.item {
             layout="responsive"
             type="slides"
             [slide]="selectedSlide"
-            on="goToSlide:AMP.setState({selectedSlide: event.index})"
+            on="slideChange:AMP.setState({selectedSlide: event.index})"
             class="show"
-            [class]="selectedColor == 'green' ? 'show' : 'hide'">
+            [class]="(!selectedColor || selectedColor == 'green') ? 'show' : 'hide'">
       <amp-img src="/img/green_apple_1_1024x682.jpg"
                 width="1024" height="682"
                 layout="responsive">
@@ -413,7 +413,7 @@ a.item {
             layout="responsive"
             type="slides"
             [slide]="selectedSlide"
-            on="goToSlide:AMP.setState({selectedSlide: event.index})"
+            on="slideChange:AMP.setState({selectedSlide: event.index})"
             class="hide"
             [class]="selectedColor == 'golden' ? 'show' : 'hide'">
         <amp-img src="/img/golden_apple1_1024x682.jpg"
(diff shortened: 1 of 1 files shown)


src/50_Samples_%26_Templates/Product.html
```diff
@@ -387,7 +387,9 @@ <h2 class="m1">Apple</h2>
    <!-- ## Image Gallery -->
   <!-- The `amp-carousel` component works very well for product image galleries. Learn how the `amp-carousel` component works [here](/advanced/image_galleries_with_amp-carousel/).
 
-  Notice how we are binding to the `slide` and `class` by using to the values of the variables `selectedSlide` and `selectedColor`. -->
+  Notice how we are binding to the `slide` and `class` by using to the values of the variables `selectedSlide` and `selectedColor`. 
+  It's not possible to set a default value for variables, for example we use the expression`(!selectedColor || selectedColor == 'green')`
```
kmh287: This isn't quite right as `amp-state` elements can be used to define initial state. The difference is that all variables defined in an `amp-state` need to be prefixed by the `amp-state` element's id. It's not possible to specify initial values for variables at the top level.

kul3r4: Ok, so if `selectedColor` was part of amp-state and it had `green` as value, does it mean we would not need to check `<ul class="show" [class]="(!selectedColor || selectedColor == 'green') ? 'show' : 'hide'"`

kul3r4: I trieI tried to update the code this way:
`on="tap:AMP.setState({product.selectedSlide: 0})"`
and I get:

> cdn.ampproject.org/v0/amp-bind-0.1.js:22 amp-bind: AMP.setState() failed with error: Parse error on line 1:
> {product.selectedSlide: 0}
> --------^
> Expecting ':', got '.'​​​

Is it supported to set variables from the json state in the `setState`? I remember this was not happening at the start and this is why I was using global variables.

kmh287: Sorry, I misspoke earlier. In order to (re-) set the variable in a `setState` call you'll need to do.

`AMP.setState({product: {selectedSlide : 0}})`

Which will allow you to set `product.selectedSlide` to 0.

kul3r4: Ok,
I tried to rewrite it this way:
`          <amp-selector name="color"
                  layout="container"
                  [selected]="product.selectedColor"
                  on="select:AMP.setState({
                    {product: {
                      selectedColor: event.targetOption,
                      selectedSlide: 0,
                      selectedSize: product[event.targetOption].sizes[selectedSize] != 'unavailable' ? selectedSize : product[event.targetOption].defaultSize
                    }
                  })">`
but still I get:

> error.js:128 Invalid action definition in  
<amp-selector name=​"color" layout=​"container" [selected]=​"product.selectedColor" on=​"select:​AMP.setState({
                    {product:​ {
                      selectedColor:​ event.targetOption,
                      selectedSlide:​ 0,
                      selectedSize:​ product[event.targetOption]​.sizes[selectedSize]​ != 'unavailable' ? selectedSize :​ product[event.targetOption]​.defaultSize
                    }​
                  }​)​" class=​"i-amphtml-element i-amphtml-layout-container i-amphtml-layout i-amphtml-error" role=​"listbox" selected=​"green">​…​</amp-selector>​
 : [ select:AMP.setState({
                    {product: {
                      selectedColor: event.targetOption,
                      selectedSlide: 0,
                      selectedSize: product[event.targetOption].sizes[selectedSize] != 'unavailable' ? selectedSize : product[event.targetOption].defaultSize
                    }
                  }) ] ; unexpected token []

kmh287: Looks like there's an extra curly bracket at the beginning of the argument to `setState`

So it should be:
```js
on=​"select:​AMP.setState({
  product:​ {
    selectedColor:​ event.targetOption,
    selectedSlide:​ 0,
    selectedSize:​ product[event.targetOption]​.sizes[selectedSize]​ != 'unavailable' ? selectedSize :​ 
    product[event.targetOption]​.defaultSize
  }​
}​
```

kul3r4:
sorry i closed this in error

Issue #751, opened 2017-04-18
Style menu on Product page


kul3r4:
Go to [product page](https://ampbyexample.com/samples_templates/product/preview/) and click demo. Then click on the menu on the left hand side, it's not currently styled with ampstart.

PR #752, opened 2017-04-18
Fixes menu

kul3r4:
Solves #751 
@sebastianbenz PTAL

### DIFF
--- a/src/50_Samples_%26_Templates/Product.html
+++ b/src/50_Samples_%26_Templates/Product.html
@@ -260,13 +260,13 @@ a.item {
       <!-- ## Navigation -->
       <!-- Use the [amp-sidebar](/components/amp-sidebar/) component to give customers the chance to quickly jump to any of your product categories.-->
       <amp-sidebar id="drawermenu" layout="nodisplay">
-        <div class="topheader">
-          <a href="/" class="home">Products by Example</a>
+        <div class="topheader m1">
+          <a href="/" class="caps text-decoration-none">Products by Example</a>
         </div>
         <hr/>
-        <h4 class="category expanded"><a href="/samples_templates/product_listing/preview/">Fruit</a></h4>
-        <h4 class="category"><a href="/samples_templates/product_listing/preview/">Vegetable</a></h4>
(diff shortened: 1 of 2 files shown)


src/50_Samples_%26_Templates/Product.html
```diff
@@ -260,13 +260,13 @@
       <!-- ## Navigation -->
       <!-- Use the [amp-sidebar](/components/amp-sidebar/) component to give customers the chance to quickly jump to any of your product categories.-->
       <amp-sidebar id="drawermenu" layout="nodisplay">
-        <div class="topheader">
-          <a href="/" class="home">Products by Example</a>
+        <div class="topheader m1">
```
sebastianbenz: is topheader still needed? I think we can get rid of the div and add "block p1" to the link instead.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -260,13 +260,13 @@
...
+          <a href="/" class="caps text-decoration-none">Products by Example</a>
         </div>
         <hr/>
-        <h4 class="category expanded"><a href="/samples_templates/product_listing/preview/">Fruit</a></h4>
-        <h4 class="category"><a href="/samples_templates/product_listing/preview/">Vegetable</a></h4>
-        <h4 class="category"><a href="/samples_templates/product_listing/preview/">More</a></h4>
+        <h4><a class="m1 caps text-decoration-none" href="/samples_templates/product_listing/preview/">Fruit</a></h4>
```
sebastianbenz: we could get rid of the `<h4>` tag and use the class `h4` instead

kul3r4: This will make all the links in the same line.
Would it be ok to add a <br/> after each one?

sebastianbenz: Same as above, add `class"block p1"`

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
The header has default styling, but each link is styled as `h4`. Suggestion: use h4 for heading and no default styling for links. 

Can you make the same changes to the product listing sample?

kul3r4:
Applied the latest changes, thanks!

PR #753, opened 2017-04-18
Product amp bind update

kul3r4:
Related to #749 
@kmh287 sorry I closed the other issue and I can't open it again.
I have implemented the changes here.
I still wonder if there is a way to avoid the checking at the load of the page:
`(!product.selectedColor || product.selectedColor == 'green') `


src/50_Samples_%26_Templates/Product.html
```diff
@@ -335,69 +335,71 @@ <h2 class="m1">Apple</h2>
     corresponding to the selected colour. We realized the color selection by using `amp-selector`. Find more the doc and info in The Product Page section below.
 
     The green colour is the default color for the gallery when the page loads.
-    Clicking on a preview image will reveal the image in the gallery via the tap action `on="tap:product-carousel-preview.goToSlide(index=0)"`.
+    Clicking on a preview image will reveal the image in the gallery via the tap action `on="tap:AMP.setState({selectedSlide: 0})"`.
 
     See below for more informations on how we use`amp-bind`. -->
   <div class="product-gallery">
-    <ul class="show" [class]="(!selectedColor || selectedColor == 'green') ? 'show' : 'hide'">
+    <ul class="show" [class]="(!product.selectedColor || product.selectedColor == 'green') ? 'show' : 'hide'">
```
kmh287: Now that `product.selectedColor` is initialized in an `amp-state` component, you shouldn't need to do the null check anymore. :smile:

src/50_Samples_%26_Templates/Product.html
```diff
@@ -335,69 +335,71 @@ <h2 class="m1">Apple</h2>
...
+                 [class]="product.selectedSlide == 1 ? 'selected' : '' ">
         </amp-img>
       </li>
     </ul>
   </div>
    <!-- ## Image Gallery -->
   <!-- The `amp-carousel` component works very well for product image galleries. Learn how the `amp-carousel` component works [here](/advanced/image_galleries_with_amp-carousel/).
 
-  Notice how we are binding to the `slide` and `class` by using to the values of the variables `selectedSlide` and `selectedColor`. -->
+  Notice how we are binding to the `slide` and `class` by using to the values of the variables `selectedSlide` and `selectedColor`.
```
kmh287: I think this paragraph needs to be changed to discuss the new approach of using an `amp-state` to initialize these variables.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -335,69 +335,71 @@ <h2 class="m1">Apple</h2>
...
             width="1024" height="682"
             layout="responsive"
             type="slides"
-            [slide]="selectedSlide"
-            on="slideChange:AMP.setState({selectedSlide: event.index})"
+            [slide]="product.selectedSlide"
+            on="slideChange:AMP.setState({ product:​ {selectedSlide: event.index}})"
             class="show"
-            [class]="(!selectedColor || selectedColor == 'green') ? 'show' : 'hide'">
+            [class]="(!product.selectedColor || product.selectedColor == 'green') ? 'show' : 'hide'">
```
kmh287: No more null check needed :smile:

src/50_Samples_%26_Templates/Product.html
```diff
@@ -491,7 +495,7 @@ <h2 class="m1">Apple</h2>
     Every time you select a color, we call `AMP.setState` with `{selectedColor: event.targetOption}`. This is triggering
     all the attributes and elements binded to `selectedColor` to update: the price (binded with the attribute `[text]`), the gallery, the image preview and the size selector (all binded with `[class]`). -->
       <p class="price-description">Price:
-        <span [text]="product[selectedColor || 'green'].sizes[selectedSize || 'S']">$1.99</span>
+        <span [text]="product[product.selectedColor || 'green'].sizes[product.selectedSize || 'S']">$1.99</span>
```
kmh287: This OR may not be necessary anymore either.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -544,21 +550,21 @@ <h2 class="m1">Apple</h2>
         <div class="items-center flex">
           <label for="size">Size:</label>
           <amp-selector name="size" layout="container"
-                  on="select:AMP.setState({selectedSize: event.targetOption, selectedColor: selectedColor ? selectedColor : 'green'})"
-                  [selected]="(product[selectedColor || 'green'].sizes[selectedSize || 'S'] != 'unavailable')
-                      ? selectedSize
-                      : product[selectedColor || 'green'].defaultSize">
+                  on="select:AMP.setState({ product: {selectedSize: event.targetOption, selectedColor: selectedColor ? selectedColor : 'green'}})"
```
kmh287: Is this the size selector? If so, we shouldn't need to change the color here 
right?

src/50_Samples_%26_Templates/Product.html
```diff
@@ -544,21 +550,21 @@ <h2 class="m1">Apple</h2>
...
+                  [selected]="(product[product.selectedColor || 'green'].sizes[product.selectedSize || 'S'] != 'unavailable')
```
kmh287: Shouldn't need the null check for the color, but you'll still need the one for the size.

src/50_Samples_%26_Templates/Product.html
```diff
@@ -544,21 +550,21 @@ <h2 class="m1">Apple</h2>
...
+                      ? product.selectedSize
+                      : product[product.selectedColor || 'green'].defaultSize">
```
kmh287: Null check for color no longer necessary.

src/50_Samples_%26_Templates/Product.html
kul3r4: Is the amp-state accessed on load of the page or after the first interaction with amp-bind?

kmh287: You should be able to access it from the very first time bind evaluates these expressions.

kul3r4: yes, but the sample depends on the value of variables when the page loads. For example this code:
`<p class="price-description">Price:
        <span [text]="product[selectedColor || 'green'].sizes[selectedSize || 'S']">$1.99</span>
    </p>`
note that it will launch an error when the page loads if we don't check for null.

kmh287: Bind expressions don't evaluate until the user takes some action to change bind's state. 

With that particular snippet, I think you want 

```
<p class="price-description">Price: <span [text]="product[product.selectedColor].sizes[product.selectedSize]">$1.99</span> </p>
```

Since both of those vars live on the new `amp-state` element.

kul3r4: Thanks! I will do the updates and push the new version shortly.

kul3r4:
@kmh287 PTAL

src/50_Samples_%26_Templates/Product.html
```diff
@@ -329,51 +329,51 @@ <h2 class="m1">Apple</h2>
     corresponding to the selected colour. We realized the color selection by using `amp-selector`. Find more the doc and info in The Product Page section below.
 
     The green colour is the default color for the gallery when the page loads.
-    Clicking on a preview image will reveal the image in the gallery via the tap action `on="tap:product-carousel-preview.goToSlide(index=0)"`.
+    Clicking on a preview image will reveal the image in the gallery via the tap action `on="tap:AMP.setState({selectedSlide: 0})"`.
```
kmh287: `AMP.setState({product: {selectedSlide: 0}})`

src/50_Samples_%26_Templates/Product.html
```diff
@@ -441,7 +441,9 @@ <h2 class="m1">Apple</h2>
       for each product when you switch between colours and the size you previously selected is not available for the new color -->
   <amp-state id="product">
     <script type="application/json">
-    {
+    { "selectedColor": "green",
```
kmh287: Up to you, but might want to put the opening curly brace on its own line

src/50_Samples_%26_Templates/Product.html
```diff
@@ -475,17 +477,16 @@ <h2 class="m1">Apple</h2>
       depending on the selected colour.
 
     We bind the `text` attribute to the
-      value of expression `product[selectedColor].price`. `product` is the id `amp-state` json described in the Product Configuration section.
+      value of expression `product[product.selectedColor].sizes[product.selectedSize]`. `product` is the id `amp-state` json described in the Product Configuration section.
 
-    Notice how we set a default value for the price: values defined inside `amp-state`
-      will not be accessed at the page load. We set `green` and `S` as default color and default size
-      as the correspondent variables have `null` value at default.
+    Notice how we set $1.99 as default value for the price:  expressions
```
kmh287: "the default value" ?

kul3r4: 1.99 is the default value for the price before the expression is evaluated -> the value that will be shown at load time.
I can remove "value" and let it be "Notice how we set $1.99 as default for the price", would it make clearer?

kmh287: Sorry, I meant there should probably be a "the" between "as" and "default". Completely up to you. I think you should keep "value".

kul3r4: Right, thanks! I changed it

kul3r4:
@kmh287 thank you very much for the review!

PR #774, opened 2017-04-28
Clean up dynamic css classes sample

sebastianbenz:
* Remove social share samples a they were a bit inconsist. I think the
sample works well without them.
* Format the text snippets
* fix viewer docs

### DIFF
--- a/src/20_Components/amp-dynamic-css-classes.html
+++ b/src/20_Components/amp-dynamic-css-classes.html
@@ -10,7 +10,6 @@
     Include the amp-dynamic-css-classes extended component.
   -->
   <script async custom-element="amp-dynamic-css-classes" src="https://cdn.ampproject.org/v0/amp-dynamic-css-classes-0.1.js"></script>
-  <script async custom-element="amp-social-share" src="https://cdn.ampproject.org/v0/amp-social-share-0.1.js"></script>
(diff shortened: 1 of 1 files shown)


sebastianbenz:
@jmadler PTA

jmadler:
The intent of using amp-social-share was to demonstrate more complex usage of these boolean flags rather than just conditional display.  I think the reordering is a helpful way to demonstrate that, and this change doesn't seem to replace it with something else that equally facilitates that lesson.  I'll defer to you though.  If you want to move forward with that change, I'd suggest either using the if-linkedin class, or deleting it.  Good catch on that one btw :)

As for the other change (span tag to p tag), the formatting changes might be helpful but I'd recommend switching the p tags back to span tags and using classes to achieve the formatting you're looking for.  P tags are block-level with a flow content model, and the content of these tags aren't discrete paragraphs, but are intra-paragraph level phrasing content.

I guess it could be argued that these are indeed individual paragraphs.  If that's the case, I'd suggest using div tags instead of p tags.  I tend to avoid P tags wherever possible as they allow developers to omit the closing tag, which I find to be unexpected behavior that leads to bugs, especially for novice developers.  I consider this a particularly important issue for developing samples as the target audience often copy/pastes, and that can increase the likelihood of this class of bugs being introduced.

jmadler:
(viewer docs lgtm)

PR #779, opened 2017-04-28
providing resource for alt attribute guidance

garcialo:
linking to resource for determining alt attribute
also modified alt attributes used in sample code to something more appropriate for the example images


sebastianbenz:
Fantastic! Thanks!

PR #787, opened 2017-05-02
Update amp-usernotication style

### LINKED ISSUE #739: use AMP Start design for notification
image

there is an AMP Start one that @camelburrito implemented but is not in use on AMP Start. that one matches the site design so it looks more intentional.

sebastianbenz:
* uses ampstart style
* fixes #739


sebastianbenz:
screen shot 2017-05-02 at 1 26 16 pm

src/30_Advanced/amp-user-notification_with_Server_Endpoint.html
```diff
@@ -21,12 +21,12 @@
...
   <style amp-custom>
-    footer {
-      margin-bottom: 50px;
-    }
-    amp-user-notification {
-      padding: 8px;
-      background: beige;
+    /* Center the notification content */
+    amp-user-notification > div {
+      padding: 1rem;
```
kul3r4: could we use p1 instead of padding: 1rem;?

sebastianbenz: thought about the same - in this case we have all relevant styling in once place. We need some clear rules on when to use basscss and when not.

kul3r4: ok! we can follow this convention when we create styling classes.

PR #791, opened 2017-05-02
improve amp-mustache sample preview

sebastianbenz:
* use fixed-height layout as content doesn't change height
* add padding around preview
* show a preview for the `not` sample


sebastianbenz:
@jmadler sorry for not spotting these in the review. I should've run the sample locally.

jmadler:
LGTM

PR #794, opened 2017-05-04
Improve tabbed

kul3r4:
@sebastianbenz Applied style from amp-start. I think it looks nicer now.


src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -632,23 +628,23 @@ <h2 class="m1">Apple</h2>
     Learn how to implement tabs in AMP [here](/advanced/tab_panels_with_amp-selector/preview/). -->
   <amp-selector role="tablist"
   layout="container"
-  class="ampTabContainer p1">
+  class="ampTabContainer ampstart-headerbar-nav">
   <div role="tab"
-    class="tabButton h4 p1"
+    class="tabButton h4 www-component-anchor ampstart-nav-item"
```
sebastianbenz: Hm - not sure if we should use non-ampstart classes such as `www-component-anchor` in a sample. Wdyt?

kul3r4:
@sebastianbenz I removed that class.

PR #823, opened 2017-05-17
Ads with amp-bind to fallback to x-origin instead of FIE

aghassemi:


aghassemi:
created https://github.com/ampproject/amphtml/issues/9395 for `amp-bind` to fix it

PR #824, opened 2017-05-17
replace amp-ad runtime in source view

sebastianbenz:
Fixes all links opening in a new tab.

I'm not proud, but it works for now.


sebastianbenz:
@aghassemi PTAL

sebastianbenz:
@aghassemi good point - done

PR #827, opened 2017-05-18
Fixed canonical URL. Added viewport comment.

tomayac:
The canonical URL needs to include a trailing slash, as this is where [/introduction/hello_world/](https://ampbyexample.com/introduction/hello_world/) actually lives.
Additionally, the viewport declaration [is currently not showing](https://ampbyexample.com/introduction/hello_world/), presumably because the comment was missing.


tomayac:
@sebastianbenz I don't see this live yet, has it been deployed?

kul3r4:
@tomayac I just deployed the latest version, could you check it again?
Thanks!

tomayac:
Thumbs up 👍 Thanks for the fix and deploy!

PR #858, opened 2017-06-15
Remove experiment flags for amp-bind

choumx:
Now that amp-bind is launched. Also revert changes in #823.

/to @sebastianbenz @kul3r4

PR #859, opened 2017-06-16
Remove experiment from amp-bind

ericlindley-g:
/to @kul3r4 
/cc @choumx

### DIFF
--- a/src/20_Components/amp-bind.html
+++ b/src/20_Components/amp-bind.html
@@ -1,7 +1,3 @@
-<!---{
-    "experiments": ["amp-bind"],
-    "preview": "default"
-}--->
 <!--
   ## Introduction
 


kul3r4:
@ericlindley-g thanks for this! there is #858 already opened for this, however both #858 and #859 are failing the CI. There may be an update needed for the validator, do you know if the AMP validator has already been updated?
@choumx FYI

ericlindley-g:
Ah, sorry about that! Will close this issue. As for the validator, it should already be updated and working, since pages with amp-bind already appeared in production for origin trials. Though it's suspicious that the checks for the other PR are failing now...

choumx:
Looks like a new validation rule FYI. It was a quick fix.

Issue #829, opened 2017-05-19
server-side sorting/filtering using binding to amp-list[src]


ericlindley-g:

kul3r4:
I could see this in:
- [AMP for ecommerce page](https://ampbyexample.com/amp_for_e-commerce%3A_getting_started/)
- [amp-list example](https://ampbyexample.com/components/amp-list/)
other ideas?

ericlindley-g:
Agreed — I think it's a perfect fit for the product category/browse/search-result page for the e-comm sample. For now amp-carousel doesn't support amp-list, so it might help if the items were shown more as a list or grid, rather than a set of carousels. WDYT?

[Here's an example](https://www.bhphotovideo.com/c/search?Ns=p_PRICE_2%7c0&ci=16158&N=4288586281) in the wild of what server-side sorting/filtering w/out page reload can look like for a product category/browse/search-result page.

Would also be nice to call out the general functionality in the amp-bind and amp-list pages (though depending on the amount of effort, you could either include an example inline, or just link to another example)


The general feature would also be useful for a blog or news publication category/search-results page, so it might be worth making an example for that as well at some point

kul3r4:
@choumx I am working on a "sorting menu" like [this](https://www.bhphotovideo.com/c/search?setNs=p_OVER_ALL_RATE%7c1&Ns=p_OVER_ALL_RATE%7c1&ci=16158&srtclk=sort&N=4288586281): Once you click on a sort choice from the sorting list, that choice goes on top of the list.

I was trying to realize it in [here](https://jsbin.com/bixureyoju/edit?html,console,output) by using amp-bind to change the value for each option using a temp value to do the swap. It doesn't seem to work, could you take a look? If you click on the menu choice (the one with the border) it shows the other option. 
hight to low, low to high -> low to high, high to low 
low to high, high to low -> hight to low, high to low  both are "high to low"

Do you think there could be an easiest way to achieve this?

kul3r4:
@choumx Apology, amp-selector would be a better fit here, please ignore my previous comment.

ericlindley-g:
The example works great! 

To follow up on my request above: Is it possible to present the sort items in a list view, instead of in the carousel? I think it's great to show that changing the sort order is possible in the context of the carousel, but I think that presenting items in as a list, the format that users are used to seeing (and a format that merchants are used to using), will increase the likelihood that they will understand how this will be useful to them. What do you think?

kul3r4:
[...]

PR #862, opened 2017-06-20
Add sort to product page

kul3r4:
Solves #829


kul3r4:
@sebastianbenz 
I managed to remove the rendering of the products server side and do that with amp-list/amp-bind.
This also required to change how the search works and given the search is used in the product sample as well, for now I duplicated the endpoints to live the product page as it is.

If you are ok with the approach then I will move the Product Page to the same code (html and backend)

Find a preview here: [https://amp-by-example-staging.appspot.com/samples_templates/product_browse_page/preview/](https://amp-by-example-staging.appspot.com/samples_templates/product_browse_page/preview/).

If you also try on mobile, the first time you load the page the show-more button appears, then after a search, even if the content of amp-list changes, the show more button stay, this is likely a bug and I will follow up separately.

kul3r4:
@sebastianbenz I added the support for keeping the search when sorting. PTAL

kul3r4:
@sebastianbenz I changed the layout of the page to be a list instead of a carousel. PTAL

sebastianbenz:
Few comments:

* could we add a reset button instead of an `all` list item?
* can you make the product list show up in a grid as before? This looks weird on desktop with a single column
* can you move the sorting input into the filter bar?
* suggestion for sorting input: change label to 'sort by' and values to 'price lowest', 'price highest'

kul3r4:
Hi,
thanks for reviewing, my answers in line:
could we add a reset button instead of an all list item?
I am adding it for the color, yes
can you make the product list show up in a grid as before? This looks weird on desktop with a single column
This was a suggestion from @ericlindley-g, you can find the discussion in #829 
can you move the sorting input into the filter bar?
will do
suggestion for sorting input: change label to 'sort by' and values to 'price lowest', 'price highest'

sebastianbenz:
Hm. I'd still prefer the grid view. I think the sorting will still be obvious (we could add more products to make this more explicit). Mobil will still be a single column.

kul3r4:
@sebastianbenz you can take a look at existing websites like:
[ebay](https://www.ebay.co.uk/sch/i.html?_from=R40&_trksid=p2380057.m570.l1313.TR0.TRC0.H0.Xcamera.TRS0&_nkw=camera&_sacat=0)
[bh](https://www.bhphotovideo.com/c/search?Ns=p_PRICE_2%7c0&ci=16158&N=4288586281)
there you can find the pattern of using a list for the category page

sebastianbenz:
Yes, but then the product list items need to be responsive. The whitespace
too the right is too ugly imho.

kul3r4:
I agree, we need to find some the content to add to the right

kul3r4:
@sebastianbenz @ericlindley-g I think it would be nice to recap where we are with this PR to decide next steps and see if we can break down additional features:
[staging](https://amp-by-example-staging.appspot.com/samples_templates/product_browse_page/preview/) contains a version where: 
- sorting server side is implemented
- desktop view has color filters on the left and mobile has color filter in a ligthbox (this is using amp-list and amp-bind together as well)

Outstanding problems:
- amp-list requires a fixed height and we are using amp-list for showing results . As the content of the amp-list changes based on the query, if the query returns a small result (like only one product) then the space used by the amp-list is too high but there is no way of resizing
- should we agree on what is the best design for the list of products? I think list is a cleaner way of showing results in comparison to the grid, this is also supported by existing samples I pointed above
- if we keep the list layout, I would need to add something on the right of the product to not leave the blank space on desktop. I suggest an additional column with some product info, WDYT?
- at the moment it's possible to reset the color filter but not the search (you would need to reload). Any recommendation here?

Thanks you very much for your support and review, this sam

ericlindley-g:
Sorting looks amazing! It's great to see that this all comes together so well

w/r/t grid v. list, I think we could make either work if we make some changes to the design.

Grid view: needs more products to be comprehensible: when it's just a single row of items it looks more like a "featured products" area. I'm not sure this will be enough to make the interaction understandable, though. Also might help to put the items on "cards", but it's hard to be sure.

List view: This felt much more "clear" to me, and also makes sorting more straightforward, because the sort order is only top to bottom (as opposed to the case in rows, where the sort is left to right, and top to bottom). However, it needs a visual design that fills up more of the horizontal space to look OK on desktop. This could mean adding more information, or spreading the existing information across a wider card.

In the end, I recommended the list view because that seemed like the most straightforward way to make this interaction clear to the user, but if others feel strongly, I could see either interaction being clear with enough design thought.

/cc @spacedino for any additional thoughts

sebastianbenz:
I'm happy with both grid or list, the only thing we need to fix for the list is the horizontal space. Grid seems like a more common implementation. We need to add more products anyway if we want to demo pagination as well...

ericlindley-g:
Sounds good — I'm hoping to include a GIF of this interaction in the amp-bind blog post coming soon. What do you think about merging this version for now so we can direct people to it in the short term? Then we can settle on either grid or list (or both, if we want a toggle between them like [this page](https://www.usnews.com/best-colleges/rankings/national-universities?_mode=list) on desktop)

Another option is adding more products and switching back to grid, but I'm not 100% sure that will fix the clarity issue of the interaction.

sebastianbenz:
+1 for adding more products and switching back to grid -> on mobile you still get a single column you can use for your GIF.

@kul3r4  might be worth checking out [this article](http://heydonworks.com/practical_aria_examples/#toolbar-widget) to see if we can do anything to improve accessibility of the sorting/filter buttons.

kul3r4:
Just deployed on [staging](https://amp-by-example-staging.appspot.com/samples_templates/product_browse_page/preview/) a new version with 2 more items and put back the grid for desktop. @sebastianbenz WDYT?

sebastianbenz:
Looks great on desktop! Seems like there are still a few quirks on mobile:

1. Some results don't render correctly in the list

screen shot 2017-07-05 at 11 24 43 am

2. Can you move the close button in the filter lightbox to the bottom?

screen shot 2017-07-05 at 11 25 20 am

3. Suggestion: on mobile can you put the filter and sort input in the same line?

kul3r4:
@sebastianbenz Thanks! just fixed those and deployed to staging.

src/50_Samples_%26_Templates/Product_Browse_Page.html
```diff
@@ -1,218 +1,415 @@
...
-        }
-    }
-  </script>
-</amp-analytics>
-</body>
+         </script>
+      </amp-state>
+      <amp-lightbox id="filter-lightbox" layout="nodisplay">
+         <div>
+            <h3 class="ml1">Filter results</h3>
```
sebastianbenz: s/ml1/m1 mt2/

sebastianbenz:
great! one small style nit. 

Is it possible to close the Lightbox when you select a filter option?

kul3r4:
@sebastianbenz I removed the close button and now it closes when selecting a color.

sebastianbenz:
@kul3r4 sorry for not being precise. We should leave the close button as well -> in case someone doesn't want to select a filter.

kul3r4:
@sebastianbenz you are right, I added it back

sebastianbenz:
Great!

PR #863, opened 2017-06-21
Add amp-access-laterpay

trodrigues:


src/20_Components/amp-access-laterpay.html
```diff
@@ -0,0 +1,73 @@
...
+  <!-- This section is visible to all users. No special markup is needed in this case.-->
+  <section class="preview">
+    <h3>The Article</h3>
+    <p>
+      This is a preview of the article.
+    </p>
+  </section>
+
+  <!-- This section is only shown if the value of the `access` and `error` attributes 
+  on the response for the authorization requests both evaluate to `FALSE`-->
```
sebastianbenz: Maybe add that the later pay dialog will be shown here. Are the `id` and `class` predefined?

trodrigues:
Do you mean having it work on the Playground?

Whenever I try to open anything on the Playground on the local version I get a "could not fetch the given url" pop up and I see a 404 to http://localhost:8000/playground/fetch

Would I need to follow [these instructions](https://github.com/ampproject/amp-by-example#running-the-backend-server) to get this to work?

src/20_Components/amp-access-laterpay.html
trodrigues: Ah, yes, good point. Yes, the id and class are predefined. I guess I should also add a note about that?

sebastianbenz:
The playground doesn't work locally. However, I was expecting the sample to show a preview of the login button/dialog,.

trodrigues:
Oh, I see what you mean. I hadn't noticed the component was actually instantiated and running on that page.

The thing is, the dialog only gets rendered after getting the configuration back from the server with the data for the pricepoints. This means that our server, and the associated merchant account needs to have a setup with the merchant domain. Because it has no setup for `localhost:8000` the request fails.

Once this is merged and running on ampbyexample.com we could set a configuration for that domain and a configuration for the page of the example as an "article" and then it would be able to render and work properly.

sebastianbenz:
@trodrigues that be fantastic if you could setup a configuration for ampbyexample.com. If you want to test it first, you could create a configuration for https://amp-by-example-staging.appspot.com as well.

trodrigues:
Ok, so I've tested it locally and set it up to be used with our sandbox environment (which is the same we tell our customers to use when implementing our solutions on their websites and doesn't require an actual payment).

### Testing it locally

If you want to test it while running it locally, add:
```
"configUrl": "https://connector.dev.laterpaytest.net"
```
inside the "laterpay" object in the configuration. So it would look like:
```
  <script id="amp-access" type="application/json">
  {
    "vendor": "laterpay",
    "laterpay": {
      "articleTitleSelector": ".preview > h3",
      "configUrl": "https://connector.dev.laterpaytest.net"
    }
  }
  </script>
```

You'll also need to run it with the AMP development mode by adding `#development=1` in the URL and run the local ampbyexample environment on port 8081 (our server does a domain check and we currently have it set it up on our DEV environment for `localhost:8081`).

### The sandbox setup

Like I mentioned before, our Sandbox environment is setup to accept requests from ampbyexample.com.

Right now, we also require the development parameter for AMP to be true, so it would be https://ampbyexample.com/components/amp-access-laterpay/#development=1

I'll remove this requirement with a PR to our amp-access-laterpay extension. I added this in the past because I wanted users to be sure that they were running the Sandbox environment but we also make that clear on the payment workflow so I think that's unnecessary.

sebastianbenz:
...and it's live: https://ampbyexample.com/components/amp-access-laterpay/#development=1

trodrigues:
Great, thanks!

PR #871, opened 2017-06-26
Embeds: make previews configurable

sebastianbenz:


sebastianbenz:
@bpaduch can you test whether this fixes the amp-list problem?

bpaduch:
The preview's height is definitely shorter, but I expected that what I see in the [preview](https://bp-samples.firebaseapp.com/examples/amplist.basic.preview.html) file itself would be the same as what displays in the [embed](https://bp-samples.firebaseapp.com/examples/amplist.basic.embed.html) file:
preview-v-embed.

I also expected the "Preview" tab to be the active tab, but it's not. Even though, in the template, I specified `active-tab=preview`. This is what I specified in the [amplist.basic.template.html](https://bp-samples.firebaseapp.com/examples/amplist.basic.template.html):

```html
<amp-iframe height="75"
            layout="fixed-height"
            sandbox="allow-scripts allow-forms allow-same-origin"
            resizable
            src="https://bp-samples.firebaseapp.com/examples/amplist.basic.embed.html?active-tab=preview&preview-height=75">
  <div overflow tabindex="0" role="button" aria-label="Show more">Show full code</div>
  <div placeholder></div>
</amp-iframe>
```

Also, because the template file is generated from the abe.generatePreview(config) process, can I pass the "active tab" and "preview height" in the config?  Otherwise, the generated template file will be overwritten each time we run the gulp task to generate embeds.

sebastianbenz:
@bpaduch how did you test this? I didn't deploy this to npm to avoid breaking the doc generation process. (pinged you via hangout regarding this on Monday, but you were OOO - sorry for not updating the issue).

I've uploaded this as a beta version, you can install via: `npm install amp-by-example@beta`

bpaduch:
Hmm, I guess I just tested with the old one. Retesting.

sebastianbenz:
@bpaduch do you automatically embed the generated templates into the docs? I thought you'd be copy pasting them into the markdown docs. In that case it should be ok to manually adjust the config parameters.

bpaduch:
Oh, you're correct.  I do manually add them to the docs.  All's good.

sebastianbenz:
Does the amp-list overflow work? Have you got a link to the sample?

PR #873, opened 2017-06-29
Fix playground 404 error

### LINKED ISSUE #857: ABE Playground returns 404 in the preview frame
On opening the [ABE Playground](https://ampbyexample.com/playground/), it shows "404 page not found" text in the preview window and it's impossible to preview the result of live editor.

By opening the page in incognito mode, it works.

The error message in the console in DevTools is the following:
```
(index):16807 GET https://ampbyexample.com/playground/playground/preview/ 404 ()
(anonymous) @ (index):16807
```

The exact line that causes error is:

```
previewIframe.src = PREVIEW_REQUEST_URL;
```

and the error message pop-up is saying

```
Failed to load resource: the server responded with a status of 404 ()
```

sebastianbenz:
* Write source directly to the iframe instead of using a SW.
* Makes editing smoother by only rendering the preview if the browser is
idle.
* Store scroll position after update.

Fixes #857


sebastianbenz:
Sorry - saw too late that you already approved this one. Added another fix. Can you please take another look. Thanks!

sebastianbenz:
Thanks for the review!

PR #875, opened 2017-06-30
Fix product

### LINKED ISSUE #874: Fix Product Page errors
When viewing the Product Page with #development=1, many errors are shown such as:

- width/height may not appear in div
- amp-img is missing role and tabindex
- default value of class does not match first expression result
- Clicking on a 2nd image in carousel errors 'Invalid slide value', although the slide still transitions.

https://ampbyexample.com/samples_templates/product_page/preview/#development=1

kul3r4:
Fixes #874 
@sebastianbenz PTAL

FYI I introduced 3 different variables for selectedSlide state as the 3 carousels have different number of slides. When there is only one (golden carousel), the code was failing when setting selectedSlide=1 from another carousel.
This is because of [new checks](https://github.com/ampproject/amphtml/blob/fdc7ec22ce08ca8ea1b53cbc6d46135c5ff9557a/extensions/amp-carousel/0.1/slidescroll.js#L504)

@choumx FYI


kul3r4:
Removed, thanks!

PR #879, opened 2017-07-03
Fix: analytics not working for playground

sebastianbenz:


sebastianbenz:
Thanks for the review!

PR #881, opened 2017-07-03
Improve load performance

andreban:
- Adds link rel="preload" to the `validator.js` Javascript.
- Adds async to the `validator.js` script.

Since the JS file is a bit large and the script tag is in the end
of the file this will make the request for the validator to start
as soon as the head of the file is received. This should 
improve TTI (Time to Interactive).

It seems Polymer makes the UI visible after domInteractive is
fired. Making validator.js async allows it to fire before it's
downloaded and executed. This should improve FMI (
First Meaningful Paint).

### DIFF
--- a/playground/src/elements/webui-mainpage.html
+++ b/playground/src/elements/webui-mainpage.html
@@ -129,61 +129,63 @@
           // Show preview 
           preview.refresh(editor.getEditorValue());
 
-          // Validate - result is of type amp.validator.ValidationResult.
-          // See validator.proto for the provided fields.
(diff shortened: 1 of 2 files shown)


sebastianbenz:
Thanks! Not sure though if this is enough. We need to delay access to the validator until it's been loaded: https://github.com/ampproject/amp-by-example/blob/master/playground/src/elements/webui-mainpage.html#L134 as well.

sebastianbenz:
Closing this as it will become redundant after the playground re-design.

sebastianbenz:
...but thanks for the suggestion!

Issue #610, opened 2017-02-08
Request: Include transition in CSS bind example


ericlindley-g:
Currently I believe that the amp-bind example demo's how the binding protocol can change the CSS class on interaction. This is great, but I think it would be good to show that CSS transitions are possible between states (if that is the case :)

I'm not sure what the best way to show this is, but it could be as simple as a block that changes color and moves from one side of the page to the other, over a duration of ≥ 1 second

WDYT?

/cc @kul3r4 @choumx @sebastianbenz

sebastianbenz:
We're already doing this in the product sample. The selector button state changes are animated.

ericlindley-g:
I see now: looks like just the box-shadow is animated? It's great that it's demo'ing that transitions are possible, but it's pretty subtle right now.

This isn't high-priority, but:
If we just look at the product page, I think a transition might help smooth out transitions where the entire carousel + thumbnails is swapped out when selecting different color options. A fade in/out between .1s and .3s might be nice for the entire carousel+thumbnails, to drive home the point that the transitions can be smooth.

Eventually, if we find a good use-case, it might be cool to have an interaction that is entirely based on amp-bind + CSS class changes.

sebastianbenz:
Sounds good.

@kul3r4 can you take a look once you're done with the product sample?

PR #892, opened 2017-07-06
Add fadein

kul3r4:
Solves #610 by adding a fade in effect

### DIFF
--- a/src/50_Samples_%26_Templates/Product_Page.html
+++ b/src/50_Samples_%26_Templates/Product_Page.html
@@ -153,6 +153,17 @@ This sample showcases how to build a product page in AMP HTML.
     form#order amp-selector[name=color] div:not([selected]):not([disabled]) {
       opacity: 0.6;
     }
+    .animated {
+      animation-duration: 1s;
+      animation-fill-mode: both;
+    }
+    @keyframes fadeIn {
+      0% {opacity: 0;}
+      100% {opacity: 1;}
+    }
+    .fadeIn {
+      animation-name: fadeIn;
+    }
     .hide {
       display: none;
     }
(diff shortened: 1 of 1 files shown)


src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -153,6 +153,17 @@
     form#order amp-selector[name=color] div:not([selected]):not([disabled]) {
       opacity: 0.6;
     }
+    .animated {
```
sebastianbenz: I'd merge this class into the fadeIn class as these parameters are usually specific to an animation.

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -403,16 +414,16 @@ <h2 class="m1">Apple</h2>
         <amp-img on="tap:AMP.setState({product: {selectedSlideForGreen: 0}})"
                   src="/img/green_apple_1_60x40.jpg"
                   width="60" height="40"
-                  class="selected"
-                  [class]="product.selectedSlideForGreen == 0 ? 'selected' : '' "
+                  class="selected animated fadeIn"
```
sebastianbenz: I wouldn't fade the elements in initially (this doesn't work for carousels anyway).

sebastianbenz:
Looks great!

I wouldn't animate the thumbnails though as they already have a transition (elevation). The problem is that it's currently not visible because of the outline:

screen shot 2017-07-06 at 8 28 39 pm

Can you remove the outline (and the fadeIn) from these?

kul3r4:
@sebastianbenz 

- removed the fade in from the preview
- removed the outline
- merged the css class for animation

PTAL

sebastianbenz:
Can you merge `fadeIn` into the `show` class? Or are we using show anywhere without fade?

kul3r4:
We use it for the preview as well, so I can't merge those two.

PR #896, opened 2017-07-14
enable cors for server rendered samples

sebastianbenz:


sebastianbenz:
I need this to be able to download sample source code into the playground

PR #897, opened 2017-07-14
amp-url-converter: assume http

sebastianbenz:
...otherwise 'example.com' gives a different result than
'http://example.com'.

### DIFF
--- a/api/scripts/amp_url_converter.js
+++ b/api/scripts/amp_url_converter.js
@@ -38,7 +38,7 @@ class AmpUrlConverter {
       this.showError('Empty input');
       return;
     }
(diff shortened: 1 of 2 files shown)


api/scripts/amp_url_converter.js
```diff
@@ -38,6 +38,9 @@ class AmpUrlConverter {
       this.showError('Empty input');
       return;
     }
+    if (!/^http[s]?\:\/\//.test(urlString)) {
```
cramforce: `/i` for case insensitive?

sebastianbenz: good point - done

PR #904, opened 2017-07-20
Fix upload filters

sebastianbenz:
This will remove the dreaded "Too many files to watch" warning + uploads should be faster.


sebastianbenz:
@kul3r4 friendly ping

PR #917, opened 2017-08-02
Fix shopping cart

### LINKED ISSUE #914: Product Page - Add to cart is broken
Looking at https://ampbyexample.com/samples_templates/product_page/preview/ 
`Add to card` is supposed to redirect the user however it looks like the XHR to `https://ampbyexample.com/samples_templates/product_page/add_to_cart?__amp_source_origin=https%3A%2F%2Fampbyexample.com` does not return a JSON response and therefore we get `Failed to parse response JSON: SyntaxError: Unexpected token < in JSON at position 0​​​` error` and user is not redirected.

kul3r4:
Fixes #914


sebastianbenz:
Thanks a lot for fixing!

PR #929, opened 2017-08-30
Disable "show more" on product preview after use

ithinkihaveacat:
Disables the "show more" button on https://ampbyexample.com/samples_templates/product_page/preview/ after it's used (since it doesn't do anything after first use).

/cc @kul3r4


kul3r4:
Does it give any warning when running locally related to the fact that the variable is not set initially?

ithinkihaveacat:
@kul3r4 No, nothing at all without `#development=1`. With `#development=1` there are some warnings but I don't think they're related:

screenshot from 2017-08-30 13 28 15

kul3r4:
Thanks!

PR #935, opened 2017-09-07
Add a second load more button - potentially more

kul3r4:
We are receiving lots of requests how to implement something similar to infinite scrolling which I know it's not available at the moment, so I am trying to demonstrate what is possible at the moment with amp-bind and amp-list beside the current sample which shows a "show more" just once.

I came up with this approach which uses amp-bind to dinamically change the src name by incrementing a number, but I am not sure how to hide or disable the button when there are no items anymore, can this be an event on `amp-list` where, when there is an error, I could set something on AMP state? @choumx do you have any ideas how to better implement this?

@ericlindley-g FYI
@ithinkihaveacat who added the logic to disable the button when we only had one


aghassemi:
@kul3r4 Can think of couple of workaround that might work:
1- Button is inside an amp-form that does an xhr to see if there are more items and uses the result of that XHR to decide to hide itself.
2- Every template in `amp-list` contains the button but CSS is used to hide all but the one in the last `li`. Now if backend decided there is no more items left, it can return a `nomore` flag in the json which can also be used in the template to hide the button in the last item.

kul3r4:
@aghassemi Thanks a lot! I'll give it a try with the option 2, I just moved the button inside the template and looks nice, need to work on the backend now.

kul3r4:
@sebastianbenz PTAL

sebastianbenz:
I'm not sure I would recommend this approach. Usually a Load More button requests more items, but in this case it simply returns longer lists, so it doesn't really solve the problem. At least that's my understanding of a load more functionality, if it's purely an UX issue then this approach is fine...

kul3r4:
Agree, this is the only way I know to use `amp-list` to implement a load-more pattern. 
A different approach would be having different calls corresponding different `amp-list` elements and then having the page published server side; in this scenario the page would contain as many `amp-list` as needed.
@ericlindley-g have you thought about this scenario?

sebastianbenz:
I think Ali's first suggestion could be a possible approach.

sebastianbenz:
e.g. if there is no more content available the form could return a 404 which then hides the load more button.

kul3r4:
@sebastianbenz how does this change the fact that the amp-list returns longer lists?

sebastianbenz:
amp-list supports objects as src, you could append the form result to amp-state and render the state in the list (not 100% sure if that works).

kul3r4:
@sebastianbenz Added a second amp-state to hold the items, PTAL

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -751,10 +767,22 @@ <h4 class="p2">Products that have been bought together</h4>
...
+          myState: { items: myState.items.concat(event.response.items)},
+          product: {  showMoreButtonDisabled: true }
+        }); submit-success: AMP.setState({
+          myState: { items: myState.items.concat(event.response.items)},
+          product: {timesShowMoreClicked: product.timesShowMoreClicked + 1 }
+        });">
+      <input type="hidden" name="timesShowMoreClicked" value="0" [value]="product.timesShowMoreClicked.toString()">
+      <input type="submit" value="Show more"
+        class="ampstart-btn caps m1 mb3 show"
+        [class] = "(product.showMoreButtonDisabled == true ? 'ampstart-btn caps m1 mb3 hide' : 'ampstart-btn caps m1 mb3 show')">
```
sebastianbenz: Do you need `ampstart-btn caps m1 mb3 ` when the button is hidden?

src/json/more_related_products0.json
```diff
@@ -0,0 +1,30 @@
+{
+  "items": [{
+    "id": 7,
+    "img": "/img/product1_640x426.jpg",
+    "name": "Apple",
```
sebastianbenz: Can you give the items different names? e.g. Premium Apple

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -751,10 +767,22 @@ <h4 class="p2">Products that have been bought together</h4>
         </a>
       </template>
     </amp-list>
-    <!--
-    The button `tap` action sets the value of the `relatedProductsSrc` to an extended list of related products.
-    -->
-    <button [disabled]="listSrcButtonDisabled" class="ampstart-btn caps m1 mb3" id="listSrcButton" on="tap:AMP.setState({listSrcButtonDisabled: true, product: {relatedProductsSrc: '/json/more_related_products.json'}})">Show more</button>
+    <form method="GET"
```
sebastianbenz: Can you document the ampstate and the amp form?

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -550,6 +557,8 @@ <h2 class="m1">Apple</h2>
       "selectedSlideForRed": 0,
       "selectedSlideForGolden": 0,
       "selectedSlideForGreen": 0,
+      "timesShowMoreClicked": 0,
```
sebastianbenz: suggestion: rename `timesShowMoreClicked` to `moreItemsPageIndex`

src/json/more_related_products0.json
```diff
@@ -0,0 +1,30 @@
+{
```
sebastianbenz: nit: rename this file to more_related_products_page0 (and the others accordingly)

backend/product-browse.go
```diff
@@ -253,6 +256,40 @@ func handleSearchRequest(w http.ResponseWriter, r *http.Request, page Page) {
 	http.Redirect(w, r, route, http.StatusSeeOther)
 }
 
+func handleLastLoadMoreRequest(w http.ResponseWriter, r *http.Request) {
+	EnableCors(w, r)
+	SetContentTypeJson(w)
+	timesShowMoreClicked := r.URL.Query().Get("timesShowMoreClicked")
+	if timesShowMoreClicked == "1" {
```
sebastianbenz: also rename to moreItemsPageIndex

backend/product-browse.go
```diff
@@ -253,6 +256,40 @@ func handleSearchRequest(w http.ResponseWriter, r *http.Request, page Page) {
...
+		w.WriteHeader(http.StatusNotFound)
+	}
+	productsFile, err := ioutil.ReadFile(buildShowMorePath(timesShowMoreClicked))
+	if err != nil {
+		w.WriteHeader(http.StatusNotFound)
+	}
+	var productsRoot JsonRoot
+	err = json.Unmarshal(productsFile, &productsRoot)
```
sebastianbenz: Why have a marshall/unmarshall step? You could simply write the file contents to the response.

kul3r4: Removed!

backend/product-browse.go
kul3r4: done

src/json/more_related_products0.json
kul3r4: done

src/json/more_related_products0.json
kul3r4: done

src/50_Samples_%26_Templates/Product_Page.html
kul3r4: removed

src/50_Samples_%26_Templates/Product_Page.html
kul3r4: done

src/50_Samples_%26_Templates/Product_Page.html
kul3r4: done

kul3r4:
@sebastianbenz Can we leave the loading indicator for a second iteration?

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -752,9 +760,33 @@ <h4 class="p2">Products that have been bought together</h4>
...
+    <form method="GET"
+        action="/json/more_related_products_page"
+        action-xhr="/json/more_related_products_page"
+        target="_top"
+        on="submit-error: AMP.setState({
+          myState: { items: myState.items.concat(event.response.items)},
+          product: {  showMoreButtonDisabled: true }
+        }); submit-success: AMP.setState({
+          myState: { items: myState.items.concat(event.response.items)},
+          product: {moreItemsPageIndex: product.moreItemsPageIndex + 1 }
```
sebastianbenz: just an idea: it is possible to include a boolean var `hasMorePages` in the result? We could use this to hide the show more button.

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -752,9 +760,33 @@ <h4 class="p2">Products that have been bought together</h4>
...
+        });">
+      <input type="hidden" name="moreItemsPageIndex" value="0" [value]="product.moreItemsPageIndex.toString()">
```
sebastianbenz: do we need the `.toString()` here?

sebastianbenz:
Np - to add the loading later

src/50_Samples_%26_Templates/Product_Page.html
kul3r4: Added, thanks

backend/product-browse.go
```diff
@@ -83,6 +85,7 @@ func InitProductBrowse() {
 	RegisterSample("samples_templates/product_page", renderProduct)
 	RegisterSampleEndpoint("samples_templates/product_browse_page", SEARCH, handleSearchRequest)
 	http.HandleFunc("/samples_templates/products", handleProductsRequest)
+	http.HandleFunc(SHOW_MORE_PATH, handleLastLoadMoreRequest)
```
sebastianbenz: nit: s/handleLastLoadMoreRequest/handleLoadMoreRequest

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -285,11 +286,17 @@
...
       justify-content: flex-start;
     }
+
     .close-gallery-button {
       z-index: 1;
       position: absolute;
       right: 0;
     }
+
+    .listSrcButton {
```
sebastianbenz: do we still need this rule?

src/50_Samples_%26_Templates/Product_Page.html
```diff
@@ -752,9 +760,31 @@ <h4 class="p2">Products that have been bought together</h4>
       </template>
     </amp-list>
     <!--
-    The button `tap` action sets the value of the `relatedProductsSrc` to an extended list of related products.
+    While infinite scrolling is not possible at the moment in AMP, you can use `amp-bind` to dynamically change the
+    src of `amp-list` and add more items to the page.
+    We bind the `src` attribute to the id of an `amp-state` component so that the `amp-list` will use the items
+    from that component as a `src`. We append items to the `amp-state` everytime the user click on the Show more button, see below for more info.
```
sebastianbenz: Can you add a sentence describing the trick you've been using to init the state with the same content as the amp-list.

PR #951, opened 2017-10-02
Add style to Polymer version on main page and detail page

kul3r4:
@sebastianbenz This adds: 

- Style to the main page localhost:8080/polymer. Please note that since I needed to change the css and it only works with absolute URLs, in order to run it locally you need to copy inside the `style` of `sample-home` the added css (which was moved from the template `index.html`)

- Anchor support

- I am trying a different approach from how AMPByExample is at the moment on the main page. Instead of displaying all the content (categories and samples description), I would like to try and experiment showing less in the main page (only the categories) and having an extra page for the category which shows the samples for the category. The reason for this is that I feel like there is too much content in the home page and too much scrolling to do. 
Let me know what you think, this is also the reason I haven't styled the list page yet, just in case you say that is not doable at all.


sebastianbenz:
Why doesn't the css work? Would a separate subdomain, e.g. polymer.ampyexample.com help?

kul3r4:
It's explained in the [bug](https://github.com/Polymer/polymer-cli/issues/892), it's not related where the css is hosted, but more how the build checks for css files only inside the polymer folder.

polymer/index.html
```diff
@@ -10,11 +10,15 @@
     <!-- See https://goo.gl/OOhYW5 -->
     <link rel="manifest" href="/manifest.json">
 
-    <script src="/polymer/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
-
-    <link rel="import" href="/polymer/src/abe-polymer-app.html">
+    <link rel="stylesheet" type="text/css" href="http://localhost:8080/css/example.css">
```
sebastianbenz: localhost urls?

polymer/src/sample-detail.html
```diff
@@ -96,15 +98,6 @@
...
 
-    <!--
-      shop-network-warning shows a warning message when the items can't be rendered due
-      to network conditions.
-    -->
-    <shop-network-warning
-        hidden$="[[!failure]]"
-        offline="[[offline]]"
-        on-try-reconnect="_tryReconnect"></shop-network-warning>
-
```
sebastianbenz: why did you remove the offline warning?

polymer/src/sample-category-data.html
```diff
@@ -26,6 +26,11 @@
   <script>
   (function() {
 
+    function sanitizeString(string) {
```
sebastianbenz: What does this method do?

polymer/src/sample-detail.html
kul3r4: Because I am not adding that component at the moment, for simplicity.

polymer/src/sample-category-data.html
kul3r4: It sanitizes strings like this: ""<p>Get started with AMP and learn how to build your first AMP page.</p>\n""

polymer/index.html
kul3r4: Sorry, that was a leftover

kul3r4:
@sebastianbenz available on [staging](https://amp-by-example-staging.appspot.com/polymer/)

sebastianbenz:
I like the new navigation. In the case of the home screen you could then remove the no longer necessary sidebar.

Few things I've noticed looking at the sample:

* Make the whole card clickable and remove the action
* page load feels quite slow (too slow)
* sometimes pages flicker as if they're loaded multiple times
* category pages are un-styled
* scroll position is not reset
* console displays quite a few 404s
* white frame around content

kul3r4:
Hi, I don't think the scroll position should be reset.
From [https://medium.com/@owencm/designing-great-uis-for-progressive-web-apps-dd38c1d20f7#.ofgfr058y](https://medium.com/@owencm/designing-great-uis-for-progressive-web-apps-dd38c1d20f7#.ofgfr058y):

> tapping back should bring back to the list page with the same scroll position they were at before they pressed on the item.

kul3r4:
@sebastianbenz PTAL, new version is on [staging](https://amp-by-example-staging.appspot.com/polymer)

kul3r4:
@sebastianbenz PTAL

kul3r4:
@sebastianbenz:
center cards on desktop -> done
make it possible to navigate from detail pages to category and home page -> I added the header on top. For now, the ads section will point to the original one, not the polymer one.
Can you transition between pages, e.g. fade? -> I added a fade effect on the list of cards

kul3r4:
@sebastianbenz ping

sebastianbenz:
Is it live on staging?

kul3r4:
Yes, and I just noticed that I missed the menu for mobile (it only appears for desktop)

sebastianbenz:
Noticed a few issues with the demo:

* does not work in desktop safari 
* cards are not correctly aligned on desktop
* pink card content (after moving back and forth)
* cards stopped working after a few navigations, see console log
* cards fade in, but sample pages don't (can you show a loading indicator while the page is loading? Otherwise there's only a blank page visible)
* few errors on initial page load in console

screen shot 2017-11-06 at 9 13 36 am
screen shot 2017-11-06 at 9 13 42 am
screen shot 2017-11-06 at 9 22 26 am
screen shot 2017-11-06 at 9 14 12 am
screen shot 2017-11-06 at 9 14 26 am

kul3r4:
@sebastianbenz Latest version is on [staging](https://amp-by-example-staging.appspot.com/polymer)

PR #956, opened 2017-10-03
Update amp-sidebar.html - remove experiment on toolbar

camelburrito:
Update amp-sidebar.html - remove experiment on toolbar

### DIFF
--- a/src/20_Components/amp-sidebar.html
+++ b/src/20_Components/amp-sidebar.html
@@ -1,6 +1,5 @@
 <!---
   {
-    "experiments": ["amp-sidebar toolbar"],
     "preview": "default"
   }
 --->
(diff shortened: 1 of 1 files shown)


sebastianbenz:
The sample currently doesn't validate as the sidebars are not a direct child of body. A quick workaround would be to disable validation, which I'm not too happy about. The problem is the dummy divs around the sidebar to put them together into a code snippet. A better way would be to get rid of them by splitting the doc section into multiple parts, e.g.:

```
<!-- 
   ...
  -->
      <amp-sidebar id="sidebar-right" layout="nodisplay" side="right">
        ....
      </amp-sidebar>

      <!-- The sidebar trigger: -->
      <button on="tap:sidebar-right.toggle" class="ampstart-btn caps m2">Toggle sidebar</button>

      <!-- The toolbar target: -->
      <div id="target-element-right">
      </div>
```

sebastianbenz:
Closed as fixed in other PR

Issue #679, opened 2017-03-04
iFrame needs title attribute


weboverhauls:
Framed content (iframe & frame elements) require a title attribute to describe the content. "This provides a label for the frame so users [of assistive technology] can determine which frame to enter and explore in detail."

WCAG 2.4.1 Bypass Blocks: A mechanism is available to bypass blocks of content that are repeated on multiple Web pages. (Level A)

There are 5 instances on the amp-iframe page
https://ampbyexample.com/components/amp-iframe/

Code example:
`<iframe class="i-amphtml-fill-content -amp-fill-content" name="amp_iframe0" allowfullscreen="" frameborder="0" sandbox="allow-scripts allow-same-origin allow-popups" src="https://player.vimeo.com/video/140261016#amp=1"></iframe>`

Suggested code:
`<iframe title="Netflix Branding: The Stack" class="i-amphtml-fill-content -amp-fill-content" name="amp_iframe0" allowfullscreen="" frameborder="0" sandbox="allow-scripts allow-same-origin allow-popups" src="https://player.vimeo.com/video/140261016#amp=1"></iframe>`

References:
https://www.w3.org/TR/2016/NOTE-UNDERSTANDING-WCAG20-20161007/navigation-mechanisms-skip.html
https://www.w3.org/TR/WCAG20-TECHS/H64.html

lisaychuang:
Question - looking at the [amp-iframe page](https://ampbyexample.com/components/amp-iframe/) and [Github repo code](https://github.com/ampproject/amp-by-example/blob/master/src/20_Components/amp-iframe.html),  is this issue regarding adding a title attribute to the examples?

For example:

```
# current code
<amp-iframe width="500" height="281"
                          layout="responsive"
                          sandbox="allow-scripts allow-same-origin allow-popups"
                          allowfullscreen
                          frameborder="0"
                          src="https://player.vimeo.com/video/140261016">
  </amp-iframe>
```

```
# Suggested change
<amp-iframe width="500" height="281"
                          title="Netflix Branding: The Stack"
                          layout="responsive"
                          sandbox="allow-scripts allow-same-origin allow-popups"
                          allowfullscreen
                          frameborder="0"
                          src="https://player.vimeo.com/video/140261016">
  </amp-iframe>
```

[...]

PR #964, opened 2017-10-04
Added amp-iframe title attribute & indentation

lisaychuang:
PR addresses issue #679  - updating examples in code base:

1. Added `title` attribute to `<amp-iframe>` example to describe its content.
> Video of Sintel, an independently produced short film

[W3C H64: Using the title attribute of the frame and iframe elements:](https://www.w3.org/TR/WCAG20-TECHS/H64.html)
"title attribute of the frame or iframe element to describe the contents of each frame. This provides a label for the frame so users can determine which frame to enter and explore in detail. It does not label the individual page (frame) or inline frame (iframe) in the frameset."

[WCAG 2.4.1 Bypass Blocks:](https://www.w3.org/TR/2016/NOTE-UNDERSTANDING-WCAG20-20161007/navigation-mechanisms-skip.html) A mechanism is available to bypass blocks of content that are repeated on multiple Web pages. (Level A)

2. Modified indentation for amp components

### DIFF
--- a/src/30_Advanced/Integrating_Videos_in_AMP_an_Overview.html
+++ b/src/30_Advanced/Integrating_Videos_in_AMP_an_Overview.html
@@ -38,7 +38,9 @@
     Use `amp-video` for hosted videos.
     Find more examples [here](/components/amp-video/).
   -->
-  <amp-video width="480" height="270" src="video/tokyo.mp4"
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Thanks for also fixing indentation

PR #971, opened 2017-10-04
fix typo and whitespaces

sebastianbenz:


sebastianbenz:
@morsssss I've copied your changes to a new PR. PTAL

morsssss:
Thanks, Sebastian!

Scanning through this, I happened to notice that "getting" in line 92 should be "get".

Do I have a new career now as a proofreader? 😛

sebastianbenz:
😀 - we definitely could use a professional proofreader

Thanks!

PR #991, opened 2017-10-26
Add linked dropdowns example

cvializ:
example

### DIFF
--- a/src/30_Advanced/Linked_Dropdowns.html
+++ b/src/30_Advanced/Linked_Dropdowns.html
@@ -9,26 +9,35 @@
   depend on the value the user selects in the first dropdown.
   Here we select cities in a chosen country.
 -->
-<!-- -->
+<!-- #### Setup -->
+<!--
+  Import the `amp-bind`, `amp-list` and `amp-mustache` components, and
+  include the standard AMP boilerplate.
+-->
 <!doctype html>
 <html ⚡>
 <head>
-  <meta charset="utf-8">
+  <title>Linked Dropdowns</title>
   <link rel="canonical" href="<%host%>/advanced/linked_dropdowns/">
+
+  <meta charset="utf-8">
   <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
   <script async src="https://cdn.ampproject.org/v0.js"></script>
 
+  <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+  <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+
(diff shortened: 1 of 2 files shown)


src/30_Advanced/Linked_Dropdowns.html
```diff
@@ -0,0 +1,94 @@
...
+  </style>
+</head>
+
+<body>
+
+  <!-- #### Usage -->
+  <!--
+  We use two amp-list elements to render the options within each `<select>`.
+  The first dropdown triggers a request, which is batched with the request
+  triggered by the other `<amp-list>` and  the `<amp-state>` element, so only
```
kul3r4: remove extra space between 'and  the'

src/30_Advanced/Linked_Dropdowns.html
```diff
@@ -0,0 +1,95 @@
...
+      margin: 10px 20px;
+    }
+  </style>
+</head>
+
+<body>
+
+  <!-- #### Usage -->
+  <!--
+  We use two amp-list elements to render the options within each `<select>`.
```
kul3r4: - add quotes for `amp-list`

src/30_Advanced/Linked_Dropdowns.html
```diff
@@ -0,0 +1,95 @@
...
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+
+  <!-- #### Setup -->
+  <!--
+    Import the `amp-bind`, `amp-list` and `amp-mustache` components.
+  -->
+  <div>
```
kul3r4: do we need this `div`?

src/30_Advanced/Linked_Dropdowns.html
```diff
@@ -0,0 +1,95 @@
...
+    <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+    <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+    <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+  </div>
```
kul3r4: I think this `</div>` can be removed

src/30_Advanced/Linked_Dropdowns.html
cvializ: Hmm, I added it so that the 3 scripts would appear together in the example:
screen shot 2017-10-26 at 1 49 33 pm

Without the `<div>`, only the first script tag is shown, instead of all 3. Is there another way to group these? Or do you think it would be better to show the whole `<head>` tag instead of adding these invalid `<div>` tags?

cvializ: I checked and this causes the build to fail, so I think I'll show the full `<head>` instead.

PR #993, opened 2017-10-26
Do not show boilerplate code. Move autosuggest to advanced as per ear…

cvializ:
…lier suggestion

### DIFF
--- a/src/30_Advanced/Autosuggest.html
+++ b/src/30_Advanced/Autosuggest.html
@@ -10,9 +10,6 @@ The example below suggests cities in an address form, but the list of
 suggestions may contain any content, such as anticipated search queries, or
 product listings with images.
 -->
-<!--
-## Setup
--->
 <!-- -->
 <!doctype html>
 <html ⚡>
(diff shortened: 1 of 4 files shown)


src/30_Advanced/Autosuggest.html
```diff
@@ -12,22 +12,25 @@
...
 -->
+<!-- -->
 <!doctype html>
 <html ⚡>
 <head>
   <title>Autosuggest form field example</title>
   <meta charset="utf-8">
   <script async src="https://cdn.ampproject.org/v0.js"></script>
-  <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+  <!-- First we include `amp-form` -->
```
sebastianbenz: If it makes sense it'd be nice to explain why we need the respective components.

src/30_Advanced/Autosuggest.html
```diff
@@ -12,22 +12,25 @@
 -->
 <!--
 ## Setup
-
-First, the markup declares the page as AMP, contains the AMP boilerplate, and
-includes the required AMP extensions.
 -->
+<!-- -->
```
aghassemi: I believe if you move `<-- ## setup -->` inside and just before the `<!-- First we include...` you won't need this extra `<!-- -->` to exclude the boiler plate.

cvializ: Thanks, that's a good suggestion!

PR #998, opened 2017-10-27
Add a paged list example

### LINKED ISSUE #839: sample: paging for results using amp-list[src]
Could be integrated into existing sample—the idea is that a set of results can be paginated without page reload, using amp-list[src]

cvializ:
Fixes #839 

example2

### DIFF
--- a/backend/config.go
+++ b/backend/config.go
@@ -21,5 +21,6 @@ const (
 	AMP_CLIENT_ID_COOKIE = "AMP_ECID_GOOGLE"
 
 	CATEGORY_COMPONENTS       = "components"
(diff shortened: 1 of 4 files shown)


kul3r4:
@cvializ Nice! Would that be possible to use items from an external source and use more real names and descriptions? for example in product-browse.go we read the content of a json file in `handleLoadMoreRequest`. This could be a second iteration if that is not doable in a short time.

cvializ:
Thanks that's an excellent idea, and it would definitely improve the example. I do think that might be better as a second iteration after this ABE sprint. Thanks!

PR #1000, opened 2017-11-01
show favorite count in bubble

sebastianbenz:


sebastianbenz:
This is what it looks like:

favorite

sebastianbenz:
thanks for the instant review!

PR #1001, opened 2017-11-01
Added the combobox sample, in the 30_Advanced section.

diogo-bastos:
This is a sample containing a drop down list combo box. I created this sample as I needed a combo similar to ones in the Chosen or JQuery libraries for my most recent project.

### DIFF
--- a/src/30_Advanced/ComboBox.html
+++ b/src/30_Advanced/ComboBox.html
@@ -5,8 +5,7 @@
   ## Introduction
 
   This sample demonstrastes a classic dropdown combo, with a text field which filters the dropdown elements 
-  according to the user's input. The difference between this example and the auto-suggest example, is that this example
-  has been built 100% client side, for maximum performance and responsiveness.
+  according to the user's input. The options are rendered via amp-list.
 -->
 <!doctype html>
 <html ⚡>
@@ -23,6 +22,8 @@
   -->
   <script async custom-element="amp-selector" src="https://cdn.ampproject.org/v0/amp-selector-0.1.js"></script>
   <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+  <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+  <script async custom-element="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/3.2.1/css/font-awesome.min.css">
   <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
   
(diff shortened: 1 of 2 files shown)


diogo-bastos:
I made the git commit with the wrong email. I added that email to my git account now.

diogo-bastos:
It seems the travis build failed with the error "The attribute 'option' may not appear in tag 'li'". However, the 'option' attribute is used for an 'li' tag in the amp-selector example: [](https://ampbyexample.com/components/amp-selector/). Is there something I am not doing correctly?

diogo-bastos:
That sounds like a good idea that would make the sample a lot more dynamic! I will implement this change this weekend. 

I would not use it for static options though, because having the options already in the DOM is more responsive than waiting for an additional RTT when the user opens the list.

sebastianbenz:
Hm - curious how much of a difference it makes.

diogo-bastos:
Changed the client side code so that the dropdown options are loaded via amp-list with a mustache template. Sebastian, was this what you had in mind? I still need to add the server side file of this example though.

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,177 @@
...
+                </button>
+            </div>
+            <amp-list class="autosuggest-box"
+                        layout="fixed-height"
+                        height="420"
+                        src="/GetDropdownList"
+                        id="autosuggest-list"
+                        hidden>
+                <template type="amp-mustache">
+                    <amp-selector keyboard-select-mode="focus"
```
sebastianbenz: this will have to go outside the amp-list.

sebastianbenz:
Almost. My idea was to still have the options json embedded in the page in an amp-state element (this avoids a server-request):

```
<amp-state id="options">
    {
       items: [
          {"title": "..., ...},
          {"title": "..., ...},
       ]
    }
  </script>
</amp-state>
```

Then you can bind the amp-list `src` to the options object:

```
   <amp-list [src]="options" ...>
```

And update the options object when the input field is focused:

```
<input ... on="tap:AMP.setState({options: options}), ...">
```

You might have to experiment with how to best trigger the amp-list refresh. 

Not sure this works though. ;-)

sebastianbenz:
Actually, it would be nice to have two samples, one demonstrating the local approach and one with a remote JSON list. (the remote one probably first as the markup is simpler). You can put a JSON file into src/json/food-options.json and then use it like this: `<amp-list src="/json/food-options.json" ...>`.

diogo-bastos:
Alright, I will to try to finish the remote sample tonight then. As for the local sample, should I make it in the same html file or in a different one?

sebastianbenz:
Great! Same file would be good I think.

diogo-bastos:
Added the json file and fixed the template for the remote sample.

As for the local sample, I tried adding the amp-list "src" attribute as the id of an amp-state ("options" in this case) but the amp-list component just tried to make a GET request for some remote endpoint "options", instead of binding to the amp-state. Is there something I'm missing?

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+            overflow-y: scroll;
+        }
+    </style>
+</head>
+<body>
+
+  <!-- ## Basic Usage -->
+  <!--
+  This example includes two inputs. One is the input in which the user writes the text and the other is a hidden input, 
+  where the id value of the selected option is stored and could be submited in a form.
```
sebastianbenz: s/submited/submitted

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+<head>
+  <title>ComboBox</title>
+  <meta charset="utf-8">
+  <link rel="canonical" href="<%host%>/..\30_advanced/combobox/">
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <script async src="https://cdn.ampproject.org/v0.js"></script>
+  <!-- ## Setup -->
+  <!--
+  The amp-selector component is used for the dropdown list options. The amp-bind component is used for the filtering of the dropdown options
+  according to the user's input. The fontawesome css icons are imported for the arrow icon.
```
sebastianbenz: Can you document each import separately? Otherwise they won't be visible in the doc view of the sample.

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+  -->
+  <script async custom-element="amp-selector" src="https://cdn.ampproject.org/v0/amp-selector-0.1.js"></script>
+  <script async custom-element="amp-bind" src="https://cdn.ampproject.org/v0/amp-bind-0.1.js"></script>
+  <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/3.2.1/css/font-awesome.min.css">
```
sebastianbenz: Would you mind not using font-awesome here? I find it a bit overkill for a single icon (as it quite impacts loading performance). We usually use material icons, e.g. [this](https://material.io/icons/?utm_source=google&utm_medium=cpc&utm_campaign=1001467%20%7C%20Material.IO%20%7C%20Global%20%7C%20en%20%7C%20Hybrid%20%7C%20Text%20%7C%20BKWS&utm_term=%7Bkeyword%7D#ic_keyboard_arrow_down) and [this](https://material.io/icons/?utm_source=google&utm_medium=cpc&utm_campaign=1001467%20%7C%20Material.IO%20%7C%20Global%20%7C%20en%20%7C%20Hybrid%20%7C%20Text%20%7C%20BKWS&utm_term=%7Bkeyword%7D#ic_keyboard_arrow_up) one. You could either copy them to `src/img/*` or inline them via css (what I'd recommend doing).

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+        .autosuggest-box {
+            box-shadow: 0px 2px 6px rgba(0,0,0,.3);
+            overflow-y: scroll;
+        }
+
+        input.search-box {
+            padding: 5px 8px;
+            font-size: 1em;
+            background-color: #fff;
+            font-family: Arial,"Helvetica Neue",Helvetica,sans-serif;
```
sebastianbenz: Is there a particular reason why you're using a custom font here?

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        .select-option:focus {
+            box-shadow: inset rgb(94, 158, 215) 0px 0px 45px;
+        }
+
+        .select-option:focus {
+            background: #fff;
+        }
+
```
sebastianbenz: nit: remove empty line

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+    </style>
+</head>
+<body>
+
+  <!-- ## Basic Usage -->
+  <!--
+  This example includes two inputs. One is the input in which the user writes the text and the other is a hidden input, 
+  where the id value of the selected option is stored and could be submited in a form.
+
+  The dropdown list's options are rendered server side, instead of using amp-list to maximize responsiveness.
```
sebastianbenz: I think this no longer applies.

src/30_Advanced/ComboBox.html
```diff
@@ -0,0 +1,181 @@
...
+</head>
+<body>
+
+  <!-- ## Basic Usage -->
+  <!--
+  This example includes two inputs. One is the input in which the user writes the text and the other is a hidden input, 
+  where the id value of the selected option is stored and could be submited in a form.
+
+  The dropdown list's options are rendered server side, instead of using amp-list to maximize responsiveness. 
+  The 'option' attribute of the amp-selector's options, contain the id of the element, as well as its title, seperated by a comma.
```
sebastianbenz: s/'option'/`option`

nice trick with storing title and id in the option 👍

sebastianbenz:
Regarding amp-list binding. Yeah, that's expected. Currently, you can't omit the src attribute of an amp-list, so it will always try to make an request. Maybe it's best to leave the server-side rendered until there is a better approach. WDYT?

src/30_Advanced/ComboBox.html
diogo-bastos: Done!

src/30_Advanced/ComboBox.html
diogo-bastos: Alright, I don't mind. Removed the font-awesome import and added material icons instead.

src/30_Advanced/ComboBox.html
diogo-bastos: Done!

src/30_Advanced/ComboBox.html
diogo-bastos: Done!

src/30_Advanced/ComboBox.html
diogo-bastos: Removed.

src/30_Advanced/ComboBox.html
diogo-bastos: Done!

diogo-bastos:
I think so too. I'll come back to this example if I remember some way to do this locally or if another option becomes available in the amp standard.

src/30_Advanced/ComboBox.html
diogo-bastos: Not really, it was just a mistake. I removed this in the latest commit.

diogo-bastos:
Thank you very much for your patience and guidance, Sebastian! See you soon!

PR #1028, opened 2017-12-07
Add speech bubble for comments

kul3r4:
speech


kul3r4:
@sebastianbenz Updated the speech bubble

bubble2

PR #1029, opened 2017-12-07
Update amp-selector sample

sebastianbenz:
Demonstrate how to create a radio button list.

### DIFF
--- a/src/20_Components/amp-selector.html
+++ b/src/20_Components/amp-selector.html
@@ -22,7 +22,7 @@
 amp-selector, button {
   margin: 1rem;
 }
-amp-selector:not([disabled]) [option][selected]:not([disabled]) {
+amp-selector:not([disabled]) amp-img[option][selected]:not([disabled]) {
   outline-color: #b60845;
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-selector.html
```diff
@@ -108,20 +134,17 @@
 
   <!-- ## Forms -->
   <!--
-Inside a form, `amp-selector` behaves like a radio-button/checkbox group and submits the selected value(s) against the name of the amp-selector.
+Inside a form, `amp-selector` behaves like a radio-button/checkbox group and submits the selected value(s) against the name of the amp-selector. Here we've styled an amp-selector to function like a radio list. Using `amp-selector` for this has the advantage that bindings using `amp-bind` are easier to implement.
```
kul3r4: When you say 

> bindings using `amp-bind` are easier to implement

is there already a sample we could link to? Or maybe add a reference to amp-bind, otherwise it can be confusing.

sebastianbenz:
PTAL

PR #1031, opened 2017-12-07
enable validation for scrollbound tutorial

aghassemi:
`<amp-animation>` is now valid outdie of `body`, enabling full validation for this sample.


sebastianbenz:
Thanks for remembering!

PR #1033, opened 2017-12-08
Use hidden in with amp-bind instead of classes

kul3r4:
I made an attempt to improve the Product page code by using [hidden] in elements using amp-bind. I am not sure how to do with default values where I can't use hidden otherwise they will be hidden at default, so I think for those it's still worth using classes.
@sebastianbenz WDYT?

### DIFF
--- a/src/60_Samples_%26_Templates/Product_Page.html
+++ b/src/60_Samples_%26_Templates/Product_Page.html
@@ -165,12 +165,6 @@ This sample showcases how to build a product page in AMP HTML.
       animation-fill-mode: both;
       animation-name: fadeIn;
     }
-    .hide {
-      display: none;
-    }
-    .show {
-      display: block;
-    }
 
     .unavailable {
       position: relative;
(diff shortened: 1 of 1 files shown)


sebastianbenz:
Have you got a sample for ' default values where I can't use hidden'?

sebastianbenz:
I think you can replace it here `<ul class="show" [class]="product.selectedColor == 'green' ? 'show' : 'hide'">` as well.

...and here as well: [class] = "(product.hasMorePages == false ? 'hide' : 'ampstart-btn caps m1 mb3 show')">

kul3r4:
@sebastianbenz you are right it works.
I was thinking that I would not be able to use `[hidden]` without `hidden` but it actually works, so the default, which is green and is visible at load, can just have `[hidden]` without `hidden`.

sebastianbenz:
Does the fadein still work?

kul3r4:
Yes

PR #1119, opened 2018-01-12
Fix scroll to top in playground

### LINKED ISSUE #1107: Scroll-to-top sample does not work in playground
It is because `<header id="top-header">Header</header>` is not included in the sample itself and is rendered by ABE template.

kul3r4:
Fixes #1107

@sebastianbenz Before I was referencing the header added by the template and that will not work in the playground which has a different header. I think this approach is better because it explains better how I am implementing it by using an additional element as a point to scroll to.


sebastianbenz:
Thanks for fixing this

PR #1136, opened 2018-01-23
Amp stories

sebastianbenz:


src/stories/10_Introduction/_Hello_World.html
```diff
@@ -0,0 +1,66 @@
+<!-- 
+  ## Introduction
+
+  How to tell stories with AMP.
```
kul3r4: is there a link to a doc already?

src/stories/10_Introduction/_Hello_World.html
```diff
@@ -0,0 +1,66 @@
...
+        font-size: 2.875em;
+        font-weight: normal;
+        line-height: 1.174;
+        text-transform: uppercase;
+      }
+    </style>
+  </head>
+
+  <body>
+    <!-- ## Writing the Story -->
```
kul3r4: Would that make sense to use a simple AMP tag, like `amp-img` just to say that AMP stories support AMP tags as normal AMP pages?

src/stories/10_Introduction/_Hello_World.html
```diff
@@ -0,0 +1,66 @@
...
+    <!-- This is the story tag -->
+    <amp-story standalone>
+      <!-- This is the cover page -->
+      <amp-story-page id="cover">
+        <amp-story-grid-layer template="vertical">
```
kul3r4: are there more details on the `amp-story-grid-layer` like which other templates it support?

templates/css/example.css
```diff
@@ -95,6 +95,24 @@ code {
 .abe-code-preview {
   background: #f9f7f7;
 }
+.abe-story-preview {
```
kul3r4: does it make sense to have css for stories in a separate file and add it to the page only if it's a story instead for all the pages?

src/stories/10_Introduction/_Hello_World.html
sebastianbenz: They don't amp-story must be the only child of body.

src/stories/10_Introduction/_Hello_World.html
sebastianbenz: This would go into different samples...

templates/css/example.css
sebastianbenz: I wouldn't that this is justified as long as it's only 5 extra lines

src/stories/10_Introduction/_Hello_World.html
kul3r4: ok! thanks for clarifying

src/stories/10_Introduction/_Hello_World.html
kul3r4: ok

sebastianbenz:
Split this into two PRs (first one is #1156)

PR #1139, opened 2018-01-24
Add an animation as placeholder for amp-list

kul3r4:
Add an animation as placeholder for amp-list.
@aghassemi not sure what I am doing wrong but I still see the default placeholder "..." of `amp-list`, is it possible to overwrite it with a custom animation?

### DIFF
--- a/src/20_Components/amp-list.html
+++ b/src/20_Components/amp-list.html
@@ -121,10 +121,10 @@
 
   <!-- ## Using a placeholder animation -->
   <!--
-    You can use a custom placeholder like an animation to improve the user experience while the list is loading. In order to
-    see the placeholder, you can toggle a slow 3g network from the Chrome dev console network panel.
+    You can use a custom placeholder like an animation to improve the user experience while the list is loading.
+    We are using an endpoint which intentionally delays the response by 10 seconds.
   -->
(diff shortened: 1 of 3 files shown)


aghassemi:
@kul3r4 I believe `noloading` attribute should also be added to `amp-list` since loading icon and `placeholder` are mutually exclusive features. (our default loader shows up even if there is a placeholder since placeholders may not themselves be a loading indicator. e.g. in an amp-iframe, placeholder might just be a poster image)

kul3r4:
@aghassemi Thanks that was it!

src/20_Components/amp-list.html
```diff
@@ -68,6 +98,18 @@
       template="amp-template-id" class="m1">
   </amp-list>
 
+  <!-- ## Using a placeholder animation -->
+  <!--
+    You can use a custom placeholder like an animation to improve the user experience while the list is loading.
```
sebastianbenz: Could we show a skeleton instead of a loading indicator? See https://css-tricks.com/building-skeleton-screens-css-custom-properties/

kul3r4:
@sebastianbenz Using an endpoint with delay similar to loading indicator sample.

sebastianbenz:
Is this ready for review?

kul3r4:
@sebastianbenz yes, please review

sebastianbenz:
Two things:

1. can you make the skeleton match the content? Show each item on a card and show a card-like placeholder for each item.
2. the content never loads for me.

kul3r4:
@sebastianbenz PTAL

Issue #1059, opened 2018-01-09
Add sample for amp-date-picker


kul3r4:
This component is now available in canary. [Reference](https://github.com/ampproject/amphtml/issues/6469#issuecomment-356112911)

kul3r4:
I am working on a first sample, published it [here](https://chiara-ampbyexample.appspot.com/components/amp-date-picker/). I had to use `position: static` otherwise the date picker and the clear button were covered by the code section.

@aghassemi I was wondering what is the reason behind having those with `position: absolute` at default? Moreover, why the input with the date (the ones on top of the date picker) is rendered twice?

cvializ:
It looks like the second date rendered is the screen reader message. I think its styles to hide it visually while still being read by the screen reader are not working as expected on ABE. The position absolute allows the date picker calendar to overlay on top of content, instead of pushing content down the page.

kul3r4:
thanks! So is the first problem (screen reader message) related to the default css or is this something I should solve hiding it manually?

Regarding the position absolute, when I tried it it was not overlaying the content and it was hidden behind the code. I agree it would be best overlaying than pushing the content below, but I think the overlaying is not working at the moment.

cvializ:
I am not sure why the extra text is appearing. I think the CSS might be out of sync with the code, which would need to wait for a new release to go out that includes https://github.com/ampproject/amphtml/pull/12672 to fix it

I think the overlay was getting cut off because the preview area is `overflow: hidden`, so it clips anything within it that goes outside, and because the overlay is `position: absolute` it won't change the size of its container.

kul3r4:
@cvializ Thanks! By removing the `overflow: hidden` I can see that now it opens as an overlay.

When using the Range Date picker, is it possible to access the start and end date? I was trying something like [this](https://github.com/ampproject/amp-by-example/blob/date-picker/src/20_Components/amp-date-picker.html#L151) but doesn't work. I was imaging to be able to use `startDate` or `endDate` from the state, but doesn't seem to work.

cvializ:
Ah sorry for the current lack of documentation. `event.dates[0]` is `startDate`, and event.dates[event.dates.length-1] is `endDate`. Maybe it could be a good idea to alias those values with `startDate` and `endDate` properties on the event, like you were expecting.

kul3r4:
[...]

PR #1144, opened 2018-02-01
Fix  clear action for date-picker

kul3r4:
The clear action is currently referring to the same date-picker for 3 different samples. I updated the action using the right id.
@sebastianbenz PTAL

Related to #1059


sebastianbenz:
Stupid question: do we even need the clear action? Maybe add one dedicated sample about how to clear the input.

kul3r4:
@sebastianbenz I added an extra explanation for the clear action in the first sample. I prefer to leave for the others as it makes the samples easier: if we remove it and someone wants to try couple of times, a reload of the page will be needed to clear the selection.

PR #1145, opened 2018-02-01
Added a Geolocation tutorial

jamesshannon:
Based on sbenz's example on glitch, but with more realistic geolocation-based search results, plus use of skeleton placeholder.

NB: Couldn't get the /preview URL to work. The chrome is there, but the actual demo doesn't take up any space.

FYI @sebastianbenz


src/30_Advanced/Geolocation_with_amp-list.html
```diff
@@ -0,0 +1,140 @@
...
+
+  > `16px + 20px + (5 * (100 + 15px)) = 611px`
+
+  The placeholder list items largely use CSS to draw the skeleton elements as
+  described in [CSS-Tricks](https://css-tricks.com/building-skeleton-screens-css-custom-properties/).
+  The skeleton includes a tiny blurred map to add to the realism, and that map
+  is also used while `<amp-img>` loads the actual images to prevent a
+  flash when the placeholder is removed.
+  -->
+  <amp-list class="abe-preview" width="auto" height="611" layout="fixed-height"
```
sebastianbenz: Suggestion: add the 'noloading' attribute to hide the default loading indicator.

src/30_Advanced/Geolocation_with_amp-list.html
```diff
@@ -0,0 +1,140 @@
+<!---{
+    "preview": "default"
+  }--->
+<!--
+  ## Introduction
+
+  While the AMP Caches don't offer support for geolocation, you can enable
+  geo-based features with geolocation logic on your backaend and
```
sebastianbenz: s/backaend/backend

jamesshannon: Done.

src/30_Advanced/Geolocation_with_amp-list.html
jamesshannon: Cool! I wasn't aware of that. (I think it should be automatic if you're using placeholder.) Done.

jamesshannon:
I'm using the same freegeoip API that you used: https://glitch.com/edit/#!/caramel-wolf?path=server.js:26:11 and supplementing it with a second service to provide 'nearby cities'. 

i did find that occasionally your API (freegeoip) would return a lat/long but not a city name, so I defer to the second API's city name in that case.

Did you download and run this code locally? I'm still trying to figure out why preview isn't working:

geolocation_with_amp-list__preview__-_amp_by_example

Any ideas?

sebastianbenz:
Two things:

1. Please another class name than `abe-preview`, e.g. `geo-list`. It conflicts with an internal class. 
2. The list is hidden because the amp-list has no width. Explicitly give it a width and then it works:

```
    .geo-list {
      width: 100vw;
      max-width: 400px;
    }
```

jamesshannon:
Thanks. Done.

FWIW, `abe-preview` was intentional. The main demo *Code* page wraps the preview in that class, so I used it to ensure that my styles didn't target anything other than the preview section. But then the *Demo* page didn't wrap with that class, so I added it directly.

sebastianbenz:
...and it's live: https://ampbyexample.com/advanced/geolocation_with_amp-list/

PR #1151, opened 2018-02-06
Small tweaks for scrollbound animations sample

aghassemi:
- Fixes the carousel overflowing 
- Makes the image window smoother


kul3r4:
thanks!

PR #1152, opened 2018-02-09
Make advanced ad creative valid

zhouyx:
The fake ad impl change has hit prod. So I removed `dev-channel` opt-in.

I couldn't get the lightbox example to work for now. I'll work on adding the 360 and lightbox examples back to experimental ads in following PR. But want to get this in first. 

@sebastianbenz PTAL. Thanks!


src/amp-ads/40_Experimental_Ads/Scrollbound_Slides_360_Ad.html
```diff
@@ -17,9 +17,8 @@
   <title>Scrollbound Slides 360 Ad</title>
   <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
   <script async src="https://amp-ads.firebaseapp.com/dist/amp-inabox.js"></script>
```
sebastianbenz: Does this need to be replaced as well?

src/amp-ads/40_Experimental_Ads/Scrollbound_Slides_360_Ad.html
```diff
@@ -17,9 +17,8 @@
...
-  <script async custom-element="amp-animation" src="https://amp-ads.firebaseapp.com/dist/v0/amp-animation-0.1.max.js"></script>
   <script async custom-element="amp-lightbox" src="https://amp-ads.firebaseapp.com/dist/v0/amp-lightbox-0.1.max.js"></script>
-<script async custom-element="amp-google-vrview-image" src="https://amp-ads.firebaseapp.com/dist/v0/amp-google-vrview-image-0.1.max.js"></script>
+  <script async custom-element="amp-google-vrview-image" src="https://amp-ads.firebaseapp.com/dist/v0/amp-google-vrview-image-0.1.max.js"></script>
```
sebastianbenz: Does this need to be replaced as well?

src/amp-ads/40_Experimental_Ads/Scrollbound_Slides_360_Ad.html
zhouyx: I didn't replace this one and the stack-bind one because they are experimental ads and won't be valid. I didn't check with the the slides 360 ads, but for the stack-bind ad, we still need this patched version of `amp-inabox.js`. So I decided to leave those two examples unchanged. I'll revert all changes to those two examples.

src/amp-ads/40_Experimental_Ads/Scrollbound_Animation_Ad.html
```diff
@@ -0,0 +1,191 @@
...
+      </script>
+    </amp-animation>
+    <!--
+    **Slide cards opposite to scroll direction**:
+    The ad initially shows the top-most card. As the document scrolls, the other two cards are revealed.
+
+    To achieve that, we define 4 keyframes. Between 0% and 10% of viewport, the first card is displayed. Between 10% and 90%, the cards container will vertically translate to -384 pixels, which is the negative of double as each card's height. When reaching this point, the third card will be fully displayed. Between 90% and 100%, the same position is kept.
+    -->
+
+    <!-- ## Scrollbound Animations -->
```
zhouyx: @aghassemi PTAL at how I use `<amp-animation>` and `<amp-position-observer>` here, and the documentations.

src/amp-ads/40_Experimental_Ads/Scrollbound_Video_Ad.html
```diff
@@ -216,6 +208,17 @@
     at the last keyframe.
   -->
 
+  <!-- ## Scrollbound Animations -->
```
zhouyx: @aghassemi PTAL at how I use `<amp-animation>` and `<amp-position-observer>` here, and the documentations.

src/amp-ads/40_Experimental_Ads/Scrollbound_Animation_Ad.html
aghassemi: LGTM. thanks for fixing these!

src/amp-ads/40_Experimental_Ads/Scrollbound_Slides_360_Ad.html
zhouyx: Decided to completely remove the examples first. I'll add it back in following PR after I figure out the experimental ads issue.

zhouyx:
@sebastianbenz PTAL

templates/preview-a4a.html
```diff
@@ -5,15 +5,13 @@
...
       <script async src="https://cdn.ampproject.org/v0.js"></script>
       <script async custom-element="amp-ad" src="https://cdn.ampproject.org/v0/amp-ad-0.1.js"></script>
     {{/metadata.adSlot}}
     {{^metadata.adSlot}}
-      <!-- TODO(aghassemi,#9311): Remove when A4A can be embeded using prod binaries -->
-      <script async src="https://amp-ads.firebaseapp.com/dist/amp.js"></script>
-      <script async custom-element="amp-ad" src="https://amp-ads.firebaseapp.com/dist/v0/amp-ad-0.1.max.js"></script>
-      <script async custom-element="amp-ad-fake" src="https://amp-ads.firebaseapp.com/dist/v0/amp-ad-network-fake-impl-0.1.max.js"></script>
+      <script async src="https://cdn.ampproject.org/v0.js"></script>
+      <script async custom-element="amp-ad" src="https://cdn.ampproject.org/v0/amp-ad-0.1.js"></script>
```
sebastianbenz: Do we then still need the adSlot variable? Seems like both branches now do the same.

zhouyx: I believe not. I'll remove adSlot

zhouyx: I thought about this. Actually we would still need this. Because when we compare the performance between AMP Ads and HTML Ads, we don't want to use fake ad to cheat. We would still need the adSlot variable to request a real doubleclick ad to compare.

zhouyx: Ahhh I see your point. Removed

PR #1161, opened 2018-02-12
Add temp fix to serve lightbox ad

zhouyx:
This is a temporary fix.
Lightbox doesn't work because of amp cors check
The easiest fix would be on the runtime side, to disable the check for fake ad type. 
Another fix would be to add middleware to ampbyexample server, but I figure it's a little bit risky to make the change now.
So I introduced this temporary fix, that should be able to remove after we make the real fix on to AMP runtime. 

@sebastianbenz PTAL and had this change in for the deployment. Thanks!


templates/preview-a4a.html
```diff
@@ -144,16 +151,30 @@ <h1 id="abe-preview-header-title" class="h3 center truncate caps">{{subHeading}}
...
+        {{#metadata.adExp}}
+        <amp-ad width="{{width}}"
+                height="{{height}}"
+                type="fake"
+                data-use-a4a="true"
+                force3p="false"
+                fakesig="true"
+                src="{{a4aEmbedUrl}}"
+                disable3pfallback="false">
+        <div placeholder="" class="amp-hidden"></div>
```
zhouyx: For some reason, the ad got default placeholder on top before timeout. This is a temporary hack to remove that. Since the whole PR is temporary, I think it's OK to live with this at now.

sebastianbenz:
The ABE server supports CORS. Where is it failing?

zhouyx:
Oh that' would be great! Then could it be that I am testing with localhost? 

And to make sure, the ABE server also need to set `AMP-Access-Control-Allow-Source-Origin` to the response header? Is that also supported? How could I have that set? Thanks.

sebastianbenz:
you can test this by running gulp backend:watch locally: http://localhost:8080/amp-ads/experimental_ads/lightbox_ad/preview/

sebastianbenz:
Just tested it, Lightbox ad works. Do we need to update the Lightbox samples to use the new runtime?

sebastianbenz:
Fixed already on master.

PR #1177, opened 2018-02-19
Remove repeated imports from amp-lightbox-gallery

kul3r4:
Those are making the page not valid


sebastianbenz:
Thanks for fixing!

PR #1193, opened 2018-03-02
Fixed bug in second demo with image slider and custom item

darkkain:
Both panel are displaying by default  - it's wrong behavior. Add wrappers for ``` <amp-carousel>```.


darkkain:
I signed it!

darkkain:
@sebastianbenz no problem, glad to be useful ;)

PR #1202, opened 2018-03-20
Adds section for Ken Burns Effect

Enriqe:
Adds a section to amp-story describing how to achieve a "Ken Burns" effect using new animation presets.


src/stories/20_Features/Animations.html
```diff
@@ -212,6 +215,50 @@ <h1>whoosh-in-right</h1>
         </div>
       </amp-story-grid-layer>
     </amp-story-page>
+
+    <!-- ## Ken Burns Effect -->
```
sebastianbenz: You don't need to put every line into a separate comment, something like this should work:

```
<!--- 
## Ken Burns

You can achieve a "Ken Burns" effect by combining the 'pan' and 'zoom' animations. Here's how to do it:

* ...
* ...

-->

File: src/stories/20_Features/Animations.html [hunk 2]
```diff
@@ -212,6 +215,50 @@ <h1>whoosh-in-right</h1>
...
+            class="img-container"
+            animate-in-after="ken-burns-img1">
+            <amp-img id="ken-burns-img2"
+              src="https://picsum.photos/1024/768?image=1026"
+              layout="fixed"
+              width="1024"
+              height="768"
+              animate-in="pan-down"
+              animate-in-duration="30s"
+              animate-in-after="ken-burns-img1"
```
sebastianbenz: `</amp-img>` missing

Enriqe:
> Very cool! Is it supposed to work/validate yet?

As per #14157 it is now going to be validated :)

Moved the effect to a new section like you suggested and added separate samples for the pan and zoom animations. Let me know what you think!

src/stories/20_Features/Visual_Effects.html
```diff
@@ -0,0 +1,106 @@
+<!--
```
sebastianbenz: Sorry for not being clear enough. I meant a new category "Visual Effects" and the sample would be named Ken Burns. You can do this by renaming the file to:

```
src/stories/30_Visual_Effects/Ken_Burns.html
```

Creating a new directory (30_Visual_Effects) will create a new category. You can copy and adjust the index.json file from one of the other directories (it contains the category description).

src/stories/20_Features/Visual_Effects.html
```diff
@@ -0,0 +1,106 @@
...
+    }
+  </style>
+  <title>Visual Effects</title>
+  <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
+</head>
+<body>
+  <amp-story standalone>
+    <!-- 
+    ## Ken Burns Effect
+    Best known for his work on documentaries, Ken Burns developed a signature effect in his work which consisted of a combination of panning and zooming over an image.
```
sebastianbenz: This paragraph would then become the new introduction.

src/stories/20_Features/Visual_Effects.html
```diff
@@ -0,0 +1,106 @@
...
+</head>
+<body>
+  <amp-story standalone>
+    <!-- 
+    ## Ken Burns Effect
+    Best known for his work on documentaries, Ken Burns developed a signature effect in his work which consisted of a combination of panning and zooming over an image. 
+    We can achieve this simple yet popular and captivating effect by combining the 'pan' and 'zoom' animations in `amp-story`. Here's how to do it.
+
+    The general idea is to nest elements to combine the animations: the container element will animate the zoom, and the image inside the container will be animated with a panning.
+    * Start by creating a div which will act as the image container. Set its style class to set `position:absolute`.
```
sebastianbenz: * s/to set/to/
* can you explain why we need to position the image absolute?

src/stories/20_Features/Visual_Effects.html
```diff
@@ -0,0 +1,106 @@
...
+    * Set its animation to be `animate-in="zoom-in"` or `animate-in="zoom-out"` and its duration with `"animate-in-duration=30s"`.
+    > In this demo we use 30 seconds but your mileage may vary. Try experimenting with different values to see what works best for your desired results.
```
sebastianbenz: You can create info boxes like this: 

```
  * Set its animation to be `animate-in="zoom-in"` or `animate-in="zoom-out"` and its duration with `"animate-in-duration=30s"`.
    <div class="ampstart-card info">
    **Tip:** In this demo we use 30 seconds but your mileage may vary. Try experimenting with different values to see what works best for your desired results.
    </div>
  ...
```

src/stories/20_Features/Animations.html
```diff
@@ -212,6 +212,92 @@ <h1>whoosh-in-right</h1>
         </div>
       </amp-story-grid-layer>
     </amp-story-page>
+    
+    <!-- ## Zoom-In -->
```
sebastianbenz: Perfect!  Thanks!

Enriqe:
Updated the demo + samples. They should work now, let me know!

sebastianbenz:
Works on mobile now :-), but no longer on desktop:

screen shot 2018-03-22 at 19 26 21

src/stories/30_Visual_Effects/Ken_Burns.html
```diff
@@ -0,0 +1,143 @@
...
+    <!-- 
+      You can also achieve the effect with multiple images!
+      
+      Just combine the animations as described in the [combining animations section](/stories/features/animations/#combining-animations). Remember all the elements being animated after certain animation must include the `animate-in-after` attribute. Including the children being animated. See below for the code sample.
+          
+      Keep in mind results may vary depending on the image size and animations you use. Experiment with different animation durations and presets to see what you like best!
+    -->
+    <amp-story-page id="ken-burns-effect2">
+        <amp-story-grid-layer template="vertical">
+          <div animate-in="zoom-in" animate-in-duration="30s" class="img-container">
```
sebastianbenz: I'd suggest making the animation a bit shorter (e.g. 5s). Having to wait 30s to see the effect is a bit long.

src/stories/30_Visual_Effects/Ken_Burns.html
```diff
@@ -0,0 +1,143 @@
...
+              width="1600"
+              height="1200">
+            </amp-img>
+        </div>
+      </amp-story-grid-layer>
+    </amp-story-page>
+
+    <!-- ## Using Multiple Images -->
+    <!-- 
+      You can also achieve the effect with multiple images!
```
sebastianbenz: I had to see the demo to understand what was going on. Maybe change the intro to: 

"You can even implement a slideshow transitioning between multiple pages using this effect."

src/stories/30_Visual_Effects/Ken_Burns.html
```diff
@@ -0,0 +1,143 @@
...
+                layout="fixed"
+                width="1600"
+                height="1200"
+                animate-in="pan-down"
+                animate-in-duration="30s"
+                animate-in-after="ken-burns-img1">
+              </amp-img>
+            </div>
+          </div>
+        </amp-story-grid-layer>
```
sebastianbenz: It'd be nice to show 1 or 2 more images - what do you think?

src/stories/30_Visual_Effects/Ken_Burns.html
Enriqe: The problem is that the nature of the effect is that it's a slow panning and zooming. If we use shorter times then the animation will run faster, beating the purpose of the effect. But I guess for the sake of showing the support of multiple images it makes sense.

Enriqe:
> Works on mobile now :-), but no longer on desktop:

How are you testing this? It works on desktop for me locally.

src/stories/30_Visual_Effects/Ken_Burns.html
Enriqe: Sounds good!

sebastianbenz:
On desktop, you'll see a white frame once the vertical resolution is > 1024px.

Enriqe:
Oh I was just able to reproduce it. I'll update you when it's fixed.

sebastianbenz:
Hey Enriqe - any updates on this?

Enriqe:
Hey @sebastianbenz, sorry for the delay. I've just been able to do some adjustments and it seems to work for me in both desktop and mobile. Can you verify that it works on your side as well? Thanks!

Enriqe:
Hey Sebastian, were you able to verify it works on your side for desktop / mobile?

sebastianbenz:
Sorry, the problem still persists on desktop.

Enriqe:
This will be fixed with [a fix in the runtime side (14334)](https://github.com/ampproject/amphtml/pull/14334).

src/stories/20_Features/Animations.html
```diff
@@ -258,57 +258,61 @@ <h1>zoom-out</h1>
...
-        <h1>pan-left</h1>
-        <amp-img animate-in="pan-left"
-                 animate-in-duration="4s"
-                 layout="fixed"
-                 src="https://picsum.photos/1024/768?image=1026"
-                 width="720"
-                 height="320">
-        </amp-img>
+      <amp-story-grid-layer template="fill">
+        <div>
```
sebastianbenz: Why do we need the div here?

src/stories/20_Features/Animations.html
```diff
@@ -258,57 +258,61 @@ <h1>zoom-out</h1>
...
-      <amp-story-grid-layer template="vertical">
-        <h1>pan-left</h1>
-        <amp-img animate-in="pan-left"
-                 animate-in-duration="4s"
-                 layout="fixed"
-                 src="https://picsum.photos/1024/768?image=1026"
-                 width="720"
-                 height="320">
-        </amp-img>
+      <amp-story-grid-layer template="fill">
```
sebastianbenz: Can you re-add the title in the second layer?

Enriqe: Sure 👍

src/stories/20_Features/Animations.html
Enriqe: Since the `amp-story-grid-layer` template=fill only applies for the first element, we must wrap all elements inside the page inside an element. I added a small text regarding this on the ken burns page, I'll add it here as well.

src/stories/20_Features/Animations.html
```diff
@@ -212,6 +222,111 @@ <h1>whoosh-in-right</h1>
...
+    <!-- ## Pan-Left -->
+    <!--
+    <div class="ampstart-card info">
+      **Tip:** When using the `fill` template of the `amp-story-grid-layer` only the first element will be displayed. Wrap elements inside a `<div>` to display everything inside it.
+    </div>
+     -->
+    <amp-story-page id="pan-left">
+      <amp-story-grid-layer template="fill">
+        <div>
+          <h1>pan-left</h1>
```
sebastianbenz: Stupid question: would it be more idiomatic to move the title into a different layer? e.g.:

```
<amp-story-grid-layer template="vertical">
   <h1>zoom-out</h1>
</amp-story-grid-layer>
```

Enriqe: We could do it, but with a couple of caveats:

1. Ordering-wise in the markup, the layer where the title is located would have to be after the layer with the image, since the last "fill" template layer in an amp-story-page will be the element located at the top of the stacking context. 
2.  We would still need to wrap the amp-img in a div, even if it's the only element in the layer - since the "fill" template will change the dimensions of its first child, we need a div to "catch" those style changes. Only then will the image have its dimensions unchanged and the animations will work as intended.

I'll commit the changes to the PR so you can see how the code would look like.

sebastianbenz: Ah - thanks for explaining!

In this sample I'd prefer a simpler markup using a vertical layout:

```
    <amp-story-page id="pan-left">
      <amp-story-grid-layer template="vertical">
        <h1>pan-left</h1>
        <amp-img animate-in="pan-left"
                 animate-in-duration="4s"
                 layout="fixed"
                 src="https://picsum.photos/720/320?image=1026"
                 width="720"
                 height="320">
        </amp-img>
      </amp-story-grid-layer>
    </amp-story-page>
```

The animations (in particular for pan-left and pan-right) are more visible as well (they move only a few pixels with the fill layout). The reason why I'm so picky about this is that this a sample demonstrating the animations, additional markup such as the div is distracting in this case. Wdyt?

Enriqe: No problem!

Regarding the change to a vertical layout, if we take into account
[this new change on the runtime side](https://github.com/ampproject/amphtml/pull/14334), which will scale up the size of the image if it's smaller than the screen size, the image will sit on top of the title. Additionally, there will be some padding on the sides added by the vertical layout. (See screenshot). We could use a z-index to make the title be on top of the image, but not sure what you think about the padding. Let me know what you think.

image

image

Enriqe:
Discussed with @newmuis offline and we will integrate a change on the runtime side [(same PR as above)](https://github.com/ampproject/amphtml/pull/14334) to make it possible to specify an `amp-img` as direct child of template fill and still work as intended. This with the goal that the animations are easy to use for publishers.

It's important to note that for multiple images publishers would still need to wrap them inside a div, and we consider that's okay for now because it's an "advanced" and very specific use case.

Enriqe:
The change is almost ready, so we can just wait until that is merged.

sebastianbenz:
Perfect!

Enriqe:
Change was submitted and I updated the sample.

PR #1205, opened 2018-03-27
Update date picker exampels to use new API

cvializ:
Related to #1059, should fix

### DIFF
--- a/src/20_Components/amp-date-picker.html
+++ b/src/20_Components/amp-date-picker.html
@@ -74,22 +74,22 @@
   <div>
     <button class="ampstart-btn m1 caps" on="tap:simple-date-picker.clear">Clear</button>
     <amp-date-picker
+      id="simple-date-picker"
       type="single"
+      mode="overlay"
       layout="container"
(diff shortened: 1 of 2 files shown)


src/20_Components/amp-date-picker.html
```diff
@@ -100,19 +100,18 @@
   <div>
     <button class="ampstart-btn m1 caps" on="tap:simple-date-picker-2.clear">Clear</button>
```
sebastianbenz: Can you delete the duplicate clear button

src/20_Components/amp-date-picker.html
```diff
@@ -137,21 +136,19 @@
   <div>
     <button class="ampstart-btn m1 caps" on="tap:simple-date-picker-3.clear">Clear</button>
```
sebastianbenz: Can you delete the duplicate clear button. The surrounding div is also no longer needed.

sebastianbenz:
Will deploy straight away.

PR #1209, opened 2018-04-05
Serve random story-ad config

calebcordry:
This is the start to the `amp-story-auto-ads` example. This extension hits a defined endpoint and expects that endpoint to return dynamic a JSON payload it will use to build the ad.

I made this endpoint `/json/amp-story-auto-ads` but happy to change it. Also added a couple of sample configs. These will be modified and more will be more added later. 

This is my first go code :) let me know if it can be improved.

### DIFF
--- a/backend/amp-story-auto-ads.go
+++ b/backend/amp-story-auto-ads.go
@@ -1,0 +1,48 @@
+// Copyright Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


calebcordry:
Thanks! The publishers html looks something like [this](https://github.com/ampproject/amphtml/blob/master/extensions/amp-story/amp-story-ads.md#tagging).

Since the config only has one endpoint they need that endpoint to return dynamic data in order to populate different templates or show different ads.

PR #1210, opened 2018-04-09
Fix <amp-facebook> demo to reflect corresponding Github change

nainar:
This is to document changes made to the code as can be seen here: https://github.com/ampproject/amphtml/pull/14470


nainar:
@sebastianbenz could you TAL?

PR #1215, opened 2018-04-11
Update to reflect allowed password in POST form

cvializ:
We will allow password fields in AMP forms after https://github.com/ampproject/amphtml/pull/14533 validator changes reach the production validator. This was the only reference to passwords being forbidden, so I have updated it.

### DIFF
--- a/src/20_Components/amp-form.html
+++ b/src/20_Components/amp-form.html
@@ -290,7 +290,7 @@
 </form>
 <!-- ## Other form input samples -->
   <!--
-    `amp-form` supports all HTML5 form types with the exception of file, password and image.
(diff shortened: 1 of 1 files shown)


cvializ:
I considered adding a sample for password fields, but I hesitated because I am not sure we want to encourage people to use password fields, or frame password fields in AMP as a best practice. What do you think? /cc @aghassemi

sebastianbenz:
I think password fields are critical for canonical AMP sites. Also, one goal of the amp-form sample is to provide a comprehensive demo of what's possible...

cvializ:
I'll open an issue to add a password example so that the demos are comprehensive

PR #1217, opened 2018-04-11
add new user consent category

sebastianbenz:
* refactor consent samples
* fix canoincals

@martinschierle PTAL


src/70_User_Consent/index.json
```diff
@@ -0,0 +1,3 @@
+{
+  "description": "Samples demonstrating how to implement ad and analytics integrations in AMP."
```
kul3r4: Is this true? Isn't this to demonstrate the user consent?

sebastianbenz: yes, forgot to change it. done.

PR #1226, opened 2018-04-16
Add labels to form examples. Add password example.

cvializ:
Follow up to #1215 to add a password field example.

I also noticed that "Demo" view looked broken for amp-form examples, since there were no labels providing context for the individual examples

| before | after |
|---|---|
|screen shot 2018-04-16 at 14 09 46|screen shot 2018-04-16 at 14 10 15|

And it doesn't feel too redundant on the main page:
screen shot 2018-04-16 at 14 14 47


cvializ:
No problem! And thank you for fixing the invalid `>` character

PR #1229, opened 2018-04-17
Changed a few minor things on buttons

martinschierle:


src/70_User_Consent/Advanced_User_Consent_Flow.html
```diff
@@ -113,27 +117,32 @@
                   <li>{{.}}</li>
                 </template>
               </amp-list>
-              <button on="tap:myConsent.accept" class="ampstart-btn ampstart-btn-secondary caps m1">Accept</button>
+              <button on="tap:myConsent.accept" class="ampstart-btn ampstart-btn-secondary caps m1">Confirm</button>
             </div>
             <div id="choice2" hidden class="message">
               <p class="m1">This is some more information about this choice.</p>
-              <button on="tap:myConsent.reject" class="ampstart-btn ampstart-btn-secondary caps m1">Reject</button>
+              <button on="tap:myConsent.reject" class="ampstart-btn ampstart-btn-secondary caps m1">Confirm</button>
```
sebastianbenz: Is it correct to show Confirm here as well?

src/70_User_Consent/Basic_User_Consent_Flow.html
```diff
@@ -60,11 +62,17 @@
...
-  * Specify the UI to trigger the consents
-  * Specify the policy to be used for this consent
-  * You can dynamically load in the vendors to get consent for via amp-list
+  The `amp-consent` component allows to specify the required user consent for this page.
+  A CORS endpoints can be specified via the `checkConsentHref`
+  attribute. This endpoint will be queried via a POST request to determine if the
+  consent UI specified for it needs to be shown. The response should look like this:
+  ```json
+  {
+    'promptIfUnknown': boolean (true/false)
```
sebastianbenz: can you replace the single quotes with double quotes for valid json

src/70_User_Consent/Advanced_User_Consent_Flow.html
martinschierle: It felt right for me (Rudy wanted the 'reject' one to be 'confirm', and having 'accept' and 'confirm' seemed a bit weird). It feels more consistent like that. But we can check back if you prefer.

PR #1232, opened 2018-04-18
add dismiss and settings to consent flows

sebastianbenz:
//cc @martinschierle


src/70_User_Consent/Basic_User_Consent_Flow.html
```diff
@@ -48,32 +47,39 @@
...
+
   ```json
   {
     "promptIfUnknown": boolean (true/false)
   }
   ```
-  The response will indicate if the consent needs to be shown or not.
-  If it needs to be shown, the element with the id given in `promptUI` will be displayed.
+
+  It is possible, to re-trigger the consent dialog using the `myConsent.prompt()` action. One use case for this is giving users the option
```
kul3r4: remove ","

PR #1238, opened 2018-04-23
Add more date picker examples including fullscreen.

cvializ:
Add the fullscreen demos to the amp-date-picker component examples.

### DIFF
--- a/src/20_Components/amp-date-picker.html
+++ b/src/20_Components/amp-date-picker.html
@@ -22,6 +22,14 @@
     Include the `amp-mustache` component to display an info panel below the date picker.
   -->
   <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+  <!--
+    Include the `amp-lightbox` component to display a date picker inside a fullscreen lightbox view.
+  -->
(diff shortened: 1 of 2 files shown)


src/20_Components/amp-date-picker.html
```diff
@@ -226,5 +242,131 @@
...
+  <!-- ## Date picker in a form-->
+  <!--
+    `amp-date-picker` integrates with AMP forms.
+  -->
+  <form method="post"
+    class="p2"
+    action-xhr="/components/amp-form/submit-form-xhr"
+    target="_top">
+    <amp-date-picker
+      id="static-picker"
```
sebastianbenz: Can you use another ID (duplicate)

src/20_Components/amp-date-picker.html
```diff
@@ -226,5 +242,131 @@
...
+      type="single"
+      mode="static"
+      layout="fixed-height"
+      height="360"
+      format="YYYY-MM-DD"
+      input-selector="#static-picker-input">
```
sebastianbenz: Does the input need to be inside this form?

cvializ: Ah that would make sense. Another copy-paste error.

cvializ:
Fixed the bugs, PTAL

PR #1241, opened 2018-04-24
Add demos for `<amp-fx='fade-in-scroll'>`

nainar:

PR #1242, opened 2018-04-25
Revert "Add demos for `<amp-fx='fade-in-scroll'>` (#1241)"

nainar:
This reverts commit d5a43728226c27a44d85359f587dddb2e934edb1 which isn't working as expected on the live site.


sebastianbenz:
thanks for the quick response

PR #1247, opened 2018-05-02
Update AMP_for_E-Commerce_Getting_Started.html

dev7ch:
Suggestion for consistent text indention, due there have been little inconsistency

### DIFF
--- a/src/10_Introduction/AMP_for_E-Commerce_Getting_Started.html
+++ b/src/10_Introduction/AMP_for_E-Commerce_Getting_Started.html
@@ -4,57 +4,57 @@
 <!doctype html>
 <html ⚡>
 <head>
-  <meta charset="utf-8">
-  <title>AMP for E-Commerce Getting Started</title>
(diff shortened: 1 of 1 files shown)


kul3r4:
@dev7ch Thanks for your pull request, however it seems to me that the indentation is consinstent with other files like [https://github.com/ampproject/amp-by-example/blob/master/src/10_Introduction/_Hello_World.html](https://github.com/ampproject/amp-by-example/blob/master/src/10_Introduction/_Hello_World.html)

dev7ch:
@kul3r4 do you want me to close this PR? ( i see, AMP is intended with 2 spaces every where)

kul3r4:
@dev7ch if we are using 2 spaces everywhere I don't think we should change one file to 4, so I am in favour of closing it.

dev7ch:
sure, i ll close it, but you could may have a look screenshot, this show you what i meant and what is inconsistent.

btw. why do you use 2 Spaces in AMP?

PR #1254, opened 2018-05-03
amp-geo no longer experimental

sebastianbenz:
Also:

* fix JSON syntax error in amp-geo example
* mention amp-geo in geolocation sample


sebastianbenz:
Thanks for the quick review!

PR #1257, opened 2018-05-04
use metadata for amp-date-picker experiment

kul3r4:
Use metadata instead of the usual experiment


kul3r4:
Ok! I will wait those to be in place before merging this.

PR #1258, opened 2018-05-04
Add elements hidden example with video, youtube, and ads to amp-lightbox-gallery

cathyxz:
Add an example to `<amp-lightbox-gallery>` that includes `<amp-youtube>`, `<amp-video>` and `<amp-ad>`.


kul3r4:
Thanks!

PR #1260, opened 2018-05-04
Make the amp-date-picker example valid AMP

cvializ:
After https://github.com/ampproject/amphtml/pull/15106 and this PR are merged then the amp-date-picker examples should be valid.


kul3r4:
thanks

PR #1277, opened 2018-05-18
add amp-story-consent sample

sebastianbenz:


src/stories/50_User_Consent/Story_User_Consent.html
```diff
@@ -0,0 +1,167 @@
...
+
+  <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.1.js"></script>
+  <script async custom-element="amp-video" src="https://cdn.ampproject.org/v0/amp-video-0.1.js"></script>
+  <script async custom-element="amp-list" src="https://cdn.ampproject.org/v0/amp-list-0.1.js"></script>
+
+  <!-- ## Setup -->
+  <!-- Import the `amp-story` extension for creating an AMP Story. -->
+  <script async custom-element="amp-story" src="https://cdn.ampproject.org/v0/amp-story-0.1.js"></script>
+
+  <!-- ... and the `amp-consent` extension. -->
```
kul3r4: if you use the ... you can add the same ones at the end of the previous comment

src/stories/50_User_Consent/Story_User_Consent.html
```diff
@@ -0,0 +1,167 @@
...
+   The `amp-consent` will display the logo specified by `publisher-logo-src` in the consent dialog.
+    -->
+    <amp-story standalone bookend-config-src="<%host%>/json/bookend.json"
+               title="Story With User Consent"
+               publisher="Google"
+               poster-portrait-src="<%host%>/img/story_video_dog_cover.jpg"
+               publisher-logo-src="<%host%>/img/amp_logo_pink.svg">
+
+      <!--
+      The `amp-consent` component allows to specify the required user consent for this page.  A CORS endpoint can be specified via the `checkConsentHref` attribute. This endpoint will be queried via a POST request to determine if the consent UI specified for it needs to be shown. The response should look like this:
```
kul3r4: we are explaining here what amp-consent does but above we are already mentioning it. Maybe move the explanation above?

src/stories/50_User_Consent/Story_User_Consent.html
```diff
@@ -0,0 +1,167 @@
...
+      ```json
+      {
+        "promptIfUnknown": [true|false]
+      }
+      ```
+      The response will indicate if the consent needs to be shown or not. If it needs to be shown, the element with the id given in `promptUI` will be displayed. For AMP Stories, you can specify consent title, message and a list of vendors inside a nested `amp-story-consent` element.
```
kul3r4: maybe add the link in the doc where the list "consent title, message and a list..." is mentioned, so that if that changes in the future the sample is still valid

src/stories/50_User_Consent/Story_User_Consent.html
sebastianbenz: done

src/stories/50_User_Consent/Story_User_Consent.html
sebastianbenz: done

src/stories/50_User_Consent/Story_User_Consent.html
sebastianbenz: added a link to the docs

sebastianbenz:
PTAL

PR #1279, opened 2018-05-21
All images in caption group should belong to same lightbox group

cathyxz:
Fix an issue with the last example, where the images in the same group should be assigned to the same lightbox group.


sebastianbenz:
Thanks - did you intentionally commit go_appengine?

cathyxz:
Oh! That's why everything broke. Let me fix this PR. It was supposed to be
a single line change to the lightbox gallery.

cathyxz:
Fixed!

PR #1281, opened 2018-05-22
Update amp-access-laterpay to use 0.2

trodrigues:


trodrigues:
Nope.

We're using different endpoints on our servers (and the extension was updated to work with those) and there's a new, configuration method as an alternative to our user interface (through a JWT token) but it's also not easy to showcase here anyway.

sebastianbenz:
Thanks for the update!

Issue #1269, opened 2018-05-16
Issue with rendering markdown on amp-bind


fstanis:
It seems that the HTML for amp-bind is slightly off: it creates 3 `<p>` tags inside a `<li>` instead of making them separate. Specifically, this bit:

```
    <!--
      [...]

      Three types of element state can be bound:
      1. `text` for [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent)
      2. `class` for [CSS classes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class)
      3. [Element-specific attributes](https://github.com/ampproject/amphtml/blob/master/extensions/amp-bind/amp-bind.md#bindable-attributes),
      for example, the `src` attribute for `amp-img` or `amp-video`.

      To create a binding, create a new attribute on an element with the syntax:

      `[attribute]="expression"`

    -->
```

Results in:

```html
<p class="mb2 px1">Three types of element state can be bound:</p>
<ol>
  <li>
    <code>text</code> for <a href="https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent">textContent</a>
  </li>
  <li>
    <code>class</code> for <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class">CSS classes</a>
  </li>
  <li>
    <p class="mb2 px1"><a href="https://github.com/ampproject/amphtml/blob/master/extensions/amp-bind/amp-bind.md#bindable-attributes">Element-specific attributes</a>, for example, the <code>src</code> attribute for <code>amp-img</code> or <code>amp-video</code>.</p>
    <p class="mb2 px1">To create a binding, create a new attribute on an element with the syntax:</p>
    <p class="mb2 px1"><code>[attribute]=&quot;expression&quot;</code></p>
  </li>
</ol>
```

While it probably should be:

```html
<p class="mb2 px1">Three types of element state can be bound:</p>
<ol>
  <li>
    <code>text</code> for <a href="https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent">textContent</a>
  </li>
  <li>
    <code>class</code> for <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/class">CSS classes</a>
  </li>
  <li>
[...]

PR #1293, opened 2018-05-31
Fixed issue with markdown on amp-bind

fstanis:
Addresses #1269 (and also removes some dangling spaces in the same file).


kul3r4:
@fstanis could you rebase from master to get the fix for the build?

PR #1295, opened 2018-05-31
Update amp-story bookend attribute

fstanis:
It seems that the lack of a `layout` attribute in `amp-story-bookend` is causing CI to fail.

If I understand correctly, we can just remove this element all together and use the `bookend-config-src` attribute instead to the same effect.

### DIFF
--- a/src/stories/30_Monetization/amp-story-auto-ads.html
+++ b/src/stories/30_Monetization/amp-story-auto-ads.html
@@ -126,7 +126,8 @@
       title="Key Highlights of AMP Conf 2018"
       publisher="The AMP team"
       publisher-logo-src="<%host%>/img/AMP-Brand-White-Icon.svg"
-      poster-portrait-src="<%host%>/img/overview.jpg">
+      poster-portrait-src="<%host%>/img/overview.jpg"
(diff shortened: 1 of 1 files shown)


fstanis:
@calebcordry PTAL

calebcordry:
Thanks for the help, So this is a problem in that I am using `v1.0` of `amp-story` which has a new bookend format, but the validator had not been updated at the point that this was committed. 

cc/ @Enriqe  so he can help sort this out.

fstanis:
Ah, I see - can we just add `layout="nodisplay"` then?

calebcordry:
That's the part that wasn't passing validation before. If you can try it and it now is valid, yes please.

fstanis:
Passes validation with `layout="nodisplay"`, updated PR.

Enriqe:
Yes, what @calebcordry said is correct and `<amp-story-bookend>` should now be ready to use with the `layout="nodisplay"` attribute.

The documentation for the new bookend is on the way and can be followed here:
https://github.com/ampproject/amphtml/pull/15465

kul3r4:
Thanks all!

PR #1308, opened 2018-06-12
[DO NOT MERGE] Fix sidebar toolbar responsiveness example

cathyxz:
Original issue: https://github.com/ampproject/amphtml/issues/15907. 
We should never be able to open sidebar when toolbar is visible. Fix ABE sample page to showcase best practice instead of anti-pattern.

### DIFF
--- a/src/20_Components/amp-sidebar.html
+++ b/src/20_Components/amp-sidebar.html
@@ -57,6 +57,12 @@
     #sidebar-right nav.amp-sidebar-toolbar-target-shown {
         display: none;
     }
+
+    @media(min-width: 784px) {
+      .control-button {
+        display: none;
+      }
+    }
   </style>
 </head>
 <body>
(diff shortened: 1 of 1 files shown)


src/20_Components/amp-sidebar.html
```diff
@@ -57,6 +57,12 @@
     #sidebar-right nav.amp-sidebar-toolbar-target-shown {
         display: none;
     }
+
+    @media(min-width: 784px) {
```
sebastianbenz: We should mention somewhere in the doc that the button needs to be manually hidden for larger screens.

htrex: About best practices, I'm not sure that changing the AMP behaviour after a couple of years it worked in some way, breaking a lot of production websites could be considered a best practice.
The behaviour in question is about an arguable interaction design decision, I think that the current AMP behaviour shouldn't change without changing the library version from 0 to 0.1.
Please think wisely! ;)

aghassemi:
This is still a good as a best practice, but we may need to fix https://github.com/ampproject/amphtml/issues/15907 due to backward compatibility breakage anyway.

cathyxz:
Given the discussion leading to https://github.com/ampproject/amphtml/pull/15990, I'm not very opinionated about not enabling sidebar open while toolbar is present anymore, since this is behavior that we now support. Closing this PR for now.

sebastianbenz:
sgtm - thanks for looking into this!

PR #1309, opened 2018-06-12
Update amp-iframe to make note of amp-vimeo

alanorozco:


sebastianbenz:
Thanks for doing this. Wondering if it would be better to use a different sample altogether.

sebastianbenz:
How about iframing ampproject.org?

alanorozco:
@sebastianbenz I'm a little worried that embedding a full website might communicate the wrong idea to authors?

PR #1314, opened 2018-06-14
Final image changes to amp-fx-collection page

nainar:
Result of finalizing UX review for fly-in-* presets.


sebastianbenz:
Thanks!

PR #1328, opened 2018-06-22
Remove amp-date-picker experiment

cvializ:
To be merged/deployed when the removal of the experiment is out of canary and in prod

### DIFF
--- a/src/20_Components/amp-date-picker.html
+++ b/src/20_Components/amp-date-picker.html
@@ -1,6 +1,3 @@
-<!---{
-    "experiments": ["amp-date-picker"]
-}--->
 <!doctype html>
 <html ⚡>
 <head>


kul3r4:
@cvializ not sure about the error, could you try to rebase and see if that fixes it?

cvializ:
It's a validator error that was fixed in #16320. Not sure when it will be in production, but I'll rebase and see if that helps

kul3r4:
@cvializ I asked you to rebase because I thought it was something fixed on ABE, not realizing you meant [https://github.com/ampproject/amphtml/pull/16320](https://github.com/ampproject/amphtml/pull/16320) which is on the AMP side. So, yes, rebasing obviously didn't make any sense.
I just reverted the change, once [https://github.com/ampproject/amphtml/pull/16320](https://github.com/ampproject/amphtml/pull/16320) will be available I will merge the change. Is there any way for easily know when it's going to be available without checking manually the release notes?

cvializ:
I think the validator team does their rollups with a 3 or 4 day turnaround? I am not sure if there's a list of validator-published PRs since their process is different. I checked with validator.ampproject.org and it looks like it's valid now though.

PR #1329, opened 2018-06-25
Fix broken find/replace across project

kristoferbaxter:
An accidental find/replace across the entire project seems to have caused a few mistakes in the LICENSE file.

Addresses https://github.com/ampproject/amphtml/issues/16272


kul3r4:
Thanks a lot!

PR #1337, opened 2018-07-04
Update amp-date-picker.html

kul3r4:


kul3r4:
@cvializ this has now been deployed

PR #1338, opened 2018-07-05
Add logic for creating hidable "hints" in sample code

fstanis:
# Adding "hint syntax" to AMP by Example

## Background

It seems useful for some samples to have a "callout" with additional information in a larger block of code that's only displayed when the user clicks on it.

An example of this is the [Google Maps Platform docs](https://developers.google.com/maps/documentation/javascript/tutorial#The_Hello_World_of_Google_Maps_v3) where `YOUR_API_KEY` is clickable. In the context of AMP, this can be useful to call out things that should be done server-side and shouldn't be copy-pasted directly (this came up for some AMPHTML Email samples that are WIP, but is probably applicable elsewhere as well).

## Usage

Comments wrapped inside `<!--~` and `~-->` are treated as hints and will apply to the element that follows them. For example:

```html
<!--~ This hint applies to everything between <div> and </div> ~-->
<div>
  <p>Hello.</p>
  <p>Also, hi.</p>
</div>
<p>This part is not clickable</p>
<!--~ This hint applies to the <br> tag ~-->
<br>
<p>This part is not clickable</p>
```

## Mechanics

Since [highlight.js](https://highlightjs.org/) has no native support for this, the mechanism used to achieve is the following:

1.  Treat comments inside `<!--~` and `~-->` as a special case, i.e. leave them in the HTML sent to highlight.js rather than processing and breaking into a new section.
2.  Store the content between `<!--~` and `~-->` in an internal state and assign the hint an ID.
3.  Replace the whole comment element with `<!--START_HINT_123-->` (where `123` is the previously assigned ID).
4.  Process the next element in the HTML  and append `<!--END_HINT-->` after its closing tag (or immediately after if it's a tag that doesn't close, like `<input>`.
5.  After calling the `highlight` function from highlight.js, process the output and use the generated `<span class="hljs-comment">` tags as hooks to replace with the HTML for displaying the hint.

## Other changes

Logic for the [endOfCurrentTag](https://github.com/ampproject/amp-by-example/blob/8d936229796a5bffdd526d4f1ea9414b1e2bd285/lib/DocumentParser.js#L184) method inside `DocumentParser.js` had to be amended to take into account [void elements](http://w3c.github.io/html/syntax.html#void-elements) and treat them as ending themselves, i.e. `<base href="/">` is treated as if the line is `<base href="/"></base>`. This is to allow applying hints to these tags (most notably `<input>`).

## Future improvements

-   It's currently impossible to have finer granulation than tag-level, e.g. a hint can't be applied to a single attribute.
-   The content inside the hint should be treated as markdown rather than HTML.
-   Copying code with a hint currently omits the hint. Ideally, it should include the hint as a comment.

### DIFF
--- a/lib/CodeSection.js
+++ b/lib/CodeSection.js
@@ -95,7 +95,7 @@ module.exports = class CodeSection {
   }
 
   appendHint(hint) {
-    if (this.currentHint === '') {
+    if (!this.currentHint) {
       this.appendCode(`<!--${hintStartPlaceholder}_${this.hints.length}-->`);
     }
     this.currentHint += hint;
--- a/lib/CodeSection.js
+++ b/lib/CodeSection.js
@@ -17,7 +17,7 @@
 "use strict";
 
 const S = require('string');
-const highlight = require('highlight.js').highlight;
+const hljs = require('highlight.js');
 const marked = require('marked');
 const renderer = new marked.Renderer();
 renderer.heading = function (text, level) {
@@ -32,10 +32,20 @@ renderer.paragraph = function (text) {
 
 const encodedTemplateRegexp = /\[\[\s*<.*?>([A-Za-z]*?)\s*(<.*?>)?(\.[A-Za-z]*)?\s*<\/span>\s*\]\]/g
 
+const hintStartPlaceholder = 'START_HINT';
+const hintStartRegexp = new RegExp(`<span class="hljs-comment">&lt;!--${hintStartPlaceholder}_(\\d+)--&gt;</span>\\n?`, 'g');
+const hintEndPlaceholder = 'END_HINT';
+const hintEndRegexp = new RegExp(`<span class="hljs-comment">&lt;!--${hintEndPlaceholder}--&gt;</span>\\n?`, 'g');
+
+// `${hintStartHtmlOpen} <hint text> ${hintStartHtmlClose} <element> ${hintEndHtml}`
+const hintStartHtmlOpen = '<label class="has-hint"><input class="show-hint" type="checkbox"><div class="hint">';
+const hintStartHtmlClose = '</div>';
+const hintEndHtml = '</label>';
+
 marked.setOptions({
   highlight: function(code, lang) {
     if (lang) {
-      return highlight(lang, code).value;
+      return hljs.highlight(lang, code).value;
     } else {
       return S(code).escapeHTML().s;
     }
@@ -65,6 +75,8 @@ module.exports = class CodeSection {
     this.codeOffset = 0;
     this.headings = [];
     this.storyPageId = '';
+    this.currentHint = '';
+    this.hints = [];
   }
 
   appendDoc(doc) {
@@ -82,12 +94,27 @@ module.exports = class CodeSection {
     this.code += code.substring(startIndex) + '\n';
   }
 
+  appendHint(hint) {
+    if (this.currentHint === '') {
+      this.appendCode(`<!--${hintStartPlaceholder}_${this.hints.length}-->`);
+    }
+    this.currentHint += hint;
+  }
+
+  endHint() {
+    let hint = this.currentHint.replace(/<!--~(.*?)~-->/, '$1')
+    hint = hint.replace(/\s+/g, ' ').trim();
+    this.hints.push(hint);
+    this.currentHint = '';
+    this.appendCode(`<!--${hintEndPlaceholder}-->`);
+  }
+
   appendPreview(code) {
     this.preview += code + '\n';
   }
 
   escapedCode() {
-    const result = highlight('html', this.codeSnippet()).value;
+    const result = this.highlight('html', this.codeSnippet());
     return this.cleanUpCode(result);
   }
 
@@ -150,7 +177,7 @@ module.exports = class CodeSection {
   normalizeDoc(string) {
     let startIndex = string.indexOf(COMMENT_START);
     if (startIndex == -1) {
-      startIndex = 
+      startIndex =
         this.stripLeadingWhitespace(string, this.commentOffset);
     } else {
       this.commentOffset = startIndex;
@@ -203,6 +230,20 @@ module.exports = class CodeSection {
       name: name
     };
     this.headings.push(heading);
-  };
-};
+  }
 
+  replaceHints(html) {
+    const result = html.replace(hintStartRegexp, (matches, id) => {
+      id = Number(id);
+      const hint = this.hints[id];
+      return `${hintStartHtmlOpen}${hint}${hintStartHtmlClose}`;
+    });
+    return result.replace(hintEndRegexp, hintEndHtml);
+  }
+
+  highlight(name, value) {
+    let result = hljs.highlight(name, value).value;
+    result = this.replaceHints(result);
+    return result;
+  }
+};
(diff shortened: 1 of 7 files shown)


juliantoledo:
@fstanis this is awesome!

juliantoledo:
@fstanis before merging I think @kul3r4 should review/test it in staging because it is a core change.

kul3r4:
@fstanis wonderful job! could you deploy a version of ABE containing this changes somewhere so that we can test it?

fstanis:
So, I ran a diff on the generated files to compare with master and it seems the only regression is in `/samples_templates/product_browse_page/` where the final `</div>` is dropped because the `<input>` is considered the end of that section.

Seems the easiest way to fix it is to just use the hint syntax there, which also serves as a good demo. :)

Will get a staging version somewhere next week for some additional reviews.

fstanis:
Staged: https://abe-staging.appspot.com/samples_templates/product_browse_page/

kul3r4:
@spacedino could you give us some feedback? In order to see this in action:

- go to https://abe-staging.appspot.com/samples_templates/product_browse_page/
- got the section "Header with search"
- scroll down in the code section until the end, the tag `input type="submit"` with the blue background has the hint: you can click on it to see the hint

lib/CodeSection.js
```diff
@@ -82,12 +94,27 @@ module.exports = class CodeSection {
     this.code += code.substring(startIndex) + '\n';
   }
 
+  appendHint(hint) {
+    if (this.currentHint === '') {
```
sebastianbenz: nit: `if(!this.currentHint)` is a bit more robust.

lib/DocumentParser.js
```diff
@@ -22,6 +22,8 @@ const elementSorting = require('./ElementSorting');
 const beautifyHtml = require('js-beautify').html;
 
 const SINGLE_LINE_TAGS = ['link', 'meta', '!doctype'];
+const VOID_TAGS = ['area', 'base', 'br', 'col', 'embed', 'hr', 'img', 'input',
+                   'link', 'meta', 'param', 'source', 'track', 'wbr'];
```
sebastianbenz: Good one! Can we merge this with `SINGLE_LINE_TAGS`?

spec/compiler/DocumentParserSpec.js
```diff
@@ -140,6 +155,13 @@ describe("DocumentParser", function() {
     });
   });
 
+  describe("void tags", function() {
+    it("base", function() {
```
sebastianbenz: nit: try to make the `it` description describe whats tested, e.g.: `it("ends void tags  automatically")`

templates/css/example.css
```diff
@@ -152,3 +152,47 @@ code {
...
+  cursor: pointer;
+  background-color: rgba(132, 183, 232, 0.25);
+  border-radius: 5px;
+  transition: background-color .3s;
+}
+.abe-code .has-hint:hover {
+  background-color: rgba(132, 183, 232, 0.5);
+}
+.abe-code .has-hint:before {
+  content: "🛈";
```
sebastianbenz: Can we find a better icon? Maybe 💡 ?

templates/css/example.css
```diff
@@ -152,3 +152,47 @@ code {
...
+  bottom: 100%;
+  border-radius: 5px;
+  padding: 5px;
+  width: 100%;
+  text-align: center;
+  margin-bottom: 7px;
+  background: #fff;
+  color: #4a4a4a;
+  line-height: 1.5rem;
+  font-family: Montserrat,-apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Oxygen,Ubuntu,Cantarell,Fira Sans,Droid Sans,Helvetica Neue,Arial,sans-serif;
```
sebastianbenz: Do we need the font here? If yes, can you please extract the font into a CSS variable.

sebastianbenz:
Took another look. Few suggestions:

* left align text
* use ABE color scheme for background color. 
* animation expand up instead of reveal (similar to toast example). What do you think about making the tooltip appear inside the doc to not overlay the code above?
* show close icon in tooltip (but keep the logic of clicking anywhere in the tooltip to close it).
* add a bit upper and lower margin & padding to `has-hint`:

      .abe-code .has-hint {
        ...
        padding: 0.5rem 0;
        margin: 0.5rem 0;
      }

fstanis:
Latest version here: https://abe-staging.appspot.com/samples_templates/product_browse_page/

@spacedino would still love to hear your feedback before we merge, if you have the time.

sebastianbenz:
Two comments:

- can you speed up the transition, e.g.: `transition: transform .25s ease-in-out;`
- the close icon does not show

spacedino:
@fstanis looks good but I am concerned that they are getting lost because they look so similar to the code snippet containers. Them not having a drop shadow is also adding to this I think. It's hard for me to tell though because I cant interact with it yet (unless I'm missing something). Maybe we can make the container light pink? Like 10% of the ABE pink?

Agree with all of Sebastians comments as well. 

- the font for the hint should be Montserrat to match the body / explanatory text.
- i think the link is broken in the below image
- can we use the same treatment for the component names through out? I realize they grey on grey wont work so maybe we can try the pink bg or maybe a 1px border?
screen shot 2018-07-27 at 11 21 15 am


screen shot 2018-07-26 at 12 20 42 pm

spacedino:
Also I think this is the ABE visual design for hints? Or is that something else....?

screen shot 2018-07-27 at 11 29 35 am

sebastianbenz:
This is what the current version looks like:

hint

sebastianbenz:
@spacedino yes-you're right. We probably should use the same background color.

fstanis:
Updated to match existing yellow hint @spacedino linked to and fixed issues with close icon (hopefully, please confirm - seems to be Mac-specific) and animation speed.

Latest version:

screenshot_2018-07-27_19-44-26

Should be live here: https://abe-staging.appspot.com/samples_templates/product_browse_page/

> can you fix this as part of your PR Filip?

Seems this one was fixed in some other PR, but I neglected to rebase master. Fixed now.

spacedino:
Oh actually seeing the yellow in practice I think I was wrong. The white one with rounded corners looks better. Apologies for logging those previous bugs here, I should've created a new ticket. 

Doe we even need an X? How about this chevron svg attached?

[ic_keyboard_arrow_down_grey600_24dp.svg.zip](https://github.com/ampproject/amp-by-example/files/2236898/ic_keyboard_arrow_down_grey600_24dp.svg.zip)

sebastianbenz:
I thought you meant using the yellow for the code background (instead of the light pink).

spacedino:
Yea I think that is better bc I'm not sure if the color text is a11y compliant with the pink. I do love the light pink though :)

fstanis:
screenshot_2018-07-27_20-20-47

Yellow variation feels a bit "off", i.e. the contrast is low:

screenshot_2018-07-27_20-21-32

> also I'm seeing a red arrow

Note to self: never use emojis as icons again.

fstanis:
Hi all, any objections to submitting this as-is (pink version from my last screenshot)?

I'm OK with making further UX changes later if needed.

sebastianbenz:
SGTM

PR #1349, opened 2018-07-11
Use fully qualified Urls for `amp-consent` x 3

aghassemi:
same as #1347 for a different file (sorry using GitHub UI and could not do a single PR)


kul3r4:
All of those have now been deployed!

PR #1366, opened 2018-07-25
Fix layout for date-countdown, simplify CSS for friend requests

fstanis:


src/amphtml-email/20_Samples/Friend_Requests_Email.html
```diff
@@ -87,46 +78,22 @@
...
-      color: white;
-    }
-
-    .friendrequest form .button-delete {
-      background: #eee;
-    }
-
-    .friendrequest form label {
-      text-transform: uppercase;
+    .friendrequest form .ampstart-btn {
```
sebastianbenz: is it really necessary to change the button styles?

sebastianbenz: Otherwise I think it's better to use your own styles.

src/20_Components/amp-date-countdown.html
```diff
@@ -32,7 +32,7 @@
     The number of days, hours, minutes and seconds remaining is bound to `{{d}}`, `{{h}}`, `{{m}}` and `{{s}}` respectively.
     Double digit format variants (`{{dd}}`, `{{hh}}`, `{{mm}}` and `{{ss}}`) are also available.
   -->
-  <amp-date-countdown timestamp-seconds="2147483648" layout="fixed-height" height="50">
+  <amp-date-countdown timestamp-seconds="2147483648" layout="flex-item">
```
sebastianbenz: I find using the `flex-item` layout here a bit "magic". Can you either explain why you're using it or use another layout?

sebastianbenz:
There's also a validation error:

screen shot 2018-07-25 at 14 31 48

fstanis:
> question: doesn't the form xhr-action URL also need to include the auth token?

I put the auth token in a hidden input instead:

https://github.com/ampproject/amp-by-example/blob/30f63476d4f6f8f797654d2ca576ee52c9a3efa8/src/amphtml-email/20_Samples/Friend_Requests_Email.html#L161-L163

> feature request: can you change the result text from `done` to either `added` or `deleted`?

I think it's better to do this server-side - to that end, I added a template for displaying result and will add a custom server endpoint in another PR.

> There's also a validation error

Seems to be an issue with the validator, created [#17078](https://github.com/ampproject/amphtml/pull/17078) to fix.

sebastianbenz:
SGTM - thanks!

sebastianbenz:
Just one thing, can you make sure that the build is passing before submitting. Thanks!

PR #1372, opened 2018-07-26
Cleanup gulpfile

sebastianbenz:
* Remove no longer needed copy scripts tasks
* Fix duplicate scripts attribute in paths variable.
* ... fix resulting eslint errors


tasks/compile-example.js
```diff
@@ -26,13 +26,18 @@ const DocumentParser = require('../lib/DocumentParser');
...
 const storyBookend = require('./bookend.json');
 
 const STORY_EMBED_DIR = __dirname + '/../api/dist/';
-const AMP_STORY_CLEANER_REGEX = ['amp-story', 'amp-story-auto-ads', 'amp-consent'].map(extension =>
-  new RegExp('<script\\s+async\\s+custom-element="' + extension + '"\\s+src="https:\\/\\/cdn\\.ampproject\\.org\\/v0\\/' + extension + '-\\d\\.\\d\\.js"><\\/script>')
-);
+const AMP_STORY_CLEANER_REGEX =
+  ['amp-story', 'amp-story-auto-ads', 'amp-consent']
+      .map(extension =>
+        new RegExp('<script\\s+async\\s+custom-element="' + extension
```
fstanis: Nit: I recommend using backticks here, that was we avoid all those `\\` and it's also easier to concatenate the `extension` variable.

gulpfile.js
```diff
@@ -173,80 +171,74 @@ gulp.task('deploy:site:prod', 'deploy to production site', function() {
...
 
 gulp.task('copy:license', 'copy license', function() {
   return gulp.src('LICENSE')
-    .pipe(cache('license'))
-    .pipe(rename(function(path) {
-      path.extname = ".txt";
-    }))
-    .pipe(gulp.dest(paths.dist.dir));
+      .pipe(cache('license'))
+      .pipe(rename(function(path) {
```
fstanis: Nit: maybe replace this kind of thing with arrow functions (there's a few other places in the file as well)?

fstanis:
We should split that gulpfile into a few modules in some future PR, as it's becoming rather difficult to maintain.

tasks/compile-example.js
sebastianbenz: * the `\\` are needed when you init a regex with a string (instead of `/.../`). 
* backticks have no real advantage here as I need to break the lines anyway because of eslint...

gulpfile.js
sebastianbenz: We haven't used these anywhere yet.

sebastianbenz:
+1 to splitting the gulpfile. But let's wait until after the re-design...

PR #1399, opened 2018-07-27
Copy button example

samthor:
cc @sebastianbenz


src/30_Advanced/Copy_Button.html
```diff
@@ -0,0 +1,93 @@
...
+       "disabled" indicator.
+       We pass the data to be copied via the `#` param. The copier "helper" page can always serve
+       the same HTML and can be cached in its own right.
+  -->
+  <amp-iframe
+      sandbox="allow-scripts"
+      width="64"
+      height="42"
+      frameborder="0"
+      src="<% api.host %>/iframe/resizable-iframe.html#This text was copied from &quot;AMP by Example&quot;!">
```
sebastianbenz: s/resizable-iframe/copier

api/iframe/copier.html
```diff
@@ -0,0 +1,56 @@
...
+<body>
+<button id="copy">Copy</button>
+<input type="text" id="_i" hidden />
+<script>
+/*
+  Decode the location from the hash, placing it in our hidden text element. To copy, we show the
+  input, focus it, and call `execCommand('copy')`, before hiding it again.
+  See: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard
+ */
+_i.value = window.decodeURIComponent(window.location.hash.substr(1));
```
sebastianbenz: Move this into the onclick handler to allow dynamically changing the fragment.

src/30_Advanced/Copy_Button.html
sebastianbenz: One other thing: amp-iframes need to be served from a different host if the `allow-same-origin` sandbox attribute is used. This is why we have a different `api.host`. In this case that is not necessary and I think it's better to serve the iframe from the same domain (to not indicate that this might be required). You can simply copy `copier.html` into the `/static` dir and then reference it via `/copier.html'.

src/30_Advanced/Copy_Button.html
```diff
@@ -0,0 +1,93 @@
...
+button#copy {
+  cursor: not-allowed;
+  width: 100%;
+  height: 100%;
+}
+/* misc page CSS */
+body {
+  font-family: Sans-Serif;
+}
+main {
```
sebastianbenz: Can you make this rule less generic, otherwise you're messing with the overall layout.

src/30_Advanced/Copy_Button.html
```diff
@@ -0,0 +1,93 @@
...
+  display: flex;
+  align-items: center;
+  padding: 1em;
+}
+main span {
```
sebastianbenz: Same as above :-)

src/30_Advanced/Copy_Button.html
samthor: Done

src/30_Advanced/Copy_Button.html
samthor: Done

api/iframe/copier.html
samthor: Done

src/30_Advanced/Copy_Button.html
samthor: Done

samthor:
All done, awaiting another review 👍

sebastianbenz:
[It's live](https://ampbyexample.com/advanced/copy_button/) (I had to make some [cosmetic changes](https://github.com/ampproject/amp-by-example/pull/1422/files) to improve the way the sample is rendered in the doc view). Thanks again!

PR #1423, opened 2018-07-31
Add sample for AMPHTML Email bookmarks in email

fstanis:


src/amphtml-email/20_Samples/Bookmark_Article_Email.html
```diff
@@ -0,0 +1,178 @@
...
+<!--
+## Introduction
+
+This sample showcases how to use AMPHTML email to create an email with a
+personalized digest of articles, giving the user an ability to bookmark some of
+them for later reading.
+-->
+
+<!-- -->
+<!-- Doctype declaration is required. -->
```
sebastianbenz: You don't need to document the boilerplate. Focus on what's specific to this sample.

src/amphtml-email/20_Samples/Bookmark_Article_Email.html
```diff
@@ -0,0 +1,178 @@
...
+      margin-top: .4em;
+      clear: right;
+    }
+  </style>
+</head>
+<!-- ## Body -->
+<!-- -->
+<body>
+  <!--
+    Each list item has a normal link as well as an `amp-form` which is only used for sending an
```
sebastianbenz: This sample has almost 100 lines of code and a single line of documentation. This is a bit hard to comprehend - in particular because the same code is repeated multiple times. Can you make this more focused?

fstanis: Any suggestions?

My aim is to make it clear the list itself is statically generated, with the `amp-form` being the only "special" part of it. I can drop it to two list items, but I'm not sure what I may do beyond that.

sebastianbenz: One possibility would be to use amp-list. This has the further advantage that the content is always fresh (e.g. only the latest articles).

src/amphtml-email/20_Samples/Bookmark_Article_Email.html
```diff
@@ -0,0 +1,123 @@
...
+
+</html>
+
+<!--
+  ## Server-side considerations
+
+  Because there is no authentication for outgoing XHR calls from AMP email
+  messages, all authentication should be done using access tokens.
+
+  In the sample above, `REPLACE_WITH_YOUR_TOKEN` represents an access token.
```
sebastianbenz: This would be a good candidate for a hint in the code section.

src/amphtml-email/20_Samples/Bookmark_Article_Email.html
```diff
@@ -0,0 +1,123 @@
+<!---{
+ "preview": "amp4email",
+ "skipValidation": "true",
+ "draft": "true"
+}--->
+<!--
+## Introduction
+
+This sample showcases how to use AMPHTML email to create an email with a
```
sebastianbenz: Can you give a more abstract overview of what this sample demonstrates, e.g.: "This sample demonstrates how to let users perform an action (maybe transaction) in an AMPHTML email. "

PR #1425, opened 2018-07-31
Revamp amp-access to be simpler and more generic

fstanis:
-   Rephrase docs to be more generic around authentication, rather than publisher-specific.
-   Make it clear that `amp-access` response allows tiered access / ACL (through `powerUser` property).
-   Simplify login page, remove Google login (to be added as separate sample in future PR).
-   Remove logout page (use redirect instead).
-   Use email of logged in user in auth response instead of "Charlie".
-   Make it possible for login to fail if email is invalid.
-   Make it clear the login page needn't be AMPHTML.
-   Remove unused `amp-access` backend handlers from hotel.go.


static/login.html
```diff
@@ -14,40 +11,32 @@
...
-          <input type="password" name="password" value="123456">
+          <input type="password" name="password" value="">
         </div>
-          <input name="returnurl" type="hidden" value="{{.ReturnURL}}">
-          <input name="token" id="token" type="hidden" value="">
-          <input type="submit" value="Login" class="button">
+        <input name="returnurl" type="hidden" value="{{.ReturnURL}}">
+        <input type="submit" value="Login" class="button">
       </form>
-      <p> Or you can use Google Sign in </p>
```
sebastianbenz: The login page is also used here: 
https://ampbyexample.com/samples_templates/comment_section/, so you'd need to change that page as well. I'd prefer to keep it though until we have a dedicated sample.

fstanis: I'm aware of it, that sample remains fully functional with this PR.

src/20_Components/amp-access.html
```diff
@@ -45,48 +50,60 @@
...
   `amp-access-hide` will default the component to be hidden and enabled later
   if the `amp-access` expression is evaluated to `TRUE` -->
-  <section class="p1" amp-access="NOT subscriber" amp-access-hide role="button" tabindex="0">
+  <section class="p1" amp-access="NOT loggedIn" role="button" tabindex="0">
     <a class="ampstart-btn caps" on="tap:amp-access.login-sign-in">Login</a>
   </section>
-  <!-- The logout link will only be shown for logged in subscribers. We use a trick here: the logout is a login action that directly redirects back to the original URL. This has the advantage that the page updates without being reloaded.  -->
-  <section class="p1" amp-access="subscriber" amp-access-hide role="button" tabindex="0">
+  <!-- The logout link will only be shown for logged in subscribers. We use a trick here: the logout is a
+    login action that directly redirects back to the original URL. This has the advantage that the page updates without being reloaded.
```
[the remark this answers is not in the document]
fstanis: The old "logout page" was just a double redirect though, not sure what the logout page should contain? Otherwise, I'm fine with having it.

static/login.html
sebastianbenz: Doesn't it use the same login page?

src/20_Components/amp-access.html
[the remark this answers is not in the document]
sebastianbenz: Ah ok, in that case it doesn't make sense. I thought it was a separate logout page.

static/login.html
fstanis: It does, but this PR doesn't break it, i.e. you can still use the same login page and the login will still work on the comment section sample.

sebastianbenz: yes, but the sample doc mentions google signin

fstanis: Ouch, missed that, good point.

src/20_Components/amp-access.html
```diff
@@ -21,8 +24,13 @@
   the authorization response -->
   <script async custom-template="amp-mustache" src="https://cdn.ampproject.org/v0/amp-mustache-0.2.js"></script>
 
-  <!-- Add the configuration for the `authorization`, `pingback` and the `login` endpoints. We provide two login endpoints, one for sign-in and one for sign-out. It is
-  also possible to create a fallback response that will be used if the authorization fails with the `authorizationFallbackResponse` attribute -->
+  <!-- Add the configuration for the `authorization` and the `login` endpoints. We provide two login endpoints, one for sign-in and one for sign-out. It is
+    also possible to create a fallback response that will be used if the authorization fails with the `authorizationFallbackResponse` attribute.
+
+    The [Pingback Endpoint](https://www.ampproject.org/docs/reference/components/amp-access#pingback-endpoint) is also defined,
```
sebastianbenz: Suggestion, rephrase to: "The [Pingback Endpoint](https://www.ampproject.org/docs/reference/components/amp-access#pingback-endpoint) will be called when a user views the document, which can be used to update metering information.

src/20_Components/amp-access.html
```diff
@@ -45,50 +53,62 @@
...
 
   <!-- Use `on="tap:amp-access.login-sign-in"` to open the login dialog when the element is
-  clicked.
+  clicked. This simply opens a popup window that takes the user to the URL defined inside the
+  configuration at the top.
 
   `amp-access-hide` will default the component to be hidden and enabled later
   if the `amp-access` expression is evaluated to `TRUE` -->
-  <section class="p1" amp-access="NOT subscriber" amp-access-hide role="button" tabindex="0">
+  <section class="p1" amp-access="NOT loggedIn" role="button" tabindex="0">
```
sebastianbenz: Can you re-add the `amp-access-hide` attribute.

kul3r4:
Why Google sign in has been removed? It was easy to show around that it's possible to use Google sign in with AMP

fstanis:
Next PR (that I'm working on right now) will bring it back for the comments sample. I just wanted to separate it from the "normal" login, to simplify the `amp-access` demo.

fstanis:
Update: #1475 "brings back" Google sign-in, so I'll merge this after that PR, as to not end up in a state where we remove it.

kul3r4:
@fstanis Can we merge this?

PR #1435, opened 2018-08-03
Fix stories bookend configuration.

gmajoulet:
Bookend configuration was still using the `shareProviders` configuration for the `v0.1`. Upgrading to `v1.0`.

See: https://www.ampproject.org/docs/design/visual_story/create_bookend


ithinkihaveacat:
Thanks! Just noticed this too.

PR #1445, opened 2018-08-08
Add amp-consent (for GDPR) and ClientID (for analytics)

morsssss:
Try it out... hope you like it!


templates/consent.html
```diff
@@ -0,0 +1,32 @@
...
+          "promptIfUnknownForGeoGroup": "eu",
+          "promptUI": "consent-popup"
+        }
+      }
+    }
+  </script>
+
+  <div id="consent-popup" class="consent-popup">
+    <div>
+      <a href="#" on="tap:consent.dismiss" class="close-button" role="button" tabindex="0">✕</a>
```
sebastianbenz: Can we move this to the right? Somehow I always expect close buttons to be on the right. Or is it just me?

templates/index.html
```diff
@@ -6,12 +6,16 @@
...
+    <script async custom-element="amp-geo" src="https://cdn.ampproject.org/v0/amp-geo-0.1.js"></script>    
     {{> head.html}}
     <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+    <meta name="amp-google-client-id-api" content="googleanalytics">
     <style amp-boilerplate>body{-webkit-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-moz-animation:-amp-start 8s steps(1,end) 0s 1 normal both;-ms-animation:-amp-start 8s steps(1,end) 0s 1 normal both;animation:-amp-start 8s steps(1,end) 0s 1 normal both}@-webkit-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-moz-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-ms-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@-o-keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}@keyframes -amp-start{from{visibility:hidden}to{visibility:visible}}</style><noscript><style amp-boilerplate>body{-webkit-animation:none;-moz-animation:none;-ms-animation:none;animation:none}</style></noscript>
     <style amp-custom>
       {{> css/ampstart.css}}
-      {{> css/shared.css }}
+      {{> css/shared.css}}
+      {{> css/consent.css}}
```
sebastianbenz: Can you inject this into shared.css instead - this way it'll be available to all pages.

templates/index.html
```diff
@@ -6,12 +6,16 @@
     <script async src="https://cdn.ampproject.org/v0.js"></script>
     <script async custom-element="amp-sidebar" src="https://cdn.ampproject.org/v0/amp-sidebar-0.1.js"></script>
     <script async custom-element="amp-selector" src="https://cdn.ampproject.org/v0/amp-selector-0.1.js"></script>
+    <script async custom-element="amp-consent" src="https://cdn.ampproject.org/v0/amp-consent-0.1.js"></script>
```
sebastianbenz: Can you move the imports to head.html. And, this is where it gets ugly, you need to surround it with checks if the sample already contains the imports to avoid duplicate imports:

```
{{^includesGeo}}
<script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>
{{/includesAmpGeo}}
{{^includesConsent}}
<script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>
{{/includesAmpConsent}}
```

The variables need to be defined here: https://github.com/ampproject/amp-by-example/blob/master/tasks/compile-example.js#L231

sebastianbenz: This might get a bit weird though for the amp-consent samples, it might make sense to disable the dialog there.

templates/index.html
```diff
@@ -117,5 +121,6 @@ <h3 class="mb1">
     {{> ampstart/footer.snip.html}}
     {{> analytics.html}}
     {{> serviceworker.html}}
+    {{> consent.html}}
```
sebastianbenz: There are a few more templates where this needs to be injected: 

* preview-*.html
* example.html

Ideally, could you create a new template `footer.html` which then includes analytics.html, service worker.html and consent.html

templates/consent.html
morsssss: Its location varies from device to device!  I'm following the designs [here](https://docs.google.com/document/d/1j4zVW5oeyGwJ7Usg14JW4yXI4KrbY0FzNmgFxM2BZro/edit) and [here](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZeADcnNEYkpi/files/MCHtA7U1iMGr6yzAke-xKMPUgUfwZnEM3bg):

templates/index.html
morsssss: Ok, done.

I'd wanted for a while to write something that simply detects which components are used on an AMP and inserts the proper `<script>`s.  So @garanj kindly did this for our new e-comm project:

https://github.com/morsssss/amp-ecomm/pull/3

I get what you say about confusion on the amp-consent samples - but I'd just guess that legally we're required to have this cookie notice sitewide.

templates/index.html
morsssss: I can simply copy the CSS into shared.css, but I feel a little guilty, since we've then got a separate HTML file, but no corresponding separate CSS file.

I guess one alternative is to create a CSS build process, which is probably outside the scope of this PR... another is to make a CSS file that simply imports all of these... or I can just do what you suggest and we can deal with the problem if we ever get even more CSS files :)

templates/index.html
morsssss: Makes sense!

The site already has a `footer.snip.html` file, though.  Is there a logical name for another footer template?  Or can I simply include the three templates you mention in `footer.snip.html` ?

sebastianbenz: That's for the ampstart template which is not used on all pages (e.g. the previews). 

How about `footer-boilerplate.html`?

templates/index.html
sebastianbenz: We have to see, if it works and validates it's fine to keep it. Otherwise I'd say we have to remove it from the page.

templates/index.html
sebastianbenz: Sorry, I meant link it from inside `shared.css`:

```
{{> css/consent.css }}
```

templates/consent.html
sebastianbenz: sgtm

templates/index.html
morsssss: Ok - done!
The only part I'm concerned about is preview.html, where analytics and serviceworker were only included this way:

```
    {{^document.isAmpStory}}
    {{> analytics.html}}
    {{> serviceworker.html}}
    {{/document.isAmpStory}}
```

is that ok?

templates/index.html
morsssss: Hmm... this didn't work for me. 🙄 I don't know your build process that well - but when I tried this, in dist/, I still see that include in `shared.css`.  But the include never happens.  Instead, that include string vanishes at some point in the build process - perhaps during minification, or something.

I imagined that the include string might have been replaced by the file contents at some point in the build, but, for me, it doesn't. 😿

templates/index.html
sebastianbenz: You can't use it there - you have to inject it manually similar to analytics.

templates/index.html
sebastianbenz: tbh, I don't understand it either - let's keep it separate similar to ampstart.css etc. Sorry for the confusion.

morsssss: Ok! Reverted to separate version.

sebastianbenz: Can you also delete the two imports from index.html

templates/index.html
morsssss: It looks fine on the amp-consent samples... but there are indeed errors as the pages have two amp-consent's.  And it doesn't appear on the geolocation-based page at all.  That is, `user_consent/geolocation-based_consent_flow`.

Dang!

@rudygalfi, @ericlindley-g - any opinion here?

templates/index.html
morsssss: Ok!

What pages use the preview* templates?  I'll test.

templates/index.html
morsssss: Yes!  I had, but I hadn't committed the new file yet, pending other changes.  Will do so now.

templates/head.html
```diff
@@ -11,6 +11,13 @@
 {{^includesIframe}}
 <script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>
 {{/includesIframe}}
+{{^includesConsent}}
```
sebastianbenz: Can you make both imports depend on the availability:

```
{{^includesGeo}}
{{^includesConsent}}
<script async custom-element="amp-consent" src="https://cdn.ampproject.org/v0/amp-consent-0.1.js"></script>
<script async custom-element="amp-geo" src="https://cdn.ampproject.org/v0/amp-geo-0.1.js"></script>    
{{/includesConsent}}
{{/includesGeo}}
```

templates/consent.html
```diff
@@ -0,0 +1,32 @@
+<amp-geo layout="nodisplay">
```
sebastianbenz: Can you only add this if amp-consent and amp-geo are not already present on the page:

```
{{^includesconsent}}
 {{^includesgeo}}
<amp-geo layout="nodisplay">
    {
      "isocountrygroups": {
        "eu": ["at", "be", "bg", "ch", "cy", "cz", "dk", "de", "ee", "el", "es", "fi", "fr", "hr", "ie", "is", "it", "li", "lv", "lt", "lu", "hu", "mt", "nl", "no", "pl", "pt", "ro", "si", "sk", "se", "uk"]
      }
    }
  </script>
</amp-geo>

<amp-consent id="consent" layout="nodisplay">
    {
      "consents": {
        "eu": {
          "promptifunknownforgeogroup": "eu",
          "promptui": "consent-popup"
        }
      }
    }
  </script>

  <div id="consent-popup" class="consent-popup">
    <div>
      <a href="#" on="tap:consent.dismiss" class="close-button" role="button" tabindex="0">✕</a>
      <p>
        we use cookies to understand how you use our site and to improve your experience. by continuing to use our site, you accept our <a href="https://policies.google.com/technologies/cookies">use of cookies</a> and <a href="https://policies.google.com/privacy">privacy policy</a>.
      </p>
      <a class="ampstart-btn ampstart-btn-secondary caps text-decoration-none inline-block" on="tap:consent.accept" class="button text-label" role="button" tabindex="1">ok</a>
    </div>
  </div>
</amp-consent>
 {{/includesgeo}}
 {{/includesconsent}}
```

templates/preview.html
```diff
@@ -44,6 +45,7 @@ <h1 id="abe-preview-header-title" class="h3 center truncate caps">Demo: {{subHea
     {{^document.isAmpStory}}
     {{> analytics.html}}
     {{> serviceworker.html}}
+    {{> consent.html}}
```
sebastianbenz: Can you add the amp-story-user consent component in here:

```
{{#isAmpStoryTag}}       
{{> analytics.html}}
{{/isAmpStoryTag}}
```

templates/head.html
morsssss: I wouId have thought that would cause problems for pages that just used <amp-geo>.  But I've done it!

templates/consent.html
morsssss: done!

templates/preview.html
morsssss: Sure! Thanks; I wasn't sure what that was.

templates/head.html
```diff
@@ -11,6 +11,13 @@
 {{^includesIframe}}
 <script async custom-element="amp-iframe" src="https://cdn.ampproject.org/v0/amp-iframe-0.1.js"></script>
 {{/includesIframe}}
+{{^includesGeo}}
+{{^includesConsent}}
+<script async custom-element="amp-geo" src="https://cdn.ampproject.org/v0/amp-geo-0.1.js"></script>    
+<script async custom-element="amp-consent" src="https://cdn.ampproject.org/v0/amp-consent-0.1.js"></script>
+{{/includesConsent}}
+{{/includesGeo}}
+
```
morsssss: Sorry, @sebastianbenz  - when I do it this way, the metadata doesn't get included properly.  To be exact, the <link rel="canonical"> doesn't get added.

morsssss: Are you sure it's ok to nest those template conditionals?

templates/index.html
ericlindley-g: I would say that if possible we should have the pages validate and have the notice site-wide. Is there an option that does both? In the sample can we just point to code that actually implements the consent for the page, rather than a faked example?

templates/head.html
morsssss: Guess not:
```
Message:
    Nesting error: includesGeo vs. includesConsent
Stack:
Error: Nesting error: includesGeo vs. includesConsent
```

morsssss: Um, now it works.  Never mind ;-)

morsssss:
That's an interesting idea - I'll leave it to sbenz to say whether he wants
to have the samples rewritten in this way.

It would take a bit of time.  Would it make sense to unleash this cookie
notice on the rest of the site while this is being discussed - or would you
prefer to wait?

ericlindley-g:
Sounds good — I assume it's better to have the consent notifications on every other page than not, so I personally certainly wouldn't block on this.

sebastianbenz:
+1 better most of the pages than nothing.

sebastianbenz:
Can you take a look at the validation errors?

sebastianbenz:
friendly ping "-)

morsssss:
Yes, sorry! I'm traveling and haven't gotten the chance to look at this.
Hopefully soon....

morsssss:
Ok, I'm back and looking at this now.  There's quite a lot here.  Looks like stories require special treatment.  And in some cases the CSS is now just a bit too big....

morsssss:
@sebastianbenz , on pages like /stories/introduction/amp_story_hello_world/index.html, the `<amp-consent>` JS seems to be getting magically removed during the build process.  Currently `head.html` contains this:

```html
{{^includesGeo}}
{{^includesConsent}}
<script async custom-element="amp-geo" src="https://cdn.ampproject.org/v0/amp-geo-0.1.js"></script>    
<script async custom-element="amp-consent" src="https://cdn.ampproject.org/v0/amp-consent-0.1.js"></script>
{{/includesConsent}}
{{/includesGeo}}
```

If I put anything else in that HTML - say, a `<meta>` tag - it gets included. But the `amp-consent` bit gets replaced by a blank line.

I thought I'd defer to your expertise and experience here :)

morsssss:
More of the validation errors are occurring on pages like `stories/introduction/amp_story_hello_world/preview/embed/index.html`, which I believe only appear in the wild in iframes on pages like `https://ampbyexample.com/stories/introduction/amp_story_hello_world/`.

I haven't found the process by which these embed pages get built and by which the unnecessary <amp-consent> gets added.  Can you perhaps explain?  I could at least make these validate.  Or let me know if I'm missing something else.

morsssss:
And, finally... on pages like `samples_templates/product_page/index.html`, the CSS is now just a bit too large.  (On that page, it's 50919 bytes.)

Looks like `example.html` contains two additional CSS files: 

```
      {{> css/example.css}}
      {{{exampleStyles}}}
```

it also appears that the extra CSS is not minified... if I'm right here?  From `gulpfile.js`:

```javascript
const sampleTemplates = Templates.get(config.templates.root, /* minify */ false,
    '<% %>');
```

Is it ok to minify that extra CSS?  Or do you recommend another remedy here?

sebastianbenz:
Oh dear. Will slowly work my way through your comments.

sebastianbenz:
1. You have to update templates/preview.html to insert consent at the right position. It also needs to be a new template for amp-story-consent:

```
  <body>
    {{{elementsAfterBody}}}
    {{^isEmbed}}
      <div id="abe-preview-header" class="ampstart-headerbar fixed flex justify-start items-center top-0 left-0 right-0 pl2 pr4">
        <a href=".." class="caps text-decoration-none">
          <i class="align-middle inline-block mr1"><svg fill="#000000" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
              <path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z"/>
              <path d="M0 0h24v24H0z" fill="none"/>
            </svg></i>
        </a>
        <h1 id="abe-preview-header-title" class="h3 center truncate caps">Demo: {{subHeading}}</h1>
      </div>
      <div id="abe-preview">
        <div class="ampstart-card m1">
          {{#metadata.experiments.length}}
          <div class="p1">
            {{> experiment.html}}
          </div>
          {{/metadata.experiments.length}}
          {{/isEmbed}}
          {{#document.sections}}
          {{{preview}}}
          {{#isAmpStoryTag}}
          {{> analytics.html}}
          {{> story-consent.html}}
          {{/isAmpStoryTag}}
          {{/document.sections}}
          {{^isEmbed}}
        </div>
      </div>
      {{/isEmbed}}
      {{^document.isAmpStory}}
      {{> analytics.html}}
      {{> consent.html}}
      {{> serviceworker.html}}
      {{/document.isAmpStory}}
  </body>
```

sebastianbenz:
2. CSS should now be minified: https://github.com/ampproject/amp-by-example/pull/1489 PTAL

morsssss:
@sebastianbenz , just merged your changes into my fork.
In the morning I'll look for a few minutes to see if I can find a neat way of fixing the problem you mention in #3 above. If I don't spot one, I believe I do see the mechanism for skipping validation:

```javascript
"metadata": {
  "skipValidation": "true"
}
```

morsssss:
@sebastianbenz , looking at this further, I don't get why `replaceAmpStoryRuntime()` gets run on `/stories/introduction/amp_story_hello_world/` in the first place.   It's there to remove certain tags from AMP Stories.  But the page that contains this example is not a story.

morsssss:
I'm looking now to fix the cases where a story embed appears as its own page, as in `https://ampbyexample.com/stories/introduction/amp_story_hello_world/preview/embed/`. There, I need to use `<amp-story-consent>` instead of `<amp-story>`, and both `<amp-geo>` and `<amp-story-consent>` need to be direct children of `<amp-story>`.

I imagine that to do this I'll need to change `generateStoryPreviewEmbed()` so that it includes my `consent-story.html` template inside `<amp-story>`.  Is that right?  Or is there an easier way?

morsssss:
MSTM (makes sense to me)!

Did these things, committed, and I see the build is still generating "CSS > 50000 bytes" errors on these two pages:

```
samples_templates/product_page/index.html
samples_templates/product_page/embed/index.html
```

Strangely, I don't see this error on my local machine.  In fact, for me, the CSS on the first of those pages is a paltry 45833 bytes.

Any idea what's going on here?  Is there a way to see the files Travis has built?

morsssss:
just wanted to ping this back up... ping!

sebastianbenz:
I could reproduce it. You can fix it by replacing the content of `templates/css/product.css` with this version: https://gist.github.com/sebastianbenz/aa0d87bd2a3f10f3c880bd6d5764a88c

sebastianbenz:
Whoop Whoop!

sebastianbenz:
Too early - afaik you only need to disable validation for the story-consent sample. All other story samples should validate.

sebastianbenz:
The build passes, but you can you try re-enabling validation for the stories (except amp-story-consent).

morsssss:
Sure.  Reenabled validation for two stories, as a test.

I also removed "us", which I don't think I'd ever left in there inadvertently before...

morsssss:
Yeah, sadly, the story embed pages don't work because they need `<amp-story-consent>`, which brings us back to the build issues mentioned above.

morsssss:
Woo!
Thanks for your persistence 😎

PR #1448, opened 2018-08-08
Add <amp-orientation-observer> demo

nainar:


src/50_Visual_Effects/Basics_of_orientation_effects.html
```diff
@@ -0,0 +1,304 @@
...
+      height: 500px;
+      margin: auto;
+      overflow: hidden;
+    }
+
+    #ipsum {
+      background-image: linear-gradient(white, deeppink);
+    }
+
+    amp-img img {
```
sebastianbenz: Can you make this rule more specific, otherwise it'll mess with the generated page.

src/50_Visual_Effects/Basics_of_orientation_effects.html
```diff
@@ -0,0 +1,304 @@
+<!--
```
sebastianbenz: Does it make sense to enable a standalone preview?

```
<!---{
    "preview": "default"
}--->
```

You can separate samples there by adding headings in between that don't have a comment (and hence won't be shown in the doc view).

src/50_Visual_Effects/Basics_of_orientation_effects.html
```diff
@@ -0,0 +1,304 @@
...
+  ## Tilt based translation animations
+
+  Let's create a tilt based animation where the clock's hand moves 
+  as you tilt the device along a specific axis.
+  -->
+
+  <!--
+  The first animation is along the `beta` axis, to see this in action rotate your phone along the y axis. 
+
+  Here you can see that we are using the `beta` event to progress through the animation.
```
sebastianbenz: Can you merge the three last sentences into a single paragraph? I think looks a bit weird that every sentence is a separate paragraph.

src/50_Visual_Effects/Basics_of_orientation_effects.html
```diff
@@ -0,0 +1,304 @@
...
+        }
+      ]
+    }
+    </script>
+  </amp-animation>
+
+  <!--
+  The third animation is along the `alpha` axis, to see this in action rotate your phone along the horizontal axis:
+
+  Here you can see that we are using the `alpha` event to progress through the animation.
```
sebastianbenz: Is this sentence still needed after the previous one?

src/50_Visual_Effects/Basics_of_orientation_effects.html
```diff
@@ -0,0 +1,304 @@
...
+  - **`on:beta`**: This event is triggered by orientation observer as deivce orientation 
+  along the `beta` axis changes. The event provides a percentage value
+  (decimal between 0 and 1) representing how far the device is within the `beta` axis.
+  - **`anim.seekTo(percent=event.percent)`**: We will define an `amp-animation`
+  that will do the spinning in the next step, here we are coupling `amp-orientation-observer` and
+  `amp-animation` by triggering a `seekTo` action on the animation as `beta` events occurs.
+  This is how we specify that we like to progress through the animation
+  timeline as the device moves along the `beta` axis.
+  - **`beta-range`**: Each axis has a range of degrees it tracks. 
+  For the `beta` axis the default range is -180 to 180 degrees.
```
sebastianbenz: Does it make sense to add something like: "In this sample we set the beta range to `0 180` as the text cannot scroll beyond it's first paragraph." ?

sebastianbenz:
forgot: and can you please fix the htmllint errors:

```
Message:
    8 errors found in /home/travis/build/ampproject/amp-by-example/src/50_Visual_Effects/Basics_of_orientation_effects.html
[L187:C33] The value of attribute [ width ] must be in double quotes. (attr-value-double-quotes)
[L187:C41] The value of attribute [ height ] must be in double quotes. (attr-value-double-quotes)
[L223:C7] The id value [ clock-scene-1 ] must be unique. (id-unique)
[L228:C33] The value of attribute [ width ] must be in double quotes. (attr-value-double-quotes)
[L228:C41] The value of attribute [ height ] must be in double quotes. (attr-value-double-quotes)
[L265:C7] The id value [ clock-scene-1 ] must be unique. (id-unique)
[L271:C33] The value of attribute [ width ] must be in double quotes. (attr-value-double-quotes)
[L271:C41] The value of attribute [ height ] must be in double quotes. (attr-value-double-quotes)
[22:45:51] 'validate' errored after 52 s
[22:45:51] Error in plugin 'run-sequence(lint:html)'
Error
```

src/50_Visual_Effects/Basics_of_orientation_effects.html
nainar: Removed.

src/50_Visual_Effects/Basics_of_orientation_effects.html
nainar: Done.

src/50_Visual_Effects/Basics_of_orientation_effects.html
nainar: Done.

src/50_Visual_Effects/Basics_of_orientation_effects.html
nainar: It seems like the "View Demos" option doesn't show up on mobile devices and since this component will be mainly used on Mobile devices, I don't think a standalone preview adds much value.

sebastianbenz: View Demo does show (it's a bug if not). Only the playground is disabled on mobile.

nainar:
@sebastianbenz could you take a look at the changes made after your last review?

sebastianbenz:
Did you see my comment regarding previews from 4 days ago? 

> View Demo does show on mobile (it's a bug if not). Only the playground is disabled on mobile.

src/50_Visual_Effects/Basics_of_orientation_effects.html
nainar: Oops. Sorry I missed this comment. 

I remember running a search for this code and didn't see the `View Demo` button in the Chrome device emulator. I don't remember what page that was. Can't repro that now. Added the preview mode.

nainar:
@sebastianbenz sorry for missing your comment! Made the changes!

sebastianbenz:
Thanks! The validation fails because it doesn't find the amp-orientation-observer component. Does it need to be a direct child of body?

nainar:
Nope. There is no such constraint.

nainar:
Fixed the validation issues in the latest push. The generated files didn't have the `<amp-orientation-observer>` because they were being obfuscated by a previous tag.

PR #1456, opened 2018-08-14
Update the documentation around amp-bind.

sparhami:
- Add a link to what events you can listen to
- Describe the state object a little more explicitly

Context for this update: https://github.com/ampproject/amphtml/issues/17471. Seems that some clarity around what the events you can use with `setState` and how they differ would be useful.


kul3r4:
Thanks!

PR #1471, opened 2018-08-23
Make all stories self-canonical

ithinkihaveacat:
This is an attempt to arrange for all the AMP Story examples to have a self-referential `<link rel=canonical>` [as required](https://www.ampproject.org/docs/reference/components/amp-story#required-markup-for-amp-story).

That is, a URL such as `https://ampbyexample.com/stories/introduction/amp_story_hello_world/preview/embed/` should have a `<link rel=canonical>` of

```html
<link rel=canonical href=https://ampbyexample.com/stories/introduction/amp_story_hello_world/preview/embed/>
```


sebastianbenz:
Hm, not sure about this approach as it means that the documented samples no longer appear in search (and the stories don't link back to the documented sample). 

How about using a relative path instead? `<link rel=canonical href=".">`? [Search](<link rel=canonical href=) supports it.

ithinkihaveacat:
How about this approach? #1490

Issue #1497, opened 2018-09-03
AMP Story previews missing


ithinkihaveacat:
https://ampbyexample.com/stories/introduction/amp_story_hello_world/preview/embed/ 404s. It previously existed, and is the "view story" link from https://ampbyexample.com/stories/introduction/amp_story_hello_world/.

Not sure what happened here. Some of the code around this was fiddled with recently (in #1490 and #1489) but everything works as expected when building from a fresh checkout locally.

sebastianbenz:
https://github.com/ampproject/amp-by-example/pull/1489 is the culprit. Will fix.

Issue #1485, opened 2018-08-29
amp-story-consent iframe not loading


morsssss:
I saw this here:
`https://ampbyexample.com/stories/user_consent/story_user_consent/`

I think the problem is that the domain of the iframe is the same as that of the main page, which violates [this policy](https://github.com/ampproject/amphtml/blob/master/spec/amp-iframe-origin-policy.md).

Thus the sample can't be seen.

sebastianbenz:
It's OK to embed iframes from the same origin:

"An amp-iframe must not be in the same origin as the container unless they
do not allow allow-same-origin in the sandbox attribute.
<https://www.ampproject.org/docs/reference/components/amp-iframe#sandbox>"

However, allow-same-origin is required for iframes to be able to access
cookies or local storage, which is required by amp-consent. Who was I to
believe that stories won't need cookie or local storage access....

morsssss:
eek!

Issue #1433, opened 2018-08-03
Monetization stories cover page displays the title and text twice


gmajoulet:
Monetization stories cover page displays the title and text twice:

![image](https://user-images.githubusercontent.com/1492044/43651032-49c0395e-970f-11e8-978f-da7534808315.png)

It is due to [`poster0.png`](https://github.com/ampproject/amp-by-example/blob/master/src/img/poster0.png) having text on the image, when the `amp-story-page` itself also adds text.

PR #1498, opened 2018-09-03
Story fixes

sebastianbenz:


sebastianbenz:
This will fix #1497

sebastianbenz:
...and it also fixes #1485

sebastianbenz:
Sorry for putting so much into a single PR.

sebastianbenz:
Oh - and also fixes #1433

tasks/compile-example.js
```diff
@@ -297,6 +297,7 @@ module.exports = function(config, indexPath, updateTimestamp) {
       if (document.isAmpStory) {
         generateStoryPreviewEmbed(stream, example, args, {
```
ithinkihaveacat: The `generateStoryPreviewEmbed` function seems to have only 2 arguments? Am I reading that right?

templates/stories/page-switch.html
```diff
@@ -1,22 +1,72 @@
...
+// Enable story experiment
+enableExperiment();
+
+function enableExperiment() {
+  var expCookie = getCookie(COOKIE_AMP_EXP);
+  if (expCookie) {
+    if (expCookie.indexOf(AMP_STORY) != -1) {
+      return;
+    } else {
+      expCookie += '%2C' + AMP_STORY;
```
ithinkihaveacat: Scattering some `encodeURIComponent()` and `decodeURIComponent()` through the `setCookie()` and `getCookie()` functions might eliminate the need for `%2C`. Or replace `%2C` with `encodeURIComponent(' ')`. Depends how "helpful" the helper functions are supposed to be I guess.

templates/stories/page-switch.html
```diff
@@ -1,22 +1,72 @@
...
     }
-  });
-  story.dispatchEvent(event);
-});
+  } else {
+    expCookie = AMP_STORY;
+  }
+  setCookie(COOKIE_AMP_EXP, expCookie, 356);
```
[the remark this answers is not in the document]
sebastianbenz: extracted a constant, one year is as good as any other value I assume...

tasks/compile-example.js
sebastianbenz: fixed

templates/stories/page-switch.html
sebastianbenz: they're only used in a single place, so I didn't bother. But you're right - fixed it.

sebastianbenz:
Adressed the comments. Changed also the way the api is served on localhost. Previews should now work properly there as well. PTAL

ithinkihaveacat:
Tests locally just fine & LGTM!

sebastianbenz:
Thanks!

PR #1508, opened 2018-09-06
Replace text node in SVG logo with path

fstanis:
The current logo doesn't display correctly when the Montserrat font is not present in the system:

screenshot_2018-09-06_22-23-44

This PR fixes that by replacing the text node with a path in the SVG.


sebastianbenz:
Makes sense. Thanks!

Issue #1457, opened 2018-08-14
AMP Playground should automagically import component scripts


tomayac:
Currently it's a bit burdensome to include component scripts manually. Turns out, @garanj has built a Gulp plugin called [amphtml-autoscript](https://github.com/garanj/amphtml-autoscript) that does it automagically. Can we incorporate this (or its logic) into the Playground?

sebastianbenz:
Yes please 😀

garanj:
Hi @sebastianbenz ,

Thanks for your advice on start points. I had a go at building and it seems pretty straightforward (even for me!) to hook into the necessary events for a new module. Works nicely.

I've two questions, and would welcome some guidance:

1.  I was thinking of subscribing to `Validator.EVENT_NEW_VALIDATION_RESULT`, and making use of the fact that validation results were already available. If not, possibly `Editor.EVENT_INPUT_NEW`, but I think really the former.

    In either case, using `editor.setSource` can cause a loop, as it fires `Editor. EVENT_INPUT_NEW`, which triggers `editorUpdateListener`, validation etc.

    I've a couple of options for how to set the text of the editor without causing an event loop. Any preferences on how to do it? One might be a further method in `Editor`, or adding an optional parameter to `setSource` indicating whether or not to fire events?

2.  I've used your snippet for getting a list of components, and added in getting version numbers. Works great. However, I'm thinking that the client should really not be making this request to GitHub direct, instead either:

    1.  An endpoint within Playground that provides the list (and periodically refreshes this with GH)
    2.  A list formed as part of the build process

    Do you have a view on either approach? I think GH would probably limit requests, if coming from clients, and this could be a problem, and additionally, the response is needlessly large compared to the list that is created from it. I favour approach 1 above, and wonder if you have views on where this should sit?

Apologies if these seem obvious questions!

kul3r4:
@sebastianbenz can we close this?

sebastianbenz:
Yes!

Chiara Chiappini <notifications@github.com> schrieb am Mi. 26. Sep. 2018 um
18:45:

PR #1513, opened 2018-09-11
Importing component scripts (FR #1457)

garanj:
A start point for the FR from @tomayac #1457.

In summary, this PR adds the feature to automatically insert `<script>` elements for AMP components to the `<head>` element, when the user adds code requiring that component.

## Frontend

The frontend adds a module `auto-importer` which listens for events from the AMP Validator. Where the validation result is a failure, the errors are filtered to determine whether any correspond to missing imports.

The resulting missing imports list is cross-referenced with the parsed state of the `<head>` element (as the former can sometimes report missing elements that are actually present), to arrive at a list of elements to insert.

The logic for insertion is as follows:

*   If the AMP engine `<script>` element already exists, insert any missing `<script>` component elements directly beneath it.
*   If the AMP engine does not exist, then insert both the AMP engine and required component elements as the last elements in the `<head>` element.

Currently more complex scenarios seeking to group together existing elements, sort them etc are not dealt with, but the output of parsing the `<head>` element does contain what would be required for this.

## Backend

The frontend, on initialization, obtains a mapping of AMP component to version number from the backend, using the endpoint  `/playground/amp-component-versions`

This mapping is formed from the GitHub AMP extensions directory structure (thank you @sebastianbenz).

The mapping returned by the endpoint is stored in Memcache, backed by Datastore. Requests that cause a stale list to be served, trigger an update which is achieved through a named TaskQueue entry.

## Demo

An demo can be seen [in this video](https://www.youtube.com/watch?v=IaUxHTxAauY)


garanj:
Looks like Go linting fails, yet running `gofmt -l .` now shows no files to change. I guess this is what happens to amateurs like me...

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+	"google.golang.org/appengine/log"
+	"google.golang.org/appengine/memcache"
+	"google.golang.org/appengine/taskqueue"
+	"google.golang.org/appengine/urlfetch"
 )
 
+const (
+	COMPONENTS_URL                 = "https://api.github.com/repos/ampproject/amphtml/git/trees/master?recursive=1"
+	COMPONENTS_MEMCACHE_KEY        = "amp-components-list"
+	COMPONENTS_UPDATE_FREQ_SECONDS = 600
```
sebastianbenz: 10 minutes feels very frequent. A daily update should be fine (theoretically weekly in sync with the AMP runtime releases).

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+
+type AmpComponentsList struct {
+	Id         int64
+	Timestamp  int
+	Components []byte
+}
+
 func InitPlayground() {
 	http.HandleFunc("/playground/fetch", handler)
+	http.HandleFunc("/playground/amp-component-versions", components)
```
sebastianbenz: nit: extract the playground path prefix `/playground` into a constant

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+}
+
+func updateComponents(ctx context.Context) (bool, error) {
+	log.Infof(ctx, "Fetching components list from GitHub")
+	list, err := fetchComponents(ctx)
+	if err != nil {
+		return false, err
+	}
+	addComponentsToStores(ctx, list)
+	return true, nil
```
sebastianbenz: Remove the bool return value as it's not needed.

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+}
+
+func getComponentsFromDataStore(ctx context.Context) (*AmpComponentsList, error) {
```
sebastianbenz: nit: s/get/fetch 

get implies returning a value that's already there, whereas fetch indicates that it's potentially a long running operation.

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+	log.Infof(ctx, "Retrieving components from datastore")
+	var list AmpComponentsList
```
sebastianbenz: nit: s/list/components

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+		COMPONENTS_DATASTORE_KEY_ID, nil)
+	err := datastore.Get(ctx, key, &list)
+	if err != nil {
+		log.Infof(ctx, "Error retrieving components from datastore")
+		return nil, err
+	}
+	return &list, err
+}
+
+func getComponentsFromMemCache(ctx context.Context) (*AmpComponentsList, error) {
```
sebastianbenz: same as above: s/get/fetch

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+	log.Infof(ctx, "Retrieving components from memcache")
+	var item AmpComponentsList
```
sebastianbenz: s/item/components

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+			log.Infof(ctx, "Setting datastore components value to memcache")
+			memcacheItem := &memcache.Item{
+				Key:    COMPONENTS_MEMCACHE_KEY,
+				Object: dsItem,
+			}
+			memcache.Gob.Set(ctx, memcacheItem)
+		}
+	} else if err != nil {
+		return nil, err
+	}
```
sebastianbenz: I think you can remove the whole block:

```
 else if err != nil {
 		return nil, err
 	}
```

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+		}
+		memcache.Gob.Set(ctx, item)
+	}
+}
+
+func extractComponents(g *GitHubApiResponse) ([]byte, error) {
+	vers := make(map[string]string)
+	tree := g.Tree
+	for _, blob := range tree {
+		groups := componentRegex.FindStringSubmatch(blob.Path)
```
sebastianbenz: suggestion: extract the parsing into a separate method. Then you can avoid the deep nesting:

```
func parseAndAddComponent(blob, componentsWithVersion) {
                groups := componentRegex.FindStringSubmatch(blob.Path)
 		if len(groups) != 3 {
                     return
                }
 	       component := groups[1]
 		if strings.HasSuffix(component, "-impl") {
                   return
                }
 		ver := groups[2]
 		elem, ok := vers[component]
 		if !ok || ver > elem {
 		  componentsWithVersion[component] = ver
 		}
 	}
}
```

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+}
+
+func getComponentsFromDataStore(ctx context.Context) (*AmpComponentsList, error) {
+	log.Infof(ctx, "Retrieving components from datastore")
+	var list AmpComponentsList
+	key := datastore.NewKey(ctx, "AmpComponentsList", "",
+		COMPONENTS_DATASTORE_KEY_ID, nil)
+	err := datastore.Get(ctx, key, &list)
+	if err != nil {
+		log.Infof(ctx, "Error retrieving components from datastore")
```
sebastianbenz: How about calling `updateComponents` as a fallback? `updateComponents` could return the list of components which you could use here.

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+	ctx := appengine.NewContext(r)
+	_, err := updateComponents(ctx)
+	if err != nil {
+		http.Error(w, "Server error", http.StatusInternalServerError)
+		return
+	}
+	w.Write([]byte("Updated components"))
+}
+
+func updateComponents(ctx context.Context) (bool, error) {
```
sebastianbenz: suggestion rename to fetchAndUpdateComponents

playground/serve.go
```diff
@@ -16,27 +16,237 @@
...
+	}
+
+	if curTime-timestamp > COMPONENTS_UPDATE_FREQ_SECONDS {
+		createTaskQueueUpdate(ctx, timestamp)
+	}
+	return latest
+}
+
+func createTaskQueueUpdate(ctx context.Context, timestamp int) {
+	t := taskqueue.NewPOSTTask("/playground/amp-component-versions-task",
```
sebastianbenz: extract the path into a constant

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+        // The action taken to insert any elements to fix the report of missing
+        // tags is determined by both a combination of looking at the list of
+        // missing elements and also the current state of the <head> tag.
+        this._insertMissingElements(missing, existing);
+      }
+    }
+  }
+
+  _fetchComponents() {
+    if (!Object.keys(COMPONENT_VERSION_MAP).length) {
```
sebastianbenz: Can you wrap this into a `requestIdleCallback` to avoid firing the request during initial rendering (it's already polyfilled).

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+
+class AutoImporter {
+
+  constructor(editor) {
+    this.editor = editor;
+    this._fetchComponents();
+  }
+
+  update(validationResult) {
+    if (validationResult.status === 'FAIL') {
```
sebastianbenz: nit: use an early exit to avoid nesting

```
if (validationResult.status !== 'FAIL') {
  return
}
```

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+                console.log(`Warning: Unknown AMP component : ${tagName}`);
+              }
+            }
+            break;
+          default:
+            // no default
+        }
+      }
+    }
+    return {
```
sebastianbenz: suggestion: create the return obj at the beginning of the method

```
const missingElements = {
  components: {},
  ampRuntime: false
}
```

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+   *
+   * @return {{
+   *   missingTags: Object,<string, number>,
+   *   missingBaseScriptTag: boolean
+   * }}
+   *
+   * missingTags: The keys of this dictionary are those missing AMP components.
+   * missingBaseScriptTag: True if the AMP engine <script> tag is not present.
+   */
+  _getMissingElements(validationResult) {
```
sebastianbenz: suggestion: rename s/_get/_parse

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * Inserts missing <script> tags into the document, based on results from the
+   * AMP validator, and results from inspecting the current <head> structure.
+   *
+   * @param {!Object} missing The results from {@code _getMissingElements()}.
+   * @param {!Object} existing The results from {@code _parseHeadTag()}.
+   */
+  _insertMissingElements(missing, existing) {
+    const pos = existing.baseScriptTagEnd || existing.lastTag;
+    if (pos) {
```
sebastianbenz: again: early exit `if(!pos) return` to avoid nesting

playground/src/auto-importer/auto-importer.js
```diff
@@ -0,0 +1,245 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import CodeMirror from 'codemirror';
+
+const MISSING_ENGINE_PARAM = 'amphtml engine v0.js script';
+const AMP_BASE_URL = '"https://cdn.ampproject.org/v0.js"';
```
sebastianbenz: How hard would it be to also support auto-adding the runtime for amp ads?

garanj:
Thanks @sebastianbenz for the comments! I will work through them!

On the `<amp-carousel>` example, this of course comes straight back from the validator, as per on the command-line (`example.html` containing your HTML):

```
$ amphtml-validator example.html
example.html:19:2 The mandatory attribute 'lightbox' is missing in tag 'AMP-CAROUSEL [lightbox]'. (see https://www.ampproject.org/docs/reference/components/amp-carousel)
example.html:19:2 The tag 'AMP-CAROUSEL [lightbox]' requires including the 'amp-carousel' extension JavaScript. (see https://www.ampproject.org/docs/reference/components/amp-carousel)
example.html:19:2 The tag 'AMP-CAROUSEL [lightbox]' requires including the 'amp-lightbox-gallery' extension JavaScript. (see https://www.ampproject.org/docs/reference/components/amp-carousel)
example.html:21:2 The tag 'amp-iframe' requires including the 'amp-iframe' extension JavaScript. (see https://www.ampproject.org/docs/reference/components/amp-iframe)
```

From what I can see in the validator spec for `amp-carousel`, it looks like the wrong `TagSpec` is being identified for the `amp-carousel`: the [lightbox mode one](https://github.com/ampproject/amphtml/blob/master/extensions/amp-carousel/validator-amp-carousel.protoascii#L79) compared with the [plain one](https://github.com/ampproject/amphtml/blob/master/extensions/amp-carousel/validator-amp-carousel.protoascii#L32).

I can open a bug against the validator for this one? I think that's probably the appropriate place to try to fix this? What do you reckon? (I ask as I am very new to this....!)

garanj:
Added: https://github.com/ampproject/amphtml/issues/18091 for the amp-carousel issue.

playground/src/auto-importer/auto-importer.js
sebastianbenz: Another thing: can you move this into a separate module, e.g.: `/amp/components-provider.js` and pass an ComponentProvider instance via the constructor. 

Background: we can use the component list for auto completion, which is configured [here](https://github.com/ampproject/amp-by-example/blob/master/playground/src/editor/editor.js#L89). `amp-img` is currently only a placeholder...

playground/src/auto-importer/auto-importer.js
sebastianbenz: Sure! But only if it doesn't take too much effort.

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/serve.go
garanj: Added

playground/serve.go
garanj: Done

playground/serve.go
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Added ComponentsProvider and used requestIdleCallback.

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: I've added this.

garanj:
Hi Sebastian, I think I've made a stab at making the changes. Thank you for the review

playground/src/components-provider/components-provider.js
```diff
@@ -0,0 +1,45 @@
...
+class ComponentsProvider {
+
+  constructor(win) {
+    this.componentsMap = {};
+    this.win = win;
+    this.win.requestIdleCallback(() => this._fetchComponents());
+  }
+
+  getComponents() {
+    return this.componentsMap;
```
sebastianbenz: Can you make this return a promise?

playground/src/components-provider/components-provider.js
```diff
@@ -0,0 +1,45 @@
...
+
+class ComponentsProvider {
+
+  constructor(win) {
+    this.componentsMap = {};
+    this.win = win;
+    this.win.requestIdleCallback(() => this._fetchComponents());
+  }
+
+  getComponents() {
```
sebastianbenz: suggestion: rename to simply get as the class has only one functionality.

playground/serve.go
```diff
@@ -16,27 +16,243 @@
...
+	"google.golang.org/appengine/log"
+	"google.golang.org/appengine/memcache"
+	"google.golang.org/appengine/taskqueue"
+	"google.golang.org/appengine/urlfetch"
 )
 
+const (
+	COMPONENTS_URL                 = "https://api.github.com/repos/ampproject/amphtml/git/trees/master?recursive=1"
+	COMPONENTS_MEMCACHE_KEY        = "amp-components-list"
+	COMPONENTS_UPDATE_FREQ_SECONDS = 86400
```
sebastianbenz: nit: add a comment

```
COMPONENTS_UPDATE_FREQ_SECONDS = 86400 // one day
```

playground/src/components-provider/components-provider.js
garanj: Yes, sorry, this was rubbish.

playground/src/components-provider/components-provider.js
garanj: Done

garanj:
Apologies for that Sebastian, I believe it's fixed now. Thanks for the suggestion, my original approach was bad.

playground/src/auto-importer/auto-importer.js
```diff
@@ -40,28 +40,28 @@ class AutoImporter {
...
-        || missing.missingBaseScriptTag) {
-      const existing = this._parseHeadTag();
-      // The action taken to insert any elements to fix the report of missing
-      // tags is determined by both a combination of looking at the list of
-      // missing elements and also the current state of the <head> tag.
-      this._insertMissingElements(missing, existing);
-    }
+    this.componentsMap.then((components) => {
+      this.components = this.components || components;
+      if (validationResult.status !== 'FAIL') {
```
sebastianbenz: Move this to the beginning of the method

playground/src/auto-importer/auto-importer.js
```diff
@@ -40,28 +40,28 @@ class AutoImporter {
...
-    if (Object.keys(missing.missingTags).length
-        || missing.missingBaseScriptTag) {
-      const existing = this._parseHeadTag();
-      // The action taken to insert any elements to fix the report of missing
-      // tags is determined by both a combination of looking at the list of
-      // missing elements and also the current state of the <head> tag.
-      this._insertMissingElements(missing, existing);
-    }
+    this.componentsMap.then((components) => {
+      this.components = this.components || components;
```
sebastianbenz: I'd prefer if you'd make the state handling more explicit by passing the components variable to _insertMissingElement instead of assigning it to a field.

playground/src/components-provider/components-provider.js
```diff
@@ -21,25 +21,23 @@ export function createComponentsProvider() {
...
+    this.componentsMap = new Promise((resolve, reject) => {
+      this.win.requestIdleCallback(() => {
+        let request = new Request(COMPONENTS_URL, {
+          headers: new Headers({'x-requested-by': 'playground'})
+        });
+        fetch(request)
+            .then((r) => r.json())
+            .then((data) => {
+              resolve(data);
+            });
```
sebastianbenz: What happens if the fetch fails?

garanj: How about a number of backoff retries, then ultimately if there is no success after n tries, assign the the components map to just be empty? What do you think?

sebastianbenz: As it's not a critical feature I think it's fine to simply fail and fallback to an empty map at first. We can add a retry later.

playground/src/auto-importer/auto-importer.js
```diff
@@ -40,28 +40,28 @@ class AutoImporter {
 
   constructor(componentsProvider, editor) {
     this.componentsProvider = componentsProvider;
-    this.componentsMap = {};
+    this.componentsMap = componentsProvider.get();
```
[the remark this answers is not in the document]
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/auto-importer/auto-importer.js
garanj: Done

playground/src/components-provider/components-provider.js
garanj: Sure. Done

playground/serve.go
```diff
@@ -16,27 +16,243 @@
...
+		components.Timestamp = int(time.Now().Unix())
+	}
+	return &components, err
+}
+
+func fetchComponentsFromMemCache(ctx context.Context) (*AmpComponentsList, error) {
+	log.Infof(ctx, "Retrieving components from memcache")
+	var components AmpComponentsList
+	_, err := memcache.Gob.Get(ctx, COMPONENTS_MEMCACHE_KEY, &components)
+	if err == memcache.ErrCacheMiss {
```
sebastianbenz: When I deploy the version, the result is always `{}`. It's related to this part as it works when I replace this line with ` if err != nil {`. BTW this is only a problem when I deploy it to app engine, it works locally.

In general can you make sure to handle the errors correctly:

* can you log all errors: `log.Errorf(ctx, "..... %v", err)`
* can you make sure to return a corresponding status code if we could not retrieve the list of components.

garanj:
Regarding adding components too early, one option might be to inspect the token just before the cursor when a validation error takes place using `cm.getTokenAt()`. If this is of type `tag`, then stop the auto-import, as the cursor is within an element tag.

This would mean auto import wouldn't happen til the tag was closed or a space entered to then add attributes.

e.g:

```
const currentTag = this.editor.getTokenAt(this.editor.getCursor());
if (currentTag.type === 'tag') {
  return;
}
```

Thoughts?

sebastianbenz:
This sounds like a perfect solution!

garanj:
I've added:

- Handling of errors from GitHub better, and use of authenticated requests to the API
- Better logging of errors, including passing back any GitHub HTTP error code through to the client
- Delaying auto-import until outside of a tag entry

playground/serve.go
```diff
@@ -16,27 +16,301 @@
...
+			log.Infof(ctx, "Setting datastore components value to memcache")
+			memcacheItem := &memcache.Item{
+				Key:    COMPONENTS_MEMCACHE_KEY,
+				Object: dsComponents,
+			}
+			memcache.Gob.Set(ctx, memcacheItem)
+			return dsComponents, nil
+		}
+	} else if err != nil {
+		return nil, &ComponentsReqError{"Memcache error", http.StatusInternalServerError}
```
sebastianbenz: We should log this error. However, we should also fallback to loading the components in this case. Suggestion: 

```
if err == nil {
  return &components, nil
}

if err == memcache.ErrCacheMiss {
   log.Infof(ctx, "Components not in memcache, retrieving from datastore")
} else {
   log.Errorf(ctx, "Error when retrieving components from memcache: %v, %v", err.Code, err.Message)
}

dsComponents, err := fetchComponentsFromDataStore(ctx)
...
```

playground/serve.go
```diff
@@ -16,27 +16,301 @@
...
+	return components, nil
+}
+
+func fetchComponentsFromDataStore(ctx context.Context) (*AmpComponentsList, *ComponentsReqError) {
+	log.Infof(ctx, "Retrieving components from datastore")
+	var components AmpComponentsList
+	key := datastore.NewKey(ctx, "AmpComponentsList", "ComponentsListKey", 0, nil)
+	err := datastore.Get(ctx, key, &components)
+	if err != nil {
+		log.Warningf(ctx, "Error retrieving components from datastore")
```
sebastianbenz: can you also log the error

playground/serve.go
garanj: Great! Done

playground/serve.go
garanj: Done

garanj:
Should I add some documentation anywhere about use of GitHub API key, and how to set it up in DataStore? Does it fit in the README.md, or is it something that anyone deploying would know already..?

garanj:
I've added a readme about adding the token to Datastore.

sebastianbenz:
Fantastic Thanks!

PR #1519, opened 2018-09-17
Restrict dependency updates to `package.json` and `playground/package.json`

rsimha:


sebastianbenz:
Thanks for looking into this. I'm actually considering disabling renovate bot as it's a pain at the moment. The suggested pull requests are not helpful as they all include breaking changes which require serious effort...

sebastianbenz:
The manual update is the problem as it's non-trivial. I've done alle the straightforward ones.

Issue #1536, opened 2018-09-24
bookend.json has text/plain content type


ithinkihaveacat:
Actual:

```sh
$ curl -sSi https://ampbyexample.com/json/bookend.json | grep -i '^content-type'
content-type: text/plain; charset=utf-8
```

Expected:

```sh
$ curl -sSi https://ampbyexample.com/json/bookend.json | grep -i '^content-type'
content-type: application/json; charset=utf-8
```

The current runtime and cache seems to deal with this just fine (i.e. it's not causing any bugs), but I think it probably should be "fixed" for correctness.

PR #1538, opened 2018-09-26
Move static logic to backend/static.go, fix JSON content type

### LINKED ISSUE #1530: JSON endpoints should return `content-type: application/json`
**Describe the bug**

JSON endpoints used by this site should return a `content-type: application/json` response header.

**To Reproduce**
1. Go to https://ampbyexample.com/json/cart.json
2. Observe the response headers in the devtool
3. See `Content-Type` header being set to `text/plain`.

**Expected behavior**

`Content-Type` response header being set to `application/json`.

fstanis:
Fixes #1530 and #1536

### DIFF
--- a/backend/static.go
+++ b/backend/static.go
@@ -1,0 +1,63 @@
+// Copyright Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


fstanis:
Go already has a few extensions which it maps to MIME types ([list here](https://golang.org/src/mime/type.go?s=1425:1811#L60)), so I think this is the only one we need to manually override.

Come to think of it, it's perhaps better to use `mime.AddExtensionType(".json", "application/json")` than what I've done here, I may change that in another PR.

PR #1543, opened 2018-09-29
Add autocomplete to Playground

fstanis:
Adds autocomplete definitions to CodeMirror based on AMP Validator.

screenshot_2018-09-28_17-46-04


playground/src/editor/editor.js
```diff
@@ -184,4 +187,21 @@ class Editor {
...
+  }
+
+  setRuntime(runtime) {
+    const runtimeToHint = {
+      amphtml: 'amp',
+      amp4ads: 'amp4ads',
+      amp4email: 'amp4email',
+      amp4stories: 'amp'
+    }
+    this.loadHints(amphtmlHints[runtimeToHint[runtime.id]]);
```
sebastianbenz: The `runtime.validator` field includes the correct validation mode per runtime.

fstanis:
Added async loading - PTAL.

playground/src/editor/editor.js
```diff
@@ -184,4 +188,33 @@ class Editor {
...
+  }
+
+  setRuntime(runtime) {
+    const runtimeData = runtimes.get(runtime.id);
+    this.loadHints(runtimeData.validator);
+  }
+
+  fetchHintsData() {
+    return new Promise((resolve, reject) => {
+      window.requestIdleCallback(() => {
```
sebastianbenz: note to myself: I should create a promised version for requestIdleCallback so that we can rewrite this as 

```
runWhenIdle()
     .then( () => fetch(HINTS_URL))
     .then( (response) => response.json() )
```

fstanis: Good idea, can be a one-liner: `runWhenIdle = (options) => new Promise((resolve) => window.requestIdleCallback(resolve, options))`

Would've actually simplified this PR!

sebastianbenz:
I'm not gonna stop you from adding it ;-).

PR #1552, opened 2018-10-08
Publish standalone Document Parser

sebastianbenz:
Make it possible to standalone parse a document. Also
set different runtime types.


sebastianbenz:
yes

PR #1566, opened 2018-10-12
Change structure of redirects.json

fstanis:
Since #1564 made me realize redirects.json has potential to break the whole website if there's a duplicate redirect, I think we should use an object there instead, so ensure this doesn't happen.


sebastianbenz:
This completely went under my radar. Thanks for fixing this!

Issue #1537, opened 2018-09-25
Enhance /playground/ with features to support AMP Story ads 


ithinkihaveacat:
- [ ] Add support for AMP Story ad screen sizes (e.g. full-bleed mobile, full-screen desktop)
- [ ] Create template based on https://support.google.com/admanager/answer/9038178, make it possible to pre-fill the playground either drop-down on left pane, or the main dropdown (whichever is more appropriate, probably main dropdown?)

/cc @jasti

jasti:
Thanks! Probably makes sense to do it in the main dropdown.

PR #1567, opened 2018-10-12
Add desktop and iPhone sizes to playground

ithinkihaveacat:
Adds "iPhone 7" and "Desktop" screen sizes to the AMP for Ads dropdown, and "Desktop" to the AMPHTML dropdown.

screenshot 2018-10-12 at 16 27 06

Fixes part of #1537.

This doesn't actually preserve the ratio when the screen is resized in extreme ways, but after trying for quite a while to fix this I discovered that the existing presets don't do it either.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,41 +1,19 @@
 {
   "name": "amp-by-example",
-  "version": "0.1.5",
+  "version": "0.2.1",
   "lockfileVersion": 1,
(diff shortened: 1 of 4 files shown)


sebastianbenz:
How is your desktop one better than the responsive mode?

sebastianbenz:
Friendly ping

PR #1572, opened 2018-10-18
Adds path to amp-bodymovin-animation sample

### LINKED ISSUE #1569: Validation errors in <amp-bodymovin-animation> example
https://ampbyexample.com/components/amp-bodymovin-animation

This page has three validation errors.

kul3r4:
Partially fixes #1569


kul3r4:
@nainar by looking at the documentation, it seems loop="5" should be allowed but we are getting: The attribute 'loop' in tag 'amp-bodymovin-animation' is set to the invalid value '5'.
Could you advice how to fix it?

PR #1573, opened 2018-10-18
✨Add public linker example

calebcordry:
Adds a few example configurations as well as a small demo.

cc/ @jeffjose 

@lannka Is there any good reason to keep the `testing/*` files around?


src/30_Advanced/Joining_Analytics_Sessions.html
```diff
@@ -19,73 +25,105 @@
...
+          }
+        }
+      </script>
+    </amp-analytics>
+  </div>
+
+  <!-- ## Using AMP Linker with custom analytics -->
   <!--
-  Custom linker
+    If you wish to setup your own configuration, AMP linkeker supports that as well.
```
calebcordry: typo

src/30_Advanced/Joining_Analytics_Sessions.html
```diff
@@ -1,10 +1,16 @@
 <!---{
- "draft": "true"
+ "skipValidation": "true"
```
sebastianbenz: You can use `"hidePreview": true` instead. This will only show the source code listings, but won't add the google analytics tags to the page.

static/linker/demo-page.html
```diff
@@ -0,0 +1,87 @@
+<html ⚡️>
+<head>
+  <meta name="viewport" content="width=device-width,minimum-scale=1,initial-scale=1">
+  <script>
```
sebastianbenz: This confuses me. Can you explain why it's needed?

src/30_Advanced/Joining_Analytics_Sessions.html
calebcordry: Done.

static/linker/demo-page.html
calebcordry: It's not. Good catch. Deleted.

sebastianbenz:
Can you please fix the HTML Linter error:

```
Message:
    1 error found in /home/travis/build/ampproject/amp-by-example/src/30_Advanced/Joining_Analytics_Sessions.html
[L129:C1] Tag must be paired, missing: [ </body> ], start tag match failed [ <body> ] on line 29. (tag-pair)
```

PR #1576, opened 2018-10-22
Adds  amp-state refresh sample

kul3r4:
Related to https://github.com/ampproject/amphtml/issues/18684

### DIFF
--- a/backend/time.go
+++ b/backend/time.go
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	BIND_SAMPLE_PATH = "/" + CATEGORY_COMPONENTS + "/state_refresh/"
+	BIND_SAMPLE_PATH = "/" + CATEGORY_COMPONENTS + "/time/"
 )
 
 func InitStateRefreshSection() {
--- a/backend/time.go
+++ b/backend/time.go
@@ -28,7 +28,5 @@ func InitStateRefreshSection() {
 }
 
 func getTime(w http.ResponseWriter, r *http.Request) {
-	SendJsonResponse(w, map[string]string{
-		"time": time.Now().Format("15:04:05"),
-	})
+	SendJsonResponse(w, time.Now().Format("15:04:05"))
 }
--- a/backend/time.go
+++ b/backend/time.go
@@ -1,0 +1,34 @@
+// Copyright Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


kul3r4:
@choumx @alabiaga We can wait to merge this until https://github.com/ampproject/amphtml/pull/18836 is available on prod

src/20_Components/amp-bind.html
```diff
@@ -259,6 +259,17 @@
...
   </section>
 
+    <!-- ## Refresh state -->
+
+    <!-- `amp-state` supports the `refresh` action. This can be useful in many cases, for example the initial URL loading fails
+       and you want the user to load the state again.
+       For a complete list of actions, see [here](https://www.ampproject.org/docs/interaction_dynamic/amp-actions-and-events).
+       Clicking the button below will refresh and refetch the json in amp-state. -->
+
+    <button class="ampstart-btn caps m1" on="tap:remoteAnimals.refresh">
```
sebastianbenz: It'd be cool to demo this somehow, e.g. have a document that always returns the current time.

src/20_Components/amp-bind.html
```diff
@@ -259,6 +259,17 @@
     </amp-img>
   </section>
 
+    <!-- ## Refresh state -->
+
+    <!-- `amp-state` supports the `refresh` action. This can be useful in many cases, for example the initial URL loading fails
```
sebastianbenz: `This can be useful in many cases, for example the initial URL loading fails and you want the user to load the state again.`

Add: "Another common scenario is live content that needs to be updated, e.g. a sports live score."

src/20_Components/amp-bind.html
kul3r4: ok, will take a look

kul3r4:
@sebastianbenz added an endpoint to return the time, cannot test until the change is available on prod.

src/20_Components/amp-bind.html
```diff
@@ -262,13 +262,19 @@
...
-    <button class="ampstart-btn caps m1" on="tap:remoteAnimals.refresh">
-      Refresh
-    </button>
+    <div>   
+      <amp-state id="time" src="/components/state_refresh">
+      </amp-state>
+      <button class="ampstart-btn caps m1" on="tap:time.refresh">
+        Refresh
+      </button>
+      <div [text]=time></div>
```
sebastianbenz: For this to work, you endpoint needs to only return a string. Otherwise you need to do: `time.time`.

kul3r4: thanks, updated

backend/time.go
```diff
@@ -0,0 +1,32 @@
...
+
+package backend
+
+import (
+	"net/http"
+	"time"
+)
+
+const (
+	BIND_SAMPLE_PATH = "/" + CATEGORY_COMPONENTS + "/state_refresh/"
```
sebastianbenz: Idea: make the name of this endpoint more generic, e.g. `/time`

kul3r4: good idea, done

kul3r4:
@sebastianbenz have you checked the change is in prod before merging?

sebastianbenz:
yes - works on canary

kul3r4:
but if it doesn't work on prod yet than we shouldn't deploy it otherwise we will break the page

PR #1577, opened 2018-10-23
Refactor Compiler

sebastianbenz:
* remove unused methods
* move doc transformation to parsing step
* updated dependencies
* accidently converted all functions to use new syntax


sebastianbenz:
Fixed build

Issue #1584, opened 2018-10-26
Open Lightbox button stops working


nick-leung:
**Describe the bug**
The lightbox can't be shown again after it is closed.

**To Reproduce**
1. Go to 'https://ampbyexample.com/components/amp-bind/#hiding-and-showing-things' 
2. Scroll down to the lightbox example.
3. Click the "Open Lightbox" button.
4. Click anywhere to close the lightbox.
5. Click the "Open Lightbox" button again.

**Expected behavior**
The lightbox should appear again.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser: Chrome
 - Version: 69

PR #1585, opened 2018-10-26
Allow button to work after lightbox is closed

nick-leung:
Users were unable to open the lightbox a second time after closing the
initial lightbox. This happened because showLightbox was not being set to
false.

* Set state to change showLightbox to false when the user closes the lightbox


nick-leung:
Fixes Issue #1584.

PR #1586, opened 2018-10-26
Add snippets API

sebastianbenz:
Add an API that splits a sample file into multiple snippets. Each
snippet is a separate annotated code section with a standalone document
only containing the respective code section in the body.


lib/examples/standalone.js
```diff
@@ -17,15 +17,22 @@
 'use strict'
 
 const path = require('path');
-const {parseSample} = require('../index.js');
+const {parseSample} = require('../');
+
+/**
+ * Sample demonstrating how to parse and pre-compile sample file. The result is
+ * a document containing head, css and codesections as well as the pre-compiled
```
kul3r4: codesections -> code sections

sebastianbenz: fixed

sebastianbenz:
The API is available via a NPM module. Will document this properly once we've fully migrated.

PR #1591, opened 2018-10-30
Remove amp-google-document-embed experiment notification

cvializ:
It looks like this was left over from when the experiment was removed from the component.


kul3r4:
Thanks!

PR #1592, opened 2018-10-31
Add SXG support to ABE

ithinkihaveacat:
This PR contains two things required for ampbyexample.com to support signed-exchange (SXG). This will ultimately enable ABE content to appear in the Google Search and similar under the ampbyexample.com URL ([more info](https://wicg.github.io/webpackage/draft-yasskin-webpackage-use-cases.html#private-prefetch)).

1. It modifies the front-end to proxy requests with an `Accept` header of `application/signed-exchange`  to `https://abe-packager.appspot.com/`, which is a separate GCP node running https://github.com/ampproject/amppackager which returns "signed" versions of ABE's valid AMP resources.
2. Contains the Dockerfile and related scripts and configuration to build and deploy `https://abe-packager.appspot.com/`.

The algorithm this code is intending to follow with respect to what requests are proxied to amppackager is:

1. If the URL starts with `/amppkg/`, forward the request unmodified.
2. If the URL points to an AMP page and the `AMP-Cache-Transform` request
        header is present, rewrite the URL by prepending `/priv/doc` and forward
        the request.
3. If at all possible, don't forward requests for non-AMP pages; its
       transforms may break non-AMP HTML.
4. DO NOT forward `/priv/doc` requests; these URLs are meant to be
        generated by the frontend server only.

 ([source](https://github.com/ampproject/amppackager#productionizing))

/cc @twifkak

### DIFF
--- a/backend/packager.go
+++ b/backend/packager.go
@@ -1,0 +1,46 @@
+// Copyright Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backend
+
+import (
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/urlfetch"
+)
+
+func InitPackager() {
+	http.HandleFunc("/amppkg/", func(w http.ResponseWriter, r *http.Request) {
+		u, err := url.Parse(PACKAGER_PREFIX)
+		if err != nil {
+			log.Fatal(err)
+		}
+		p := httputil.NewSingleHostReverseProxy(u)
+		// Need to fiddle with Transport; on GAE this needs to change on every
+		// request. Might not be necessary under Go 1.11:
+		// https://cloud.google.com/blog/products/application-development/go-1-11-is-now-available-on-app-engine
+		p.Transport = &urlfetch.Transport{
+			Context: appengine.NewContext(r),
+		}
+		log.Printf("Proxying request for [%s] to [%s]", r.URL.String(), PACKAGER_PREFIX)
+		p.ServeHTTP(w, r)
+	})
+	// Blocking /priv/doc is one of the "productionizing" steps:
+	// https://github.com/ampproject/amppackager#productionizing
+	http.HandleFunc("/priv/doc", http.NotFound)
+}
--- a/backend/request.go
+++ b/backend/request.go
@@ -86,6 +86,15 @@ func GetHost(r *http.Request) string {
 	return "https://" + r.Host
 }
 
+func SetVary(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		h.ServeHTTP(w, r)
+		canonical(w.Header(), "vary")
+		add(w.Header(), "vary", "Accept")
+		add(w.Header(), "vary", "AMP-Cache-Transform")
+	})
+}
+
 func SetContentTypeJson(w http.ResponseWriter) {
 	w.Header().Set("Content-Type", "application/json")
 }
@@ -143,4 +152,47 @@ func onlyPost(next http.HandlerFunc) http.HandlerFunc {
 		}
 		next.ServeHTTP(w, r)
 	}
+}
+
+// Converts header entries associated with key to canonical form. In particular,
+// multiple headers are collapsed into one.
+func canonical(h http.Header, key string) {
+	v := h[http.CanonicalHeaderKey(key)]
+	var a []string
+	for _, vv := range v {
+		if vv != "" {
+			a = append(a, stringMap(strings.Split(vv, ","), strings.TrimSpace)...)
+		}
+	}
+	if len(a) != 0 {
+		h[http.CanonicalHeaderKey(key)] = []string{strings.Join(a, ", ")}
+	}
+}
+
+// Adds value associated with header if not already present. Assumes keys are
+// unique.
+func add(h http.Header, key string, value string) {
+	v := h[http.CanonicalHeaderKey(key)]
+	if len(v) == 0 {
+		h[http.CanonicalHeaderKey(key)] = []string{value}
+	} else {
+		a := stringMap(strings.Split(v[0], ","), strings.TrimSpace)
+		for _, vv := range a {
+			if vv == value {
+				return
+			}
+		}
+		h[http.CanonicalHeaderKey(key)] = []string{strings.Join(append(a, value), ", ")}
+	}
+}
+
+// stringMap returns a new slice containing the results of applying the function f to
+// each string in the original slice. From
+// https://gobyexample.com/collection-functions
+func stringMap(vs []string, f func(string) string) []string {
+	vsm := make([]string, len(vs))
+	for i, v := range vs {
+		vsm[i] = f(v)
+	}
+	return vsm
 }
--- a/backend/static.go
+++ b/backend/static.go
@@ -15,9 +15,15 @@
 package backend
 
 import (
+	"log"
 	"net/http"
+	"net/http/httputil"
+	"net/url"
 	"os"
 	"strings"
+
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/urlfetch"
 )
 
 const (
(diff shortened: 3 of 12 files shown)


backend/packager.go
```diff
@@ -0,0 +1,44 @@
...
+
+func InitPackager() {
+	http.HandleFunc("/amppkg/", func(w http.ResponseWriter, r *http.Request) {
+		u, err := url.Parse(PACKAGER_PREFIX)
+		if err != nil {
+			log.Fatal(err)
+		}
+		p := httputil.NewSingleHostReverseProxy(u)
+		// Need to fiddle with Transport; on GAE this needs to change on every
+		// request. Might not be necessary under Go 1.11:
```
sebastianbenz: feel free to update the go version if that helps

backend/request.go
```diff
@@ -86,6 +87,55 @@ func GetHost(r *http.Request) string {
 	return "https://" + r.Host
 }
 
+func stringMap(vs []string, f func(string) string) []string {
+	vsm := make([]string, len(vs))
+	for i, v := range vs {
+		vsm[i] = f(v)
+	}
+	return vsm
```
sebastianbenz: I like the GO way of naming variables....not.

sebastianbenz: (it's fine)

backend/packager.go
```diff
@@ -0,0 +1,44 @@
...
+		// https://cloud.google.com/blog/products/application-development/go-1-11-is-now-available-on-app-engine
+		p.Transport = &urlfetch.Transport{
+			Context: appengine.NewContext(r),
+		}
+		log.Printf("Proxying request for [%s] to [%s]", r.URL.String(), PACKAGER_PREFIX)
+		p.ServeHTTP(w, r)
+	})
+	http.HandleFunc("/priv/doc/", http.NotFound)
```
sebastianbenz: why do we need this?

backend/request.go
```diff
@@ -86,6 +87,55 @@ func GetHost(r *http.Request) string {
...
+}
+
+// Converts header entries associated with key to canonical form. In particular,
+// multiple headers are collapsed into one.
+func canonical(h http.Header, key string) {
+	v := h[textproto.CanonicalMIMEHeaderKey(key)]
+	var a []string
+	for _, vv := range v {
+		if vv != "" {
+			a = append(a, stringMap(strings.Split(vv, ","), strings.TrimSpace)...)
```
sebastianbenz: just a nit (feel free to ignore): I prefer methods to be ordered in call order (e.g. stringMap goes below canonical). That makes it easier to understand what's going on.

backend/static.go
```diff
@@ -41,18 +48,62 @@ func handleNotFound(h http.Handler) http.HandlerFunc {
...
+	return p
+}
+
+func shouldProxy(r *http.Request) bool {
+	act := r.Header.Get("amp-cache-transform")
+	if act == "" {
+		return false
+	}
+	accept := r.Header.Get("accept")
+	if accept != "" {
```
sebastianbenz: nit: I'd prefer an early exit to avoid nesting

backend/static.go
```diff
@@ -41,18 +48,62 @@ func handleNotFound(h http.Handler) http.HandlerFunc {
...
+			}
 		}
-		SetDefaultMaxAge(w)
-		h.ServeHTTP(w, r)
-	})
+	}
+	return false
+}
+
+func isAMP(r *http.Request) bool {
```
sebastianbenz: You could build a map of valid AMP files during compilation and use it here.

packager/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# (see https://blog.sgmansfield.com/2016/06/working-with-forks-in-go/), we
+# instead download the upstream, apply the appropriate patch (love love GitHub's
+# API!), and then install the patched version.
+RUN go get -d -u github.com/ampproject/amppackager/cmd/amppkg
+RUN curl -s https://github.com/ampproject/amppackager/compare/master...ithinkihaveacat:abe.patch | git -C $GOPATH/src/github.com/ampproject/amppackager apply
+RUN go install github.com/ampproject/amppackager/cmd/amppkg
+
+USER amppkg
+EXPOSE 8080
+CMD [ "bash", "-c", "./decrypt.sh && exec amppkg -config /amppkg/amppkg.toml" ]
```
twifkak: Where does the stdout from this cmd go? Does appengine store it somewhere for later perusal?

backend/request.go
ithinkihaveacat: I'm not even sure what's going on here, and I'm a short variable name aficionado … this is basically copy-and-pasted from https://gobyexample.com/collection-functions.

backend/packager.go
ithinkihaveacat: I think I'm just going to keep things as they are, it works now, and not sure of the implications of moving to the 1.11 runtime…

ithinkihaveacat:
@sebastianbenz By server-side generated AMP, do you mean AMP dynamically generated, at runtime? Do you mind if those resources are not packaged? It's awkward to do, especially since ABE sits in front of amppackager.

Even without this there's 651 static AMP files that are packaged at the moment, including (I would guess) the vast majority of pages that would be popular on search, and so there should be enough to demo. It's perhaps useful to test that everything works on a site that has all three of SXG AMP, AMP and non-AMP.

backend/packager.go
ithinkihaveacat: Blocking access to `/priv/doc/` is one of the rules for [productionizing amppackager](https://github.com/ampproject/amppackager#productionizing)—`/priv/doc/` is supposed to be a "private" namespace used only for the front-end and `amppackager` to communicate.

backend/request.go
ithinkihaveacat: I have no preference; changed in e553873.

backend/packager.go
twifkak: Isn't /priv/doc not handled by this server by default? That should be fine, I think. The important thing is that randos can't generate /priv/doc that are forwarded to the packager (without sanitization).

backend/static.go
```diff
@@ -41,18 +48,62 @@ func handleNotFound(h http.Handler) http.HandlerFunc {
...
+}
+
+func shouldProxy(r *http.Request) bool {
+	act := r.Header.Get("amp-cache-transform")
+	if act == "" {
+		return false
+	}
+	accept := r.Header.Get("accept")
+	if accept != "" {
+		a := strings.Split(accept, ",")
```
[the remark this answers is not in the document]
ithinkihaveacat: Thanks, I ripped all the "accept" parsing (because of separate discussion), but good to know that the "parser" wasn't right…

backend/packager.go
ithinkihaveacat: There's a handler for `/` connected to the `/priv/doc/` proxying process so without blocking `/priv/doc/`, requests for `/priv/doc/` can end up hitting the packager (with appropriate request headers).

The URL gets a bit mangled (so a request for `https://ampbyexample.com/priv/doc/asdfasd` gets translated into `/priv/doc/?sign=https%3A%2F%2Fampbyexample.com%2Fpriv%2Fdoc%2Fasdfasd`) and so is difficult to exploit, but it seems worthwhile to just block here? I might be misunderstanding.

backend/request.go
```diff
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/textproto"
```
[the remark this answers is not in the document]
ithinkihaveacat: Thanks! Changed in b443eef5f51ef8853955c0f828cd30d290cfb8c9.

backend/request.go
```diff
@@ -86,6 +87,55 @@ func GetHost(r *http.Request) string {
...
+		}
+		h[textproto.CanonicalMIMEHeaderKey(key)] = []string{strings.Join(append(a, value), ", ")}
+	}
+}
+
+func SetVary(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		canonical(w.Header(), "vary")
+		add(w.Header(), "vary", "Accept")
+		add(w.Header(), "vary", "AMP-Cache-Transform")
```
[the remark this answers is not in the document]
ithinkihaveacat: Well … partly! I was enjoying the go HTTP libraries so much that I thought it would be super straightforward to dedupe headers, but it turns out that whilst pretty much everything you might want to do re the networking side of HTTP is possible via the standard libraries the header munging part is not. Hence, this hassle…

I don't have any particular reason to believe this is actually necessary, but you never really know when it comes to HTTP…

backend/static.go
ithinkihaveacat: You mean gulp's `build:packager` task could create a `sitemap.json`, which is consulted here to figure out what resources the packager can handle? I can do this if necessary, just adds time + deployment complexity. (Note I'm assuming the proxying to packager never needs to be fast, and that a "slow" 404 if `isAMP()` is wrong is okay.)

packager/README.md
```diff
@@ -0,0 +1,43 @@
+Prerequistes:
```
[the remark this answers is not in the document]
ithinkihaveacat: Fixed.

packager/amppkg.toml
```diff
@@ -0,0 +1,140 @@
...
+    # Examples of paths you might want to exclude:
+    #   * Personalized content, such as user settings pages. Signed exchanges
+    #     are cached globally and served to all users. (Personalization that
+    #     happens at runtime via JS is fine.) This provides defense-in-depth,
+    #     in addition to the Cache-Control: public check.
+    #   * User-generated content, such as forums. For instance, if there's a
+    #     chance of an XSS vulnerability in your templating library, the impact
+    #     of such event is higher here: even after you've fixed the bug, caches
+    #     may serve your signed packages for up to 7 days.
+    PathExcludeRE = []
```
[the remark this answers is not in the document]
ithinkihaveacat: Fixed. (Not sure how that happened.)

backend/packager.go
twifkak: Ah, I see! Well, I guess that prevents people from wasting ABE's time by constructing a really long recursive `/priv/doc` URL.

packager/README.md
```diff
@@ -0,0 +1,43 @@
...
+Encrypt file:
+
+```sh
+openssl aes-256-cbc -md md5 -e -k $PASSWORD -in plain.pem -out encrypted.pem.enc
+```
+
+Decrypt file:
+
+```sh
+openssl aes-256-cbc -md md5 -d -k $PASSWORD -in encrypted.pem -out plain.pem.enc
```
alin04: Should these be encpryted.pem.enc and plain.pem ? The ".enc" suffix seems flipped?

packager/README.md
```diff
@@ -0,0 +1,43 @@
+Prerequisites:
```
alin04: Consider markdown styling, headings or whatnot.

packager/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Maybe switch to one of the Google-provided images? But seems to be go 1.8, not 1.10:
+# https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build#base
+FROM golang:1.10
+
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_amd64 /usr/local/bin/dumb-init
```
alin04: interesting, about the dumb-init...

packager/README.md
ithinkihaveacat: They should! Thanks.

packager/README.md
ithinkihaveacat: Tweaked the README a bit.

backend/static.go
sebastianbenz: Just realised that you're doing this already in `build:packager`, should be easy to integrate there. But this is up to you.

packager/README.md
```diff
@@ -0,0 +1,45 @@
+## Prerequisites
+
+Acquire the encrypted private key and TLS certificate (these are not in the
+repo), and place in `certs/privkey.pem.enc` and `certs/cert.pem.enc`
+respectively.
+
+## Deploy and run on Google App Engine
```
alin04: Curious why you chose App Engine Flex over GKE? I think with the latter, the docker image needs to be uploaded to a registry, whereas with App Engine, that doesn't appear necessary. Any other reasons why one would be picked over the other?

packager/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY dist /www
+COPY certs/privkey.pem.enc /amppkg
+COPY certs/cert.pem.enc /amppkg
+COPY decrypt.sh /amppkg
+RUN chmod +x /amppkg/decrypt.sh
+RUN chown -R amppkg:amppkg /amppkg
+
+# Because it's not possible to directly install a fork of a package via "go get"
+# (see https://blog.sgmansfield.com/2016/06/working-with-forks-in-go/), we
+# instead download the upstream, apply the appropriate patch (love love GitHub's
```
alin04: what's the patch?

alin04: nm, found it: https://github.com/ampproject/amppackager/pull/183

packager/README.md
ithinkihaveacat: The main reason I landed on GAE Flex was because I had something similar that already worked there, and I wanted to make the smallest change possible! Also, I have the impression that anything involving Kubernetes is more complicated than I need, although that might not actually be fair…

packager/Dockerfile
ithinkihaveacat: Technically the patch references the branch at https://github.com/ithinkihaveacat/amppackager/tree/abe.

ithinkihaveacat:
@twifkak @alin04 Thank-you so much for your thorough reviews!

twifkak:
No problem! Looking forward to @alin04's work to make this an easier experience in the future.

PR #1601, opened 2018-11-05
disable renovate

sebastianbenz:
...until we've migrated to amp.dev


sebastianbenz:
Fixed as part of #1604

PR #1604, opened 2018-11-05
Playground sandbox

sebastianbenz:


sebastianbenz:
I should work better (that's why I've added it). But it'll also enable other samples to work.

PR #1619, opened 2018-11-13
Seatmap improvements

kul3r4:
- add space around the seatmap in which you can still double tap
- fix the point of zooming by adding `layout=fill` to the `amp-selector`

@sebastianbenz the `div` strategy is necessary as explained in the comment as adding a padding directly to the svg interacts with the `amp-pan-zoom` sizing, while just having a margin outside the `amp-pan-zoom` makes the margin not clickable for zooming.

@pbalaram FYI

### DIFF
--- a/src/30_Advanced/SeatMap.html
+++ b/src/30_Advanced/SeatMap.html
@@ -80,7 +80,7 @@ how to dynamically select and unselect seats.
           margin: auto;
       }
       .seatmap  amp-selector .svg-container {
-        padding: 0 20px 0 20px;
+        padding: 0 24px;
       }
     </style>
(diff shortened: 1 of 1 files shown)


kul3r4:
@sebastianbenz ping

src/30_Advanced/SeatMap.html
```diff
@@ -79,6 +79,9 @@
           right: 0;
           margin: auto;
       }
+      .seatmap  amp-selector .svg-container {
+        padding: 0 20px 0 20px;
```
sebastianbenz: shorter: `padding: 0 20px`.

sebastianbenz: note, we usually use multiples of 8 for padding (0.5 rem). I that case I'd suggest using 24px (if it fits with the sample).

kul3r4: done

kul3r4:
@sebastianbenz PTAL

PR #1632, opened 2018-11-29
add default styles to samples

sebastianbenz:


sebastianbenz:
screenshot 2018-11-30 at 00 23 08
screenshot 2018-11-30 at 00 22 57

PR #1634, opened 2018-11-30
update deps to fix build

sebastianbenz:


sebastianbenz:
this'll hopefully fix the build

PR #1642, opened 2018-12-04
Sample migration 3

kul3r4:
@sebastianbenz


src/20_Components/amp-brid-player.html
```diff
@@ -23,7 +23,7 @@
 
   <!-- ## Basic Usage -->
```
sebastianbenz: Can you fix the heading

kul3r4: done

kul3r4:
@sebastianbenz PTAL

PR #1644, opened 2018-12-04
migrate more samples

sebastianbenz:
* Rewrote amp-bind sample to be shorter and more sample focused.
* Moved a few samples to the respective component samples
* Fixed a indentation bug resulting in doc not being formatted

### DIFF
--- a/src/amphtml-email/10_Introduction/_Hello_World.html
+++ b/src/amphtml-email/10_Introduction/_Hello_World.html
@@ -68,9 +68,7 @@ Learn how to create your first AMPHTML Email. This sample covers the basic struc
       with equivalent or slightly enhanced custom AMP HTML tags (see
       [HTML Tags in the specification](https://github.com/ampproject/amphtml/blob/master/spec/amp-html-format.md)).
 
-      <p class="mb2">
         <div class="p1 ampstart-card abe-info"><strong>Important:</strong> URLs most use absolut paths in AMPHTML email.</div>
-      </p>
     -->
     <div class="images">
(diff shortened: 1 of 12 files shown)


src/amphtml-email/10_Introduction/_Hello_World.html
```diff
@@ -65,17 +65,15 @@
...
-      </p>
     -->
     <div class="images">
       <amp-img src="https://ampbyexample.com/img/amp.jpg" width="800" height="600" layout="responsive"></amp-img>
       <!--
         ### AMP Components
 
         The following AMP components are supported in AMPHTML email messages.
-        <h4>Dynamic Content </h4>
+        **Dynamic Content </h4>
```
kul3r4: I missed this: remove the closing h4

templates/new-example.html
```diff
@@ -40,9 +41,10 @@
 </head>
 <body>
 
-  <!-- ## Basic Usage -->
-  <!--
-  TODO: Describe your example here
+  <!-- 
+    ## Basic Usage
```
kul3r4: Usage -> usage

sebastianbenz:
@kul3r4 PTAL

I've started improving the parser (to avoid the case where docs are printed as code). This might still need more work. Will address in another PR.

src/amphtml-email/10_Introduction/_Hello_World.html
sebastianbenz: this is broken anyway and needs a complete refactor...plan to do this in another PR

PR #1646, opened 2018-12-05
Sample migration 5

kul3r4:


src/20_Components/amp-call-tracking.html
```diff
@@ -40,11 +40,9 @@
 
     We also need to set a default phone number that will be shown before the endpoint responds. This is done via an `<a>` element inside the `amp-call-tracking` tag.
   -->
-  <p>
-    <strong>Call us!</strong><br>
+    <div>Call us!</div>
```
sebastianbenz: Wondering: just we remove this as well? Or does it help to explain the context?

src/20_Components/amp-date-picker.html
```diff
@@ -246,6 +237,7 @@
         input-selector="#static-picker-input">
       </amp-date-picker>
 
+  <!-- ## highlighted and blocked attributes -->
```
sebastianbenz: Uppercase highlighted

src/20_Components/amp-call-tracking.html
kul3r4: removed

kul3r4:
@sebastianbenz PTAL

kul3r4:
@sebastianbenz thanks for the feedback, addressed all from above

src/20_Components/amp-date-picker.html
```diff
@@ -264,7 +259,7 @@
       blocked="FREQ=WEEKLY;WKST=SU;BYDAY=SA,SU">
     </amp-date-picker>
 
-  <!-- ## allow-blocked-dates attribute -->
+  <!-- ## Allow-blocked-dates attribute -->
```
sebastianbenz: Not sure about this as it's a pre-defined attribute name.

PR #1650, opened 2018-12-06
Disable the blue rectangle on tab button focus

dandv:


sebastianbenz:
don't we need this for accessibility?

PR #1664, opened 2018-12-13
Sample migration 10

sebastianbenz:


src/20_Components/amp-video.html
```diff
@@ -1,7 +1,7 @@
 <!--
   ## Introduction
 
-  Embed videos into your AMP HTML files. Video source files should be served via HTTPS.
+  Use [`amp-video`](https://www.ampproject.org/docs/reference/components/amp-video) Embed videos into your AMP HTML files. Video source files should be served via HTTPS.
```
kul3r4: this 2 sentences don't make sense together. I think you are missing "to embed" instead of "Embed"

sebastianbenz: done

sebastianbenz:
@kul3r4 PTAL

PR #1666, opened 2018-12-14
Update amp-video-docking example

alanorozco:
- Docking is now a separate extension.
- Action is now either triggered from video or from slot depending on target.

/cc @aghassemi


aghassemi:
LGMT

PR #1682, opened 2018-12-19
Migrate seatmap

kul3r4:
@sebastianbenz PTAL


src/30_Advanced/SeatMap.html
```diff
@@ -36,6 +36,7 @@
       :root {
       --abe-light-pink: rgba(182, 8, 69, 0.5);
       --abe-pink: rgb(182, 8, 69);
```
sebastianbenz: Can you replace these with `--color-primary` and `--color-secondary`.

kul3r4: done

kul3r4:
@sebastianbenz PTAL

PR #1695, opened 2018-12-21
migrate visual effects

sebastianbenz:


src/50_Visual_Effects/Scroll_to_top.html
```diff
@@ -108,26 +140,17 @@
...
-
-    Aliquam tincidunt, quam nec fermentum viverra, orci dolor hendrerit ex, at suscipit tellus quam eu enim. Phasellus mollis dapibus nunc, vel sollicitudin dui consectetur id. Aliquam erat volutpat. Nunc arcu massa, fermentum in mauris sit amet, pulvinar feugiat tellus. Nam pretium tempus tortor sit amet venenatis. Aliquam hendrerit porta augue. Sed velit enim, rutrum vitae posuere at, suscipit in turpis. Sed porta diam sed arcu semper sollicitudin. Ut ac leo ut orci malesuada lobortis nec ac lorem. Cras nec turpis tellus. Nam vestibulum egestas ligula. Aliquam leo tortor, varius a iaculis vel, laoreet et neque. Vivamus sed mauris at sapien varius ultrices. Aliquam aliquam tincidunt ante, semper gravida nisl maximus vitae. In consectetur eleifend magna eu tincidunt.
-
-  </p>
-  <!-- We use the `scrollTo` action to scroll the page when the button is tapped. Find more about actions [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-actions-and-events.md).-->
-  <button id="scrollToTopButton" on="tap:top-page.scrollTo(duration=200)" class="scrollToTop">⌃</button>
+    </script>
+  </amp-animation>
+  <!--
```
kul3r4: there is no header for this section

kul3r4: Where it says "Here is a dummy spacer." there is no header. My understanding is that each comment needs an header, if it doesn't than this is fine.

sebastianbenz: Every significant section/sample should have a header - the spacer is not important.

kul3r4: ok I see1

PR #1772, opened 2019-01-02
Migrate live blog

kul3r4:


src/60_Samples_%26_Templates/Live_Blog.html
```diff
@@ -85,77 +82,85 @@
...
         .heading > #summary {
           font-weight: 500;
         }
         .heading > small {
-          color: #656565;
+          color: var(--color-primary);
         }
         .date {
-          color: #a8a3ae;
+          color: var(--color-secondary);
```
sebastianbenz: nit: color-primary is easier to read

src/60_Samples_%26_Templates/Live_Blog.html
```diff
@@ -1,8 +1,6 @@
 <!---
 
 preview: default
```
sebastianbenz: my bad, we disabled validation because of the go template literals not passing validation

src/60_Samples_%26_Templates/Live_Blog.html
kul3r4: done

src/60_Samples_%26_Templates/Live_Blog.html
kul3r4: right, I reverted the change

PR #1783, opened 2019-01-07
Migrate product browse

kul3r4:


templates/css/product.css
```diff
@@ -30,9 +43,9 @@ amp-carousel {
   left:36px;
   font-size:18px;
   font-weight:400;
-  color:#fff;
+  color:var(--color-text-light);
   text-transform:uppercase;
-  margin:0 16px;
+  margin:0 var(--space-2);
   height:32px;
```
sebastianbenz: use `space-4`

templates/css/product.css
```diff
@@ -49,8 +62,8 @@ amp-carousel {
 .header input[type=submit] {
   position:absolute;
   top:10px;
```
sebastianbenz: Would it look OK to change this to space-1 or space-2?

templates/css/product.css
```diff
@@ -65,8 +78,8 @@ amp-carousel {
...
```
sebastianbenz: Would it look OK to change this to space-1 or space-2?

templates/css/product.css
```diff
@@ -65,8 +78,8 @@ amp-carousel {
...
   right:24px;
```
sebastianbenz: use space-3

templates/css/product.css
```diff
@@ -65,8 +78,8 @@ amp-carousel {
...
-  background:#ededed;
-  padding:0 8px;
+  background:var(--color-bg-light);
+  padding:0 var(--space-1);
   width:20vw;
   height:32px;
```
sebastianbenz: use space-4

src/60_Samples_%26_Templates/Product_Browse_Page.html
```diff
@@ -90,15 +98,15 @@
              display: block;
              height: 100px;
              padding: 0;
-             padding-right: 8px;
+             padding-right: var(--space-1);
            }
            .products amp-img {
              line-height: 24px;
```
sebastianbenz: use space-3

src/60_Samples_%26_Templates/Product_Browse_Page.html
```diff
@@ -90,15 +98,15 @@
...
              font-weight: 400;
              font-size: 24px;
```
sebastianbenz: use space-3

src/60_Samples_%26_Templates/Product_Browse_Page.html
```diff
@@ -127,7 +135,7 @@
          .autosuggest-box {
             position: absolute;
             width: 100%;
-            background-color: #fafafa;
+            background-color: var(--color-text-light);
             box-shadow: 0px 2px 6px rgba(0,0,0,.3);
```
sebastianbenz: use standard box-shadow or does it need to look different?

src/60_Samples_%26_Templates/Product_Browse_Page.html
```diff
@@ -127,7 +135,7 @@
...
           }
           .hidden {
```
sebastianbenz: Do we still need the hidden rule? Or could we replace this with the attribute?

templates/css/product.css
```diff
@@ -19,7 +32,7 @@ amp-carousel {
   left:0;
   font-size:18px;
   font-weight:700;
-  color:#fff;
+  color: var(--color-text-light);
   text-transform:uppercase;
   padding:13px
```
sebastianbenz: Would it make sense to replace this either with space-2 or space-3?

kul3r4:
@sebastianbenz your comment regarding " Can you also migrate the colors.": which ones you mean? I added the rule in product.css and then used them in the code

templates/css/product.css
kul3r4: done

templates/css/product.css
kul3r4: done

templates/css/product.css
kul3r4: done

templates/css/product.css
kul3r4: done

templates/css/product.css
kul3r4: done

src/60_Samples_%26_Templates/Product_Browse_Page.html
kul3r4: done

src/60_Samples_%26_Templates/Product_Browse_Page.html
kul3r4: done

src/60_Samples_%26_Templates/Product_Browse_Page.html
kul3r4: removed

sebastianbenz:
Demo is still broken. Am I doing something wrong?

Background is no longer transparent:

screenshot 2019-01-08 at 13 54 31

This seems wrong here:

screenshot 2019-01-08 at 13 54 26

Alignment off: 

screenshot 2019-01-08 at 13 54 23

Wrong colors:

screenshot 2019-01-08 at 13 54 18

Inconsistent styling:

screenshot 2019-01-08 at 13 54 14

Error in the console:

screenshot 2019-01-08 at 13 54 08

Animation off:

screenshot 2019-01-08 at 13 56 41

Inconsistent alignment and too much whitespace below and left/right.

screenshot 2019-01-08 at 13 57 10

kul3r4:
@sebastianbenz for the wrong colors and inconsistent styling I don't have a particular css causing that but I believe it's amp-start, maybe because we import it in the <style amp-boilerplate>?

kul3r4:
@sebastianbenz fixed the layout. The css now has become too bigger, not sure how to fix it.

sebastianbenz:
You can ignore the CSS size limit. That hopefully is not going to be a problem with amp.dev as it's no longer using ampstart.

kul3r4:
@sebastianbenz Applied all the suggestions

templates/css/product.css
```diff
@@ -48,45 +61,45 @@ amp-carousel {
...
   border-radius:4px 0
 }
 
 .header input[type=search],.header input[type=text] {
   position:absolute;
-  top:10px;
-  right:24px;
-  background:#ededed;
-  padding:0 8px;
+  top:var(--space-1);
```
sebastianbenz: I was wrong, this needs to be 10px for the input field to be centered vertically

sebastianbenz:
And one more thing: can you use the same layout for the product list as in the original version (padding & margin).

kul3r4:
Since you suggested to make the header with color-primary, which color should I use for the button?

sebastianbenz:
The build fails because the product sample has gotten too big, which needs to skip the validation.

sebastianbenz:
Does color-secondary look alright for the button? Or maybe vice versa?

kul3r4:
@sebastianbenz it would look like this 
header-light-blue

sebastianbenz:
Looks good! Can you post a full screenshot as I can't reproduce it locally.

kul3r4:
@sebastianbenz it looks like:
product-browse-page

I also pushed the latest changes

sebastianbenz:
Looks good. Two things:

1. Can you add the same margins as in the original version (around
headings, sections, ...)
2. Can you align the menu icon and product.

>

kul3r4:
@sebastianbenz I missed those, sorry!
here is the new version
product-browse-page-2

sebastianbenz:
Perfect. thanks!

PR #1785, opened 2019-01-09
🐛Fix joining analytics example

calebcordry:
We introduced logic into linker amproject/amphtml#19520 to not decorate links to the same domain unless specifically enabled. That feature broke this example. This PR specifically enables the ABE domain to be decorated.


calebcordry:
cc/ @jeffjose

PR #1787, opened 2019-01-11
Migrate product page

kul3r4:


src/60_Samples_%26_Templates/Product_Page.html
```diff
@@ -290,6 +306,74 @@
...
+    .color-selector ul, .quantity-selector ul, .size-selector ul {
+      padding: 0;
+      margin: var(--space-2);
+    }
+
+    .related-products {
+      display: inline-block;
+    }
+
+    .margin1, .related-products {
```
sebastianbenz: Can you find an approach that doesn't require encoding styles into classes?

sebastianbenz: For example: you can wrap the sample in a div `<div class="product-sample">` and then use `.product-sample > * { margin: ... }` to assign all children a margin.

kul3r4:
@sebastianbenz related to the comment "The text is too big for the button:", this is the original text and I am not styling the button as I understood this will be covered by general styling. what should I do?

PR #1788, opened 2019-01-12
Migrate recipe

kul3r4:


src/60_Samples_%26_Templates/Recipe.html
```diff
@@ -423,16 +429,17 @@ <h1 class="name">Unexpected side effects</h1>
   </p>
 
   <!-- ## Embedding related content -->
-
-  <h2>Related Recipes</h2>
   <!--
   It's a good idea to recommend related content to your readers. With `amp-list` it is possible to dynamically include personalized content into your AMPs. It's best to link to the AMP version of related content in order to keep the user in the fast AMP world. [Learn more about `amp-list`](/components/amp-list/)
   -->
+  <div>
+  <h2>Related Recipes</h2>
```
sebastianbenz: I'd leave the heading out of the code block.

kul3r4:
@sebastianbenz as for the margin and padding I cannot find any difference with the version hosted in https://ampbyexample.com/samples_templates/recipe/preview/

sebastianbenz:
Ah right - but there should be :-). Can you fix this?

kul3r4:
@sebastianbenz PING

src/60_Samples_%26_Templates/Recipe.html
```diff
@@ -237,9 +237,8 @@
     }
     .heading {
       padding-bottom: var(--space-1);
-      margin: var(--space-2);
     }
-    .margin {
+    .recipe-sample {
```
sebastianbenz: This adds only margin around everything, but not the individual elements.

kul3r4:
@sebastianbenz updated the css rule according your suggestion, thanks!

PR #1824, opened 2019-02-13
fix amp-bind sample

### LINKED ISSUE #1816: amp-bind [class]
When using amp-bind to bind the [class] attribute, the initialization page reports an error: %s: "%s" is not a valid result for [class]. amp-bind null. What should I do with this problem? I also saw an error in the template page. You can see the console at this address is reporting an error: https://ampbyexample.com/components/amp-bind/source/#development=1

kul3r4:
Fixes #1816


src/20_Components/amp-bind.html
```diff
@@ -65,6 +65,7 @@
       You can dynamically change the CSS classes of an element by adding a binding to the `[class]` attribute.
     -->
     <div>
+      <amp-state id="myClass"><script type="application/json">""</script></amp-state>
```
sebastianbenz: We need to restructure the sample, as we haven't introduced the amp-state element at this point.

kul3r4: Should I move this section after the part that introduces `amp-state`?

sebastianbenz: How about moving the initialising state sample after binding text?

kul3r4: done

gaganabhay:
@sebastianbenz  Can you please merge it. Need this merge very badly. Thanks

PR #1848, opened 2019-03-01
Migrating Cart from CLIENT_ID to sessions + "Remove Items from Cart" functionality.

demianrenzulli:
Changes include:

- Migrating "Add To Cart" functionality from [ABE Product Page](https://ampbyexample.com/samples_templates/product_page/), from CLIENT_ID to sessions, based on best practices.
- Migrating [ABE Shopping Cart Page](https://ampbyexample.com/shopping_cart/) from server-side rendered, to dynamic list, to allow for removal of items from cart.
- Implementing "Remove Item from cart" functionality on Shopping Cart Page.

Notes:

- As discussed with @sebastianbenz, we are temporarily using a Node server on glitch (code [here](https://glitch.com/edit/#!/abe-cart-service)), to host the endpoints required for the different calls, instead of ABE backend.
- The solution uses the "double redirect" pattern, in combination with session usage, to ensure security, and cross-browser compatibility from cache to origin.
- I managed to add the summary of the solution in the product page. When the time comes, we can add a reference to the more detailed cross-origin post, written by @morsssss.
- The changes can be tested by running in localhost, without the need of running the backend server (as it uses a different endpoint). I can help testing on a staging URL, before launching to production, to make sure all use cases work as expected.

### DIFF
--- a/src/60_Samples_%26_Templates/Product_Page.html
+++ b/src/60_Samples_%26_Templates/Product_Page.html
@@ -691,19 +691,23 @@ This sample showcases how to build a product page in AMP HTML.
         <span [text]="product[product.selectedColor].sizes[product.selectedSize]">$5.99</span>
     </p>
 
-    <!-- ## Product page -->
+    <!-- ## Add To Cart -->
     <!-- The add-to-cart action action is implemented using  [amp-form](/components/amp-form/). In our sample we use the [amp-selector](/components/amp-selector/) for selecting different product properties.
-    Pressing the `ADD TO CART` button adds the product to a shopping cart page using the properties you have selected. Notice that the button URL contains the query `clientId={{ClientId}}`.
-    If an item is successfully added to the shopping cart, we redirect the user to the shopping cart page.
-    The redirect target is configured via a header in the server's form response ([redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)):
 
-      access-control-expose-headers:AMP-Access-Control-Allow-Source-Origin,AMP-Redirect-To
-      amp-redirect-to:http://ampbyexample.com/shopping_cart/?clientid=amp-123456789 -->
+    Pressing the `ADD TO CART` button triggers an `XHR POST` request to `/add_to_cart`. Here, we need to distinguish two scenarios:
 
-    <!-- We use the `CLIENT_ID` variable to identify the user, this enables storing a shopping cart between repeated visits of an AMP page (either via the AMP Cache or the original host). This variable can be used inside an amp-form by declaring an hidden input value with `default-value="CLIENT_ID(cart)`. Read more about variable substitution [here](https://github.com/ampproject/amphtml/blob/master/spec/amp-var-substitutions.md). -->
+    - If the user is visiting the cached version of the page, the server receives the request on the `POST` handler and redirects again to `/add_to_cart`, but transforming the initial request, into `GET`, using the `AMP-Redirect-To` header (see [redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)). Next, the `GET` handler adds the item to the session, and redirects to the final destination: the `/shopping_cart` page.
+    - If the user is visiting the page from its origin, there's no need for a 'double redirect', since the server can access the session directly on the `POST` handler, then add the item to the cart, and, finally redirect to the `/shopping_cart` page.
+     -->
+
+     <!-- 
+      The first step of the previous list is necessary, in order to move the user outside the domain of the cache (where the session can't be accessed on browsers that block third party cookies), into an intermediate step in the origin (where the session cookie, is a first party cookie).
+
+      That way, we can maintain user state across origins, in a secure way, while being compatible with all the major browsers.
+     -->
 
       <form id="order" method="POST"
-            action-xhr="/samples_templates/product_page/add_to_cart"
+            action-xhr="https://abe-cart-service.glitch.me/add_to_cart"
             target="_top" class="order margin1">
         <div class="color-selector">
           <label for="color">Color:</label>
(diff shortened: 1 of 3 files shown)


src/Shopping_Cart.html
```diff
@@ -32,19 +36,73 @@
...
+                <h3>Your Basket is Empty. </h3>
+                {{/isEmpty}}
+                {{^isEmpty}}
+                {{#cartItems}}
+                <div class="p1">{{name}} - {{price}}</div>
+                <div class="details pb1">
+                    <div class="px1">Color: {{color}}</div>
+                    <div class="px1">Size: {{size}}</div>
+                    <div class="px1">Qty: {{quantity}}</div>
+                    <button type="button" style="background: none;box-shadow: none;border-radius: 0px;width:30px" on="tap: AMP.setState({cartItem:
```
sebastianbenz: Can you move the inline styles into `style amp-custom`.

src/Shopping_Cart.html
```diff
@@ -32,19 +36,73 @@
...
+    <!-- 
+    As said, we use [amp-form](https://www.ampproject.org/docs/reference/components/amp-form) to send an `XHR POST` request to `/delete-cart-item`:
+
+    1. The form contains two hidden input fields, bound to the `cartItem` object's variables. When the remove button is clicked, this object gets updated with the details of the item to remove from cart, so the form can send them on the call to the server.
+    2. The form updates the `cartItemsList` with the response from the server, so that  `amp-list` can be refreshed with the contents of the updated cart.
+    -->
+    <form id="form-cart-delete" method="POST" target="_top" action-xhr="https://abe-cart-service.glitch.me/delete-cart-item" on="submit-success: AMP.setState({
+                cartItemsList: event.response
+            })" novalidate="">
+        <input type="hidden" name="color" value="" [value]="cartItem.color">
```
sebastianbenz: I'm wondering whether we should add an id to the shopping cart items. Identifying items via color and size is not very intuitive.

src/60_Samples_%26_Templates/Product_Page.html
```diff
@@ -691,19 +691,23 @@ <h2 class="margin1">Apple</h2>
...
+    - If the user is visiting the cached version of the page, the server receives the request on the `POST` handler and redirects again to `/add_to_cart`, but transforming the initial request, into `GET`, using the `AMP-Redirect-To` header (see [redirecting after submission](https://github.com/ampproject/amphtml/blob/master/extensions/amp-form/amp-form.md#redirecting-after-a-submission)). Next, the `GET` handler adds the item to the session, and redirects to the final destination: the `/shopping_cart` page.
+    - If the user is visiting the page from its origin, there's no need for a 'double redirect', since the server can access the session directly on the `POST` handler, then add the item to the cart, and, finally redirect to the `/shopping_cart` page.
+     -->
```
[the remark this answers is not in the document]
demianrenzulli: I tried this, but, if I don't open a new comment, ABE formats the list adding a line break next to each bullet.
Example:

- 
text

This was solved, by closing the comment and creating a new one. Any suggestions?

src/Shopping_Cart.html
demianrenzulli: Good point.
The "product" state object doesn't contain Ids, but I've noticed the related [products response does](https://ampbyexample.com/json/related_products.json?__amp_source_origin=https%3A%2F%2Fampbyexample.com).
I could just add "1, 2, 3" to each product on the "product" object, and pass it on the form, so we can use it in the cart. 
Sounds good?

src/60_Samples_%26_Templates/Product_Page.html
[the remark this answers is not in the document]
sebastianbenz: OK, that's weird. Looks like a marked.js bug then.

src/Shopping_Cart.html
sebastianbenz: perfect!

demianrenzulli: Done! Also fixed the prices, as it was always sending the same value ($1.99).

demianrenzulli:
@sebastianbenz just added your suggestions. LMK if you'd like to change anything else.

src/60_Samples_%26_Templates/Shopping_Cart.html
```diff
@@ -81,6 +96,7 @@ <h3>Your Basket is Empty. </h3>
     <form id="form-cart-delete" method="POST" target="_top" action-xhr="https://abe-cart-service.glitch.me/delete-cart-item" on="submit-success: AMP.setState({
                 cartItemsList: event.response
             })" novalidate="">
+        <input type="hidden" name="id" value="" [value]="cartItem.id">
         <input type="hidden" name="color" value="" [value]="cartItem.color">
```
sebastianbenz: I think it'd be fine to only transmit the id, color and size shouldn't be needed for deletion. This would also mean that we can simplify the setState action to `AMP.setState({ itemToDelete: {{id}} })`.

demianrenzulli: Thanks Sebastian, quick comment here: 
For this use case, "color" and "id" play the same role, but regarding "size": I'm using it as an identifier as well. For example: users can remove all "Medium" size items of id "1", but not the "Small" sizes of the same ID. So I think we can get rid of color, but size should stay. 
Thoughts?

sebastianbenz: We could encode the size in the id. But maybe it's better then to leave it as it is...

demianrenzulli: Just removed the "color" and updated the glitch service as well, to accept only product ID and size.

demianrenzulli:
@sebastianbenz just tested all the use cases (including adding items to the cart from  [the cached version](https://www.google.com/amp/s/ampbyexample.com/samples_templates/product_page/) and it works well 👍

Only thing I've noticed is that the [Shopping Cart page](https://ampbyexample.com/samples_templates/shopping_cart/) has a link rel="canonical" pointing to the previous url (https://ampbyexample.com/shopping_cart/), so it's not discoverable.

In any case, testing [the cached URL](https://ampbyexample-com.cdn.ampproject.org/c/s/ampbyexample.com/samples_templates/shopping_cart/) of the shopping cart page, it works 👍

Should I create a separate PR to correct the link rel tag?

sebastianbenz:
Good spot! Please fix in another PR. Thanks!

PR #1981, opened 2019-03-26
fix amp-dailymotion sample

jakub-g:
The example was using an ID of a deleted video


sebastianbenz:
I've updated the respective samples on ampproject/docs as well.

jakub-g:
Thanks for merging. Will this be deployed automatically to https://ampbyexample.com/components/amp-dailymotion/ or it's a manual process? Does it need a green Travis build to deploy (because I see there are some validation failures from previously merges PRs)?

PR #2028, opened 2019-06-28
Allow parseSample to override content

fstanis:
Allow `parseSample` to override file contents instead of reading from `fs`. This is useful for on-the-fly processing, e.g. the file was created in-memory and isn't stored on the disk, such as gulp `Vinyl`.

Related to ampproject/docs#2389.


fstanis:
Interesting side-effect: this significantly speeds up sample build time for amp.dev.

